PR #36, opened 2016-12-27
generalizes configuration options for the pic32mx target

balazsracz:
Makes both 530 and 795 buildable from the same source tree.


balazsracz:
@heikens this is how I could make the pic32 build work for both processors at the same time.

Please test if this works in your build environment and produces a usable binary!

thanks,
Balazs

PR #40, opened 2017-01-01
IO board application for the Z Systems dev board

balazsracz:
Forks the ek-tm4c123 board files and IO board application and creates a simple firmware for the Z systems dev board.

This firmware runs CAN, and drives the direct output and samples the direct input pins.


dpharris:
What and where is Z System's dev board?

balazsracz:
It's Dana's board. He wrote about it to the mailing list.

PR #42, opened 2017-01-06
Pic32mx ram optimize for lccbuffer

heikens:
Mods to make async_blink work on the lccbuffer.


applications/io_board/targets/freertos.armv7m.z-dev-board-123/HwInit.cxx
```diff
@@ -1 +0,0 @@
-../../../../boards/ti-ek-tm4c123gxl-z-dev-board/HwInit.cxx
```
balazsracz: something strange happened here -- did you remove some files?

boards/mc-pic32mx-lccbuffer/ISRwrapper.inc
```diff
@@ -1 +0,0 @@
-../microchip-pic32mx-generic/ISRwrapper.inc
```
balazsracz: I generally prefer to keep the link from boards/xxx-lccbuffer to boards/xxx-generic, because that means xxx-lccbuffer shows me all the files that I need to add to the target directory in a new application. No need to guess which files I should be taking from where.

boards/microchip-pic32mx-duinomitemega/ISRwrapper.inc
```diff
@@ -1 +0,0 @@
-../microchip-pic32mx-generic/ISRwrapper.inc
```
balazsracz: please don't remove these links, it will break the build for the duinomite-mega targets, since those link to here.

src/freertos_drivers/pic32mx/Pic32mxCan.cxx
```diff
@@ -62,8 +62,7 @@ class Pic32mxCan : public Node
         , hw_(module)
         , overrunCount_(0)
     {
-        messageFifoArea_ = malloc(
-            (config_can_tx_buffer_size() + config_can_rx_buffer_size()) * 16);
+        messageFifoArea_ = malloc(2 * 16 * CAN_TX_RX_MESSAGE_SIZE_BYTES);
```
balazsracz: let's keep the configuration setting the same way for the PIC32 as all the other targets.
It's ok to replace the fixed 16 constant with CAN_TX_RX_MESSAGE_SIZE_BYTES but we should keep the config_() to allow the OVERRIDE_CONST(can_tx_buffer_size, nn); to function.

applications/io_board/targets/freertos.armv7m.z-dev-board-123/HwInit.cxx
heikens: I noticed that too. I didn't touch or remove any of the other boards files, I checked out the upstream/pic32mx-ram-optimize branch, Perhaps that branch isn't complete?

boards/mc-pic32mx-lccbuffer/ISRwrapper.inc
heikens: See previous comment. I think I know what happened here. My local master still has the old links for the duinomitemega, where ISRwrapper.inc is located in the application target directory. Probably that board didn't get updated while checking out the pic32mx-ram-optimize branch.

boards/microchip-pic32mx-duinomitemega/ISRwrapper.inc
heikens: See previous comment. I may need to resync my local master.

src/freertos_drivers/pic32mx/Pic32mxCan.cxx
heikens: What is the value of config_can_rx_buffer_size()? I couldn't find it just yet. We have one (1) receive FIFO and one (1) transmit FIFO, hence the value 2. Each FIFO can have up to 32 message buffers, I have assigned a fixed 16 constant here, which is the size of the receive FIFO. Each message is 16 bytes and is given by the constant CAN_TX_RX_MESSAGE_SIZE_BYTES. The transmit FIFO only uses 1 message buffer at the moment. Do you want/need to override the number of message buffers for the receive FIFO?

balazsracz:
The unit of can_tx_buffer_size() is number of messages and the default
value is 16:
https://github.com/bakerstu/openmrn/blob/master/src/utils/constants.cxx#L135

You can override this constant on a per-application or per-target basis if
you need to by using the OVERRIDE_CONST(...) syntax.
​

PR #44, opened 2017-01-16
FreeRTOSTCP for TI Launchpad 129/1294 fixes

smcharg:
Fixes a compatibility issue due to changes in the FreeRTOS Labs
sockaddr structure which were manifest in bind operations failing to
function properly.
Change the port number used by the can_eth application from 9000 to
12021 for consistency with the hub application.

### DIFF
--- a/applications/can_eth/main.cxx
+++ b/applications/can_eth/main.cxx
@@ -55,7 +55,7 @@
 #include <sys/socket.h>
 #include <arpa/inet.h>
 
-#define version "can_eth Vrs 0.1"
+#define VERSION "can_eth Vrs 0.1"
 
 
 Executor<1> g_executor("g_executor", 0, 1024);
@@ -84,7 +84,7 @@ int appl_main(int argc, char* argv[])
     const int listen_port = 12021;
     int serial_fd = ::open("/dev/ser0", O_RDWR); // or /dev/ser0
     HASSERT(serial_fd >= 0);
-    printf(version);
+    printf(VERSION);
     printf(" started, listening on port %d\n",listen_port);
 
     GcTcpHub hub(&can_hub0,listen_port);
--- a/applications/can_eth/main.cxx
+++ b/applications/can_eth/main.cxx
@@ -55,6 +55,8 @@
 #include <sys/socket.h>
 #include <arpa/inet.h>
 
+#define version "can_eth Vrs 0.1"
+
 
 Executor<1> g_executor("g_executor", 0, 1024);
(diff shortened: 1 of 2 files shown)


balazsracz:
Thanks for debugging this issue! If Stuart can't get to it I will review
this in a few days.

--

src/freertos_drivers/net_freertos_tcp/FreeRTOSTCPSocket.cxx
```diff
@@ -177,9 +177,10 @@ int FreeRTOSTCPSocket::bind(
...
     }
-    // fr_address.sin_addr = ((const struct sockaddr_in *)(address))->sin_addr;
-    // fr_address.sin_port = address->sin_port;
-    memcpy(&fr_address, address->sa_data, sizeof(fr_address));
+
+    struct sockaddr_in *sin = (struct sockaddr_in *)(address);
+    fr_address.sin_addr = sin->sin_addr.s_addr;
+    fr_address.sin_port = sin->sin_port;
 
     int result = FreeRTOS_bind(s->sd, &fr_address, address_len);
```
balazsracz: We are using address_len here, but that points to *address. We are passing in &fr_address.
Shouldn't we change this line to sizeof(fr_address) instead?

applications/can_eth/main.cxx
```diff
@@ -55,6 +55,8 @@
 #include <sys/socket.h>
 #include <arpa/inet.h>
 
+#define version "can_eth Vrs 0.1"
```
balazsracz: consider using capitals for constants, like VERSION

src/freertos_drivers/net_freertos_tcp/FreeRTOSTCPSocket.cxx
```diff
@@ -274,7 +277,7 @@ int FreeRTOSTCPSocket::accept(
 
     Socket_t sd = FreeRTOS_accept(s->sd, &fr_address, &fr_address_len);
```
balazsracz: would you need to set fr_address.length here as well?

smcharg: The present implementation in FreeRTOS_accept will store the port and source address in the passed freertos_sockaddr structure if non-null with no further checks.  The API documentation for FreeRTOS_accept indicates that at the present the length parameter is ignored, however should be set to the size of the freertos_sockaddr structure; no mention is made of conditioning the structure itself.

PR #53, opened 2017-02-12
Add mechanism to retrieve data from user tokens on the CC3200 HTTP se…

bakerstu:


src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -575,6 +577,17 @@ void CC32xxWiFi::run_on_network_thread(std::function<void()> callback)
 }
 
 /*
+ * CC32xxWiFi::add_http_get_callback()
+ */
+void CC32xxWiFi::add_http_get_callback(std::pair<std::function<std::string()>,
+                                                 const char *> callback)
+{
+    OSMutexLock l(&lock_);
+    httpGetCallbacks_.emplace_back(callback);
```
balazsracz: write it like this:
....emplace_back(std::move(callback));
this will avoid a copy.

note that this is a good candidate function for inlining, but you have to put it into the header for that.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.hxx
```diff
@@ -208,6 +209,13 @@ public:
      * isthe function to execute.*/
     void run_on_network_thread(std::function<void()> callback);
 
+    /** Add and HTTP get token callback.
+     * @param callback the std::pair<> of the function to execute and the
+     *        matching token to execute the callback on
```
balazsracz: could you describe what a "token" is? Also the documentation of the callback function needs to come here, i.e. when will the callback be executed and what should it return.
please document the lifetime requirements on the token const char*. How long does the memory pointed to stay alive? Who owns the memory?

src/utils/format_utils.hxx
```diff
@@ -71,4 +71,31 @@ char* unsigned_integer_to_buffer_hex(int value, char* buffer);
...
 string mac_to_string(uint8_t mac[6], bool colons = true);
 
+/// Formats an IPv4 address to string.
+///
+/// @param ip a 4-byte array storing the IPv4 address. ip[3] will be printed
+/// at the beginning.
+/// @param dots true to print dots, else false to exclude the dot
+/// seperators
+///
+/// @return a string containing a dot-separated hexadecimal printout of the
```
balazsracz: it seems this is not hexadecimal but decimal.

for this reason i'm not sure what is the use of dot == false. When do you call it with !dot?

src/utils/format_utils.hxx
```diff
@@ -71,4 +71,31 @@ char* unsigned_integer_to_buffer_hex(int value, char* buffer);
...
+string ipv4_to_string(uint8_t ip[4], bool dots = true);
+
+/// Formats an IPv4 address to string.
+///
+/// @param ip a uint32_t storing the IPv4 address. most significant bytes
+/// will be printed at the beginning.
+/// @param dots true to print dots, else false to exclude the dot
+/// seperators
+///
+/// @return a string containing a dot-separated hexadecimal printout of the
```
balazsracz: same here

src/freertos_drivers/net_cc32xx/CC32xxWiFi.hxx
```diff
@@ -227,6 +235,13 @@ public:
      */
     void sock_event_handler(void *context);
 
+    /** This function handles http server callback indication.  This is public
+     * only so that an extern "C" method can call it.  DO NOT use directly.
+     * @param context1 pointer to HTTP Server Event info
+     * @param context2 pointer to HTTP Server Response info
+     */
+    void http_server_callback(void *context1, void *context2);
```
balazsracz: these should not be void*
If you're worried about the dependency intriduced by the inclusion of simplelink headers, consider forward declaring the proper type in this file and set these pointers to be properly typed.
Please change the other callbacks similarly.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -850,6 +863,52 @@ void CC32xxWiFi::sock_event_handler(void *context)
...
+        case SL_NETAPP_HTTPGETTOKENVALUE_EVENT:
+        {
+            unsigned char *ptr;
+
+            ptr = response->ResponseData.token_value.data;
+            response->ResponseData.token_value.len = 0;
+
+            for (unsigned i = 0; i < httpGetCallbacks_.size(); ++i)
+            {
+                if (memcmp(event->EventData.httpTokenName.data,
```
balazsracz: Are you purposefully allowing prefix matches on tokens? I think that's a good idea, but then we also need to send the actual token value to the callback function.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -850,6 +863,52 @@ void CC32xxWiFi::sock_event_handler(void *context)
...
+                           httpGetCallbacks_[i].second,
+                           strlen(httpGetCallbacks_[i].second)) == 0)
+                {
+                    string result = httpGetCallbacks_[i].first();
+                    memcpy(ptr, result.c_str(), result.length());
```
balazsracz: I've never used string::length() -- i always use size() for c++ container length. (We also use size() everywhere else in the codebase)

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
bakerstu: The intent is to support exact match only.  I've updated the code to reflect this.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -887,14 +884,18 @@ void CC32xxWiFi::http_server_callback(void *context1, void *context2)
...
-                    ptr += result.length();
-                    response->ResponseData.token_value.len += result.length();
+                    // clip string if required
+                    if (result.size() >= MAX_TOKEN_VALUE_LEN)
+                    {
+                        result.erase(MAX_TOKEN_VALUE_LEN);
+                    }
+                    memcpy(ptr, result.c_str(), result.size());
+                    ptr += result.size();
+                    response->ResponseData.token_value.len += result.size();
```
balazsracz: is there a reason for this to be += instead of = ?

src/freertos_drivers/net_cc32xx/CC32xxWiFi.hxx
```diff
@@ -209,38 +224,79 @@ public:
...
+     *
+     * Additional documentation on the CC32xx HTTP Web Server can be found in
+     * the
+     * <a href="http://www.ti.com/lit/ug/swru368a/swru368a.pdf">
+     * CC3100/CC3200 SimpleLink Wi-Fi Internet-on-a-Chip User's Guide</a>
+     *
      * @param callback the std::pair<> of the function to execute and the
-     *        matching token to execute the callback on
+     *        matching token to execute the callback on.  The second (const
+     *        char *) argument of the std::pair must live for as long as the
```
balazsracz: must live forever.  (there is no way to unregister a token)

PR #58, opened 2017-04-22
Bakerstu simplelink debug

bakerstu:


include/freertos/FreeRTOSConfig.h
```diff
@@ -100,6 +100,7 @@
 /* !!!! configMAX_SYSCALL_INTERRUPT_PRIORITY must not be set to zero !!!!
 See http://www.FreeRTOS.org/RTOS-Cortex-M3-M4.html. */
 #define configMAX_SYSCALL_INTERRUPT_PRIORITY     191 /* equivalent to 0xa0, or priority 5. */
+#define configASSERT( x ) if (!(x)) diewith(BLINK_DIE_ASSERT)
```
balazsracz: Do you want this to be in our production releases?
Consider filing a bug (at least in our production collaborators' issue tracker) to make sure this is optimized before release.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -339,15 +339,38 @@ void CC32xxWiFi::wlan_mac(uint8_t mac[6])
     sl_NetCfgGet(SL_MAC_ADDRESS_GET, nullptr, &len, mac);
 }
 
+void *vSimpleLinkSpawnTask(void *pvParameters)
```
balazsracz: What's the difference between this implementation and VStartSimplelinkSpawnTask ?

src/os/os.c
```diff
@@ -135,7 +135,7 @@ typedef struct task_list
 } TaskList;
 
 /** List of all the tasks in the system */
-static TaskList taskList;
+static TaskList *taskList;
```
balazsracz: = NULL;
(the previous decl was default constructed, this is not as declared here)

src/os/os.c
```diff
@@ -313,17 +313,32 @@ static void os_thread_start(void *arg)
     _impure_ptr = priv->reent;
     (*priv->entry)(priv->arg);
```
balazsracz: // Cleanup after the task has exited.

src/os/os.c
```diff
@@ -313,17 +313,32 @@ static void os_thread_start(void *arg)
...
 
-    TaskList *current = &taskList;
-    TaskList *last = &taskList;
     vTaskSuspendAll();
-    while (current->task != xTaskGetCurrentTaskHandle())
+    TaskList *current = taskList;
```
balazsracz: you can write this whole code simpler by using an iteration variable like this:
TaskList** prev_ptr = &taskList;
 Then you don't need special case for single, first and last entries.

src/os/os.c
```diff
@@ -381,8 +396,30 @@ void vApplicationGetTimerTaskMemory(StaticTask_t **pxTimerTaskTCBBuffer,
...
+    task_new->next = NULL;
+    vTaskSuspendAll();
+    if (taskList == NULL)
+    {
+        taskList = task_new;
+    }
+    else
+    {
+        TaskList *current = taskList;
+        while (current->next != NULL)
```
balazsracz: what's the reason for needing to add the new task to the end of the task list? Why can't we add it to the beginning? It would be much simpler.

src/os/os.c
```diff
@@ -815,17 +842,18 @@ void hw_idle_hook(void)
...
-    vTaskSuspendAll();
-    xTaskResumeAll();
     hw_idle_hook();
-    for (TaskList *tl = &taskList; tl != NULL; tl = tl->next)
+    vTaskSuspendAll();
+    for (TaskList *tl = taskList; tl != NULL; tl = tl->next)
     {
         if (tl->task)
         {
-            tl->name = (char*)pcTaskGetTaskName(tl->task);
```
balazsracz: where did this go?

src/os/os.c
```diff
@@ -856,19 +884,19 @@ void main_thread(void *arg)
...
+    ++task_new;
 #else
-    taskList.next = malloc(sizeof(TaskList));
-    taskList.next->task = xTaskGetIdleTaskHandle();
-    taskList.next->unused = uxTaskGetStackHighWaterMark(taskList.next->task);
-    taskList.next->next = NULL;
+    TaskList *task_new = malloc(sizeof(TaskList));
 #endif
+    task_new->task = xTaskGetIdleTaskHandle();
+    task_new->name = (char*)pcTaskGetTaskName(task_new->task);
```
balazsracz: There is a bit of code here that's repeated three times, maybe four. Could this be moved inside the helper function maybe?

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
bakerstu: This way the simplelink task gets logged into the tasklist.

src/os/os.c
bakerstu: This is redundant.  The name is assigned when the task is created.

PR #60, opened 2017-04-24
Fixes for the bootloader

balazsracz:
Adds a facility to have the bootloader hand off the Node ID to the application.
Adds verify node ID support for the bootloader (both global and addressed).

Minor fixes for the bracz-railcom board.


bakerstu:
LGTM

PR #61, opened 2017-05-12
Bakerstu simplelink wifi socket close cleanup

bakerstu:
This fixes the socket leaking.  It might also fix the FreeRTOS list issue.

### DIFF
--- a/src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
+++ b/src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
@@ -222,7 +222,7 @@ int CC32xxSocket::bind(int socket, const struct sockaddr *address,
     CC32xxSocket *s = get_instance_from_fd(socket);
     if (s == nullptr)
     {
-        return - 1;
+        return -1;
     }
 
     SlSockAddr_t sl_address;
@@ -252,7 +252,7 @@ int CC32xxSocket::listen(int socket, int backlog)
     CC32xxSocket *s = get_instance_from_fd(socket);
     if (s == nullptr)
     {
-        return - 1;
+        return -1;
     }
 
     int result = sl_Listen(s->sd, backlog);
@@ -282,7 +282,7 @@ int CC32xxSocket::accept(int socket, struct sockaddr *address,
     CC32xxSocket *s = get_instance_from_fd(socket);
     if (s == nullptr)
     {
-        return - 1;
+        return -1;
     }
 
     if (!s->listenActive)
@@ -374,7 +374,7 @@ int CC32xxSocket::connect(int socket, const struct sockaddr *address,
     CC32xxSocket *s = get_instance_from_fd(socket);
     if (s == nullptr)
     {
-        return - 1;
+        return -1;
     }
 
     SlSockAddr_t sl_address;
@@ -428,7 +428,7 @@ ssize_t CC32xxSocket::recv(int socket, void *buffer, size_t length, int flags)
     CC32xxSocket *s = get_instance_from_fd(socket);
     if (s == nullptr)
     {
-        return - 1;
+        return -1;
     }
 
     int result = sl_Recv(s->sd, buffer, length, flags);
@@ -475,7 +475,7 @@ ssize_t CC32xxSocket::send(int socket, const void *buffer, size_t length, int fl
     CC32xxSocket *s = get_instance_from_fd(socket);
     if (s == nullptr)
     {
-        return - 1;
+        return -1;
     }
 
     int result = sl_Send(s->sd, buffer, length, flags);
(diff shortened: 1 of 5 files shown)


src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
```diff
@@ -222,7 +222,7 @@ int CC32xxSocket::bind(int socket, const struct sockaddr *address,
     CC32xxSocket *s = get_instance_from_fd(socket);
     if (s == nullptr)
     {
-        return - 1;
+        return -1;
```
balazsracz: shouldn't we have something errno = EBADF here (and all other cases like this)

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -709,6 +679,19 @@ void CC32xxWiFi::run_on_network_thread(std::function<void()> callback)
 }
 
 /*
+ * CC32xxWiFi::fd_remove()
+ */
+void CC32xxWiFi::fd_remove(int16_t socket)
```
balazsracz: please rename the argument from socket to sd
this is to avoid confusion of different address spaces: fd is in our abstraction, but sd is in the simplelink abstraction.
`socket` is not sufficiently specific, because bind(), listen() etc have arguments called `socket` but they are in fact in our fd array.

Consider changing the arguments to other functions similar to this (e.g. fd_set_read, add_socket, del_socket, etc)

balazsracz: Please add documentation to the slSockets array. what is it indexed by? fd? sd? what is the value stored? fd? sd?

balazsracz: del_sockets calls the new_highest() function but the new_highest() function will never reduce fdHighest. This is a bug / performance problem -- the code does not crash as is but then the new_highest() call is unnecessary.

balazsracz: fdHighest should be renamed to sdHighest ?

src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
```diff
@@ -679,8 +680,14 @@ int CC32xxSocket::close(File *file)
     if (--references_ == 0)
     {
         mutex.unlock();
-        /* request that the socket be closed */
-        CC32xxWiFi::instance()->select_wakeup(s->sd);
+        /* request that the socket be removed from managment */
+        int16_t sd = s->sd;
+        CC32xxWiFi::instance()->fd_remove(sd);
+        portENTER_CRITICAL();
+        remove_instance_from_sd(sd);
```
balazsracz: There are five different delete functions by now:

- del_socket
- fd_remove
- remove_instance_from_sd
- Socket::close()
- ~CC32xxSocket()

which are calling each other in hard to understand patterns that are not indicated by their respective names and namespaces. This is usually a sign that the code has organically evolved and it would be time to refactor it, specifically to find the appropriate concepts, encapsulation and naming scheme.

This is not something you should do in this PR, but file a bug and get to it whenever you have time.

If you'd like we could go over the use-cases in a chat to find the appropriate encapsulation concepts that can make the code more readable.

src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
```diff
@@ -326,6 +326,7 @@ int CC32xxSocket::accept(int socket, struct sockaddr *address,
     int reserved = reserve_socket();
     if (reserved < 0)
     {
+        sl_Close(result);
```
balazsracz: FYI: bugs like this can usually be avoided by using RAII classes like unique_ptr.
Any code where you have the same command repeated more than once before intermediate "return" statements should be a candidate considered for refactoring into an RAII class. With c++11 features like lambda functions this should be really painless.

src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
bakerstu: Issue filed:

https://github.com/bakerstu/openmrn/issues/63

src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
bakerstu: get_instacne_from_fd() sets the errno appropriately.  A clarifying comment has been added to the code.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
bakerstu: renamed socket argument to sd.

bakerstu: Updated documentation.

bakerstu: Fixed.

bakerstu: I disagree only because it is inconsistent with the "fd_set" naming convention.

PR #62, opened 2017-05-12
Bracz cc3220 port

bakerstu:


src/freertos_drivers/ti/CC32xxUart.cxx
```diff
@@ -113,28 +119,55 @@ void CC32xxUart::disable()
...
+    {
+        if (txEnableAssert)
+        {
+            txEnableAssert();
+        }
+
+        send();
+        txPending = true;
+
+        MAP_IntDisable(interrupt);
```
balazsracz: I'm not quite sure why are we disabling the interrupt here if it will be enabled shortly.
tx_char is invoked inside a critical section so this should have no effect at all.

src/freertos_drivers/ti/CC32xxUart.cxx
```diff
@@ -113,28 +119,55 @@ void CC32xxUart::disable()
...
+        /* no more data left to send */
+        MAP_UARTTxIntModeSet(base, UART_TXINT_MODE_EOT);
+    }
+}
+
+/** Try and transmit a message.
+ */
+void CC32xxUart::tx_char()
+{
+    if (txPending == false)
```
balazsracz: we probably should have an else { MAP_UARTTxIntModeSet(base, UART_TXINT_MODE_FIFO); } for this if.

src/freertos_drivers/ti/CC32xxUart.cxx
```diff
@@ -159,50 +192,30 @@ void CC32xxUart::interrupt_handler()
...
-        while (MAP_UARTSpaceAvail(base))
+        if (txBuf->pending())
+        {
+            send();
+            txBuf->signal_condition_from_isr();
+        }
+        else
         {
-#if 0           
             if (MAP_UARTTxIntModeGet(base) == UART_TXINT_MODE_EOT)
```
balazsracz: why would this condition be false?

it seems there are a couple of internal assumptions for the uart driver, I wonder if we could collect those into a consistency check (a set of assertions basically). It might help finding the cause of crashes later.

src/os/OS.hxx
```diff
@@ -484,12 +484,32 @@ public:
...
+        /// @todo need to fill in the Real Time Clock infrastructure
+        return get_monotonic() + rtcOffset;
+    }
+
+    /** Set the current RTC time.
+     */
+    static void set_realtime(long long time)
+    {
+        /// @todo need to fill in the Real Time Clock infrastructure
+        rtcOffset = time - get_monotonic();
```
balazsracz: this is a pretty elegant solution, I like it.

src/os/OS.hxx
```diff
@@ -484,12 +484,32 @@ public:
...
+    /** Get the current RTC time.
+     * @return time in nanoseconds of the RTC time.
+     */
+    static long long get_realtime()
+    {
+        /// @todo need to fill in the Real Time Clock infrastructure
+        return get_monotonic() + rtcOffset;
+    }
+
+    /** Set the current RTC time.
```
balazsracz: +     * @param time current RTC time in nanoseconds (since epoch).

src/freertos_drivers/ti/CC32xxUart.cxx
bakerstu: I agree, interrupt lock removed.

src/freertos_drivers/ti/CC32xxUart.cxx
bakerstu: I agree that this conditional is always true.  It has been removed and replaced with an assert.

src/os/OS.hxx
bakerstu: Yes, I'm considering methods to adjust the rtc in small increments without ever going "backwards" in time to remain in synch with a "master" clock.  This would be similar to PTPd which uses a simple PID control look to stretch or shrink the clock period as needed.  For now, this simple implementation provides a stable API that applications can begin using.

src/freertos_drivers/ti/CC32xxUart.cxx
bakerstu: This is not required because the interrupt mode will be set appropriately the next time the CC32xxUart::send() method is called from the interrupt handler.

PR #68, opened 2017-06-18
Bakerstu dcc accessory client

bakerstu:
I'm not ready to merge this back, but I would like you to take a quick look to see if I'm on the right track at least.  I'm interested in your feedback.  My next step is to write some unit tests.

### DIFF
--- a/src/openlcb/DccAccyProducer.hxx
+++ b/src/openlcb/DccAccyProducer.hxx
@@ -66,6 +66,13 @@ class DccAccyProducer : public CallableFlow<DccAccyProducerInput>,
                         protected BitRangeNonAuthoritativeEventP
 {
 public:
+    /// DCC accessory activation values
+    enum
+    {
+        NORMAL = false, ///< normal route
+        REVERSE = true, ///< reverse route
+    };
+
     /// highest possible DCC address supported
     static constexpr uint16_t MAX_ADDRESS = 2040;
 
@@ -77,7 +84,6 @@ public:
         : CallableFlow<DccAccyProducerInput>(node->iface())
         , BitRangeNonAuthoritativeEventP(node,
                         TractionDefs::ACTIVATE_BASIC_DCC_ACCESSORY_EVENT_BASE,
-                        TractionDefs::INACTIVATE_BASIC_DCC_ACCESSORY_EVENT_BASE,
                         MAX_ADDRESS,
                         std::bind(&DccAccyProducer::state_callback, this,
                                   std::placeholders::_1, std::placeholders::_2))
@@ -117,7 +123,7 @@ private:
     ///         freed.
     Action send_query()
     {
-        send_query_consumer(input()->address - 1, &writer_, &input()->done);
+        send_query_consumer(input()->address + 4, &writer_, &input()->done);
         return wait_and_return_ok();
     }
 
@@ -128,7 +134,7 @@ private:
     ///         freed.
     Action set()
     {
-        BitRangeNonAuthoritativeEventP::set(input()->address - 1,
+        BitRangeNonAuthoritativeEventP::set(input()->address + 3,
                                             input()->value, &writer_,
                                             &input()->done);
         return wait_and_return_ok();
@@ -150,10 +156,10 @@ private:
     /// @param value value of the event pair
     void state_callback(unsigned bit, bool value)
     {
-        if (dccStateCallback_)
+        if (dccStateCallback_ && bit >= 4)
         {
-            // add one to bit because the DCC address range starts at 1, not 0
-            dccStateCallback_(bit + 1, value);
+            // add four to bit because the DCC address range starts at 1, not 0
+            dccStateCallback_(bit - 3, value);
         }
     }
 
--- a/src/openlcb/DccAccyProducer.hxx
+++ b/src/openlcb/DccAccyProducer.hxx
@@ -83,10 +83,10 @@ public:
                std::function<void(unsigned, bool)> dcc_state_callback = nullptr)
         : CallableFlow<DccAccyProducerInput>(node->iface())
         , BitRangeNonAuthoritativeEventP(node,
-                        TractionDefs::ACTIVATE_BASIC_DCC_ACCESSORY_EVENT_BASE,
-                        MAX_ADDRESS,
-                        std::bind(&DccAccyProducer::state_callback, this,
-                                  std::placeholders::_1, std::placeholders::_2))
+                      TractionDefs::ACTIVATE_BASIC_DCC_ACCESSORY_EVENT_BASE + 8,
+                      MAX_ADDRESS,
+                      std::bind(&DccAccyProducer::state_callback, this,
+                                std::placeholders::_1, std::placeholders::_2))
         , dccStateCallback_(dcc_state_callback)
         , writer_()
     {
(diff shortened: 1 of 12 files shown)


src/openlcb/NonAuthoritativeEventProducer.cxxtest
```diff
@@ -0,0 +1,114 @@
...
+#include "NonAuthoritativeEventProducer.hxx"
+
+using ::testing::_;
+using ::testing::Mock;
+using ::testing::StrictMock;
+
+namespace openlcb
+{
+
+static const uint64_t kEventBase = 0x05010101FFFF0000ULL;
```
balazsracz: 0501010118FF0000
(just so that it is in our range)

src/openlcb/NonAuthoritativeEventProducer.cxxtest
```diff
@@ -0,0 +1,114 @@
...
+
+    // dual range
+    expect_packet(":X1952422AN05010101FFFF004F;");
+    expect_packet(":X1952422AN05010101FFFF006F;");
+
+    send_packet(":X19970001N;");
+    wait_for_event_thread(); Mock::VerifyAndClear(&canBus_);
+}
+
+TEST_F(BitRangeNonAuthoritativeEventPTest, GlobalIdentifyAddressed)
```
balazsracz: alkso add a negative example, when no packets are expected.

src/openlcb/NonAuthoritativeEventProducer.cxxtest
```diff
@@ -0,0 +1,114 @@
...
+
+    // dual range
+    expect_packet(":X1952422AN05010101FFFF004F;");
+    expect_packet(":X1952422AN05010101FFFF006F;");
+
+    send_packet(":X19970001N022A;");
+    wait_for_event_thread(); Mock::VerifyAndClear(&canBus_);
+}
+
+TEST_F(BitRangeNonAuthoritativeEventPTest, IdentifyProducer)
```
balazsracz: do you have any tests for queries?

src/os/MDNS.cxx
```diff
@@ -41,6 +41,7 @@
 void mdns_publish(const char *name, const char *service, uint16_t port) __attribute__ ((weak));
 int mdns_lookup(const char *service, struct addrinfo *hints,
                 struct addrinfo **addr) __attribute__ ((weak));
+void mdns_scan(const char *service) __attribute__ ((weak));
```
balazsracz: these changes probably belong to a different PR

src/openlcb/DccAccyProducer.hxx
```diff
@@ -0,0 +1,177 @@
...
+public:
+    /// DCC accessory activation values
+    enum
+    {
+        NORMAL = false, ///< normal route
+        REVERSE = true, ///< reverse route
+    };
+
+    /// highest possible DCC address supported
+    static constexpr uint16_t MAX_ADDRESS = 2040;
```
bakerstu: I've contemplated that as well.  Not sure if we want to support broadcast, or perhaps even make it an option.

src/openlcb/NonAuthoritativeEventProducer.cxx
```diff
@@ -0,0 +1,258 @@
...
+            range = EncodeRange(eventBase_,
+                                eventBaseOff_ == 0 ? size_ * 2 : size_);
+        }
+        else if (entry.event == eventBaseOff_ && eventBaseOff_ != 0)
+        {
+            range = EncodeRange(eventBaseOff_, size_);
+        }
+        else
+        {
+            // uninteresting event range
```
bakerstu: Is that true even if my event range is not an exact power of 2?

src/openlcb/NonAuthoritativeEventProducer.cxxtest
bakerstu: Yeah, I wanted to do this.  Howe do I test for that?

src/openlcb/NonAuthoritativeEventProducer.hxx
```diff
@@ -0,0 +1,207 @@
...
+ */
+
+#ifndef _OPENLCB_NONAUTHORITATIVEEVENTPRODUCER_HXX_
+#define _OPENLCB_NONAUTHORITATIVEEVENTPRODUCER_HXX_
+
+#include "openlcb/EventHandlerTemplates.hxx"
+
+namespace openlcb
+{
+#if 0
```
bakerstu: That is the intent.  It was copied here as a quick reference.

src/openlcb/NonAuthoritativeEventProducer.hxx
```diff
@@ -0,0 +1,207 @@
...
+public:
+    /// Constructor.  Creates a new bit range producer.
+    ///
+    /// @param node the node that the producer will be bound to
+    /// @param event_base the starting event ID for the event range
+    /// @param size The total event range size in "pairs" of events.  Each
+    ///             pair of sequential events represents a single "bit" with
+    ///             a binary state.
+    /// @param state_callback Callback method for delivering the results of a
+    ///                       consumer identified.  The first unsigned parameter
```
bakerstu: The code already calls the same callback for both "event report" and "consumer identified".  The comment should be updated accordingly.

src/openlcb/NonAuthoritativeEventProducer.hxx
```diff
@@ -0,0 +1,207 @@
...
+    /// Handle an incoming consumer identified message.
+    ///
+    /// @param entry reference to this entry in the event registry
+    /// @param event event metadata
+    /// @param done notifible to wakup when finished
+    void handle_consumer_identified(const EventRegistryEntry &entry,
+                                    EventReport *event,
+                                    BarrierNotifiable *done) override;
+
+    /// Handle an incoming identify producer message.
```
bakerstu: fixed

bakerstu: Actually, is it not both?  Which hander gets called for an Identify Events Addressed?

src/os/MDNS.cxx
bakerstu: I thought I put those on another branch.  I'll have to look into this further.

src/openlcb/NonAuthoritativeEventProducer.cxx
balazsracz: Actually you're right, there is a problem when the base is not aligned. The big issue is that entry.event == eventBase_ will not be true at all in that case. The registered event is already rounded up to the power of two boundary, while eventBase is not.

If we do the encoding of which eventbase we used in the userdata uint32_t in the Entry, then we can use that to figure out which one to report here.

balazsracz: Also we need to add tests for cases when the base is not aligned.

src/openlcb/NonAuthoritativeEventProducer.cxxtest
balazsracz: send_packet(":X19970001N022B;");
wait_for_event_thread(); Mock::VerifyAndClear(&canBus_);

src/openlcb/NonAuthoritativeEventProducer.hxx
balazsracz: yes it is both. Hence the documentation should say "identify global and addressed" [/ global or addressed?], sorry for not being clear.

src/openlcb/NonAuthoritativeEventProducer.cxx
bakerstu: Is this not also a problem on the top end as well?  For example, if the range ends before a power of 2 boundary, I should not respond to a request beyond the end of my range.

PR #71, opened 2017-06-22
Bakerstu dcc accessory client

bakerstu:
I've should now have full test coverage and I'm ready to start trying this out in the NCE throttle implementation.  Please review.  JohnSL, please have a look at the abstract API in class DccAccyProducer to see if it matches your expectations for WiThrottle.


src/os/MDNS.cxx
```diff
@@ -41,6 +41,7 @@
 void mdns_publish(const char *name, const char *service, uint16_t port) __attribute__ ((weak));
```
bakerstu: These commits to MDNS.cxx, MDNS.hxx, and SocketClient.hxx do not belong in this branch.  I may need some Git help to move them over to the proper branch "bakerstu-simplelink-mdns".

src/openlcb/DccAccyProducer.hxx
```diff
@@ -0,0 +1,250 @@
...
+            if (instance->dccStateCallback_)
+            {
+                instance->dccStateCallback_(bit + 1, value);
+            }
+        }
+    }
+
+    /// Singleton instance of a BitRangeNonAuthoritativeEventP for the DCC
+    /// accessory address Well-Known Event ID range.
+    static BitRangeNonAuthoritativeEventP *eventProducer_;
```
bakerstu: Can this member be statically constructed in place rather than using operator new at runtime?  This would save a malloc().  Is there a construction order problem with this?

src/openlcb/DccAccyProducer.cxxtest
```diff
@@ -0,0 +1,274 @@
...
+            {
+                b1->unref();
+            }
+            if (b2)
+            {
+                b2->unref();
+            }
+        }
+
+        DccAccyProducerInput *response1()
```
balazsracz: did you know that response1 and response2 will always return the same data?

full_allocation_result only takes the argument as a type parameter, and always returns allocationResult_ cast to that type, which only exists once.

src/openlcb/DccAccyProducer.cxxtest
```diff
@@ -0,0 +1,274 @@
...
+    };
+
+};
+
+TEST_F(DccAccyProducerTest, SendSet)
+{
+    wait();
+    clear_expect();
+
+    class TestFlow : public ImplFlow
```
balazsracz: If you don't want to do this explicit but complicated way of testing, there is a helper function for invoking subflows that you can use: https://github.com/bakerstu/openmrn/blob/master/src/utils/async_if_test_helper.hxx#L288

That you can run directly inside the TEST_F function body and will handle the invocation, the wait for return, and the ownership of the returned pointer.

Look at openlcb/TractionThrottle.cxxtest:254 for examples.

PR #76, opened 2017-07-02
Locally build Tivaware libraries

balazsracz:
Builds Tivaware library locally in order to avoid potential conflicts such as hard/soft floating point.

tivadriverlib is a replacement for the Tivaware driverlib

tivausblib, tivausblib_device, and tivausblib_host are the replacement for the Tivaware usblib


targets/freertos.armv7m/tivadriverlib/Makefile
```diff
@@ -0,0 +1,3 @@
+OPENMRNPATH ?= $(realpath ../../..)
+include $(OPENMRNPATH)/etc/lib.mk
```
balazsracz: Please move these new library directories under targets/freertos.armv7m/freertos_drivers/
This is to match in consistency with the other middlewares that we are building there (e.g. lpc_chip_175x_6x, stm32cubef103xb, stm32cubef303x_28x_58x_98x etc).

targets/freertos.armv7m/tivadriverlib/sources
```diff
@@ -0,0 +1,15 @@
+include $(OPENMRNPATH)/etc/tivaware.mk
+DEPS  += TIVAWAREPATH
+VPATH := $(TIVAWAREPATH)/driverlib
+
+CFLAGS += -Dcodered
+
+FULLPATHCSRCS = $(wildcard $(VPATH)/*.c)
```
balazsracz: It seems you could get away without duplicating these lines if you remove the sources file, and move the above few lines into Makefile of the current directory. 
It might be possible that a small change is needed to etc/lib.mk along the lines of 

ifndef VPATH
VPATH = $(SRCDIR)
endif

and you may have to play with the order of the variable set vs include.

PR #78, opened 2017-07-07
Listener feature for the traction throttle

balazsracz:
Adds a feature to the Traction Throttle class that registers the throttle node to a virtual consist with the controlled train node. This way if another controller changes the train status (speed, function, etc) the current throttle gets the message forwarded. Allows updating the current throttle's display with the changed state of the train.


balazsracz:
@bakerstu : LMK if you don't have time to review -- this is not something that desperately needs feedback.
@JohnSL : this is FYI for you as an example of state flow code and our unittesting practices.

bakerstu:
@balazsracz : I'm very interested in this feature, and it is quite timely for some of the ongoing NCE bus work.  I'll review this tonight.

PR #80, opened 2017-08-16
Train simulator: fixed the address argument and added a name argument.

JohnSL:


applications/train/targets/linux.x86/main.cxx
```diff
@@ -69,6 +67,7 @@ using openlcb::WriteHelper;
 int port = 12021;
 const char *host = "localhost";
 const char *device_path = nullptr;
+const char *name = "Deadrail Train";
 int address = 1726;
```
balazsracz: let's change this to 1732 to not modify the default behavior.

PR #87, opened 2017-09-11
SocketClient connection status callback

JohnSL:
Adding the ability to report status back while attempting to connect to a socket.

### DIFF
--- a/src/utils/SocketClient.hxx
+++ b/src/utils/SocketClient.hxx
@@ -46,6 +46,15 @@
 #include "utils/Atomic.hxx"
 #include "utils/format_utils.hxx"
 
+enum class SocketStatus
+{
+    MDNS_LOOKUP,
+    MDNS_CONNECT,
+    STATIC_CONNECT,
+    CONNECTED,
+    CONNECT_FAILED,
+};
+
 class SocketClient : public StateFlowBase, private OSThread, private Atomic
 {
 public:
@@ -75,13 +84,15 @@ public:
     SocketClient(Service *service, const char *mdns, const char *host,
                  uint16_t port,
                  std::function<void(int, struct addrinfo *, Notifiable*)>callback,
-                 uint8_t retry_seconds = 5, uint8_t timeout_seconds = 255)
+                 uint8_t retry_seconds = 5, uint8_t timeout_seconds = 255,
+                 std::function<void(SocketStatus)>status = nullptr)
         : StateFlowBase(service)
         , OSThread()
         , mdns_(mdns)
         , host_(host)
         , port_(port)
         , callback_(callback)
+        , statusCallback_(status)
         , retrySeconds_(retry_seconds)
         , timeoutSeconds_(timeout_seconds)
         , state_(STATE_CREATED)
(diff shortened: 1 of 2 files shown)


src/utils/SocketClient.hxx
```diff
@@ -46,6 +46,19 @@
 #include "utils/Atomic.hxx"
 #include "utils/format_utils.hxx"
 
+/** Connection status that can be sent back to the "owner" of the socket so
+ * it can update display or status information while the connection attempts
+ * are progressing.
+ */
+enum class SocketStatus
```
balazsracz: please move this inside SocketClient and call it Status.
(The reason is that we are in the global namespace and we should be careful about namespace pollution)

src/utils/SocketClient.hxx
```diff
@@ -71,17 +84,21 @@ public:
...
      *                        timeout and look for a possible shutdown.
+     * @param status callback method for status as the connection attempt
+     *               progresses
      */
     SocketClient(Service *service, const char *mdns, const char *host,
                  uint16_t port,
                  std::function<void(int, struct addrinfo *, Notifiable*)>callback,
-                 uint8_t retry_seconds = 5, uint8_t timeout_seconds = 255)
+                 uint8_t retry_seconds = 5, uint8_t timeout_seconds = 255,
+                 std::function<void(SocketStatus)>status = nullptr)
```
balazsracz: please move the status callback before the integer arguments. Consider renaming it to status_callback.

src/utils/SocketClient.hxx
```diff
@@ -71,17 +84,21 @@ public:
      *                      connecting on error.
      * @param timeout_seconds time in seconds that the connect is supposed to
      *                        timeout and look for a possible shutdown.
+     * @param status callback method for status as the connection attempt
+     *               progresses
```
balazsracz: Note that the callback will be invoked on a different thread than the service executor.

src/utils/SocketClient.hxx
```diff
@@ -154,8 +171,17 @@ private:
         STATE_SHUTDOWN, /**< shutdown */
     };
 
-    /** thread that will handle the blocking address resolution.
-     * @return should never return
+    /** Called from a new thread that is created by this workflow when it calls
```
balazsracz:
```
Entry point to the thread; this thread performs the synchronous network calls (address resolution and connect). The function returns only when the thread needs to be terminated (i.e. after shutdown() is invoked).

When this method [...]
```

src/utils/SocketClient.hxx
balazsracz: there is a space missing after > (also in the callback line)

src/utils/SocketClient.hxx
```diff
@@ -46,6 +46,19 @@
...
+/** Connection status that can be sent back to the "owner" of the socket so
+ * it can update display or status information while the connection attempts
+ * are progressing.
+ */
+enum class SocketStatus
+{
+    MDNS_LOOKUP,
+    MDNS_CONNECT,
+    STATIC_CONNECT,
+    CONNECTED,
```
bakerstu: I don't see this "CONNECTED" state posted anywhere.  Did I miss something?

balazsracz: i think that gets implicit when the connection callback is invoked. It could probably be removed or we can add a (redundant) status update call, just before the connection callback is invoked.

PR #88, opened 2017-09-11
prevent multiple datagrams from interleaving

balazsracz:
Checks in the datagram sender whether there is already a pending datagram for a given src/dst pair, and waits until that datagram gets ack/nack-ed before sending out a second datagram.
This is necessary for standards compliance and also needed for performing a datagram based interaction with localhost (e.g. memory config read).
Misc fixes:
- Makes the MemoryConfigHandler be able to hand off reply datagrams to a MemoryConfigClient even while the server flow is busy. This is needed for a localhost query to successfully complete.

Tested:
- adds unittest for localhost datagram
- adds unittest for sending two interleaved datagrams to a remote client
- adds unittest for exercising the MemoryConfigClient against localhost to download a memory space stringed into many datagrams.

### DIFF
--- a/src/openlcb/DatagramCan.cxx
+++ b/src/openlcb/DatagramCan.cxx
@@ -52,13 +52,15 @@ long long DATAGRAM_RESPONSE_TIMEOUT_NSEC = SEC_TO_NSEC(3);
 /// The base class of AddressedCanMessageWriteFlow is responsible for the
 /// discovery and address resolution of the destination node.
 class CanDatagramClient : public DatagramClient,
-                          public AddressedCanMessageWriteFlow
+                          public AddressedCanMessageWriteFlow,
+                          public LinkedObject<CanDatagramClient>
 {
 public:
     CanDatagramClient(IfCan *iface)
         : AddressedCanMessageWriteFlow(iface)
         , listener_(this)
         , isSleeping_(0)
+        , sendPending_(0)
     {
         /** This flow does not use the incoming queue that we inherited from
          * AddressedCanMessageWriteFlow. We skip the wait state.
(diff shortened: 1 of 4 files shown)


src/openlcb/DatagramCan.cxx
```diff
@@ -77,13 +79,37 @@ class CanDatagramClient : public DatagramClient,
...
+                if (!c->sendPending_) continue;
+                if (c->nmsg()->src.id != nmsg()->src.id) continue; 
+                if (!async_if()->matching_node(c->nmsg()->dst, nmsg()->dst))
+                    continue;
+                // Now: there is another datagram client sending a datagram to
+                // this destination. We need to wait for that transaction to
+                // complete.
+                c->waitingClients_.push_front(this);
+                return wait();
+            }
```
bakerstu: I wonder if you want to set sendPending_ before you leave the AtomicHolder context.  I don't fully understand the threading model here, but it seems to me that you might have a race condition.

balazsracz: For readability I would want to keep modifications to sendPending_ inside the register_handler and unregister_handler functions. This way it is easy to understand the contract.
The race condition is benign so long as all the datagram clients are running on the same service executor, since then the executor thread will prevent them from running in parallel. The atomic holder here is needed more for the iteration over the linkedobject list, because it is not guaranteed that the datagram clients are also *created* on the main executor thread.
I will try to move the entire register_handlers() call inside the lock and see if that is possible. (There is a limit to what operations we are allowed to do inside a critical section; a mutex acquisition is not allowed. The handler registry might be using mutexes internally, so it's not clear that this is safe.)

bakerstu: I think that threading model makes sense.  I assumed the Atomic is used in order to avoid the overhead of a mutex.  If a mutex is a possibility without adding much overhead (reuse an existing mutex for example), it might be preferred since technically the list size is unbounded.  Though practically speaking, the list size should not get that big.

balazsracz: I checked and it can't be moved inside the atomic, because the dispatcher register_handler uses an OSMutex inside.
In practice this list is exactly two entries long for all freertos targets.

PR #90, opened 2017-09-27
Increasing the level of abstraction for TractionThrottle.

JohnSL:
The change here is to make it easier to create replacements for TractionThrottle.


src/openlcb/TractionThrottle.hxx
```diff
@@ -250,18 +266,20 @@ public:
 
     /// Determine if a train is currently assigned to this trottle.
```
balazsracz: please copy or move the comments describing these functions to the base class.

PR #91, opened 2017-10-02
Add support for storing a train name based on a SNIP reply.

bakerstu:


bakerstu:
This feature has been added elsewhere.

PR #93, opened 2017-10-07
Speed txn tx n plant init

SpeedTxn:
Just for a dry run!


applications/txn_plant/targets/freertos.armv7m.ek-tm4c123gxl/hardware.hxx
```diff
@@ -0,0 +1,55 @@
...
+GPIO_PIN(MCP23017_INT_N, GpioInputPU, A, 5);
+
+typedef GpioInitializer<                          //
+    SW1_Pin, SW2_Pin,                             //
+    LED_RED_Pin, LED_GREEN_Pin, LED_BLUE_RAW_Pin, //
+    USB1_Pin, USB2_Pin,                           //
+    UART0RX_Pin, UART0TX_Pin,                     //
+    BOOSTER_H_Pin, BOOSTER_L_Pin,                 //
+    RAILCOM_TRIGGER_Pin, RAILCOM_CH1_Pin,         //
+    CAN0RX_Pin, CAN0TX_Pin> GpioInit;
```
balazsracz: You should list MCP23017_INT_N_Pin here, otherwise it will not be initialized.

applications/txn_plant/targets/freertos.armv7m.ek-tm4c123gxl/HwInit.cxx
```diff
@@ -0,0 +1,425 @@
...
+
+    // Initializes all GPIO and hardware pins.
+    GpioInit::hw_init();
+
+    /* Setup the system clock. */
+    MAP_SysCtlClockSet(SYSCTL_SYSDIV_2_5 | SYSCTL_USE_PLL | SYSCTL_OSC_MAIN |
+                       SYSCTL_XTAL_16MHZ);
+
+    MAP_SysCtlPeripheralEnable(SYSCTL_PERIPH_I2C1);
+    MAP_GPIOPinConfigure(GPIO_PA6_I2C1SCL);
```
balazsracz: This line and the next four can be replaced by an entry in the gpio pin initializer in hardware.hxx:
see example 
```
GPIO_HWPIN(BOOSTER_H, GpioHwPin, B, 6, T0CCP0, Timer);
```
so you would use something like
```
GPIO_HWPIN(I2C_SCL, GpioHwPin, A, 6, I2C1SCL, I2CSCL);
GPIO_HWPIN(I2C_SDA, GpioHwPin, A, 7, I2C1SDA, I2C);
```
then add I2C_SDA_Pin and I2C_SCL_Pin into the gpioinit array.

The reason is twofold:
- consistency with the other declarations
- it makes sure not to forget any of the pin setup calls. For example here there is a missing call to enable the clock to the gpio peripheral; you were lucky someone else did that earlier but if this was not on port A but say on port C, you would be getting a hard fault here and scratching your head for a while.

applications/txn_plant/targets/freertos.armv7m.ek-tm4c123gxl/main.cxx
```diff
@@ -0,0 +1,278 @@
...
+  /// offset in this file is unspecified, callees must do lseek.
+  void factory_reset(int fd)
+  {
+    for( int i=0; i<NUM_INPUTS;++i) {
+      //char tmp[10];
+      //string str;
+      //unsigned_integer_to_buffer(i, tmp);
+      //cfg.seg().producers().entry(i).description().write(fd,tmp);
+      cfg.seg().producers().entry(i).debounce().write(fd,3);
+      cfg.seg().producers().entry(i).event_on().write(fd,(NODE_ID << 16) + 8+(i*2) );
```
balazsracz: You have two choices about the event IDs here. What you did here is fine if you want a fixed layout of event IDs for your node. This is not recommended for general purpose nodes; in the standard we recommend always new event IDs to be assigned when a factory reset is performed. I have code for that:
```
stack.factory_reset_all_events(cfg.seg().internal_config(), fd);
```
Note you still need the write call for the debounce parameter.

src/freertos_drivers/common/MCP23017GPIO.cxx
```diff
@@ -0,0 +1,177 @@
...
+        if (int_status)
+        {
+            interrupt_lockout = OSTime::get_monotonic() + intLockTime_;
+        }
+
+        /* read remote data and update local data copy*/
+        dataInA_ = register_read(GPIOA);
+        dataInB_ = register_read(GPIOB) << 8;
+
+        if (directionShaddow_ != direction_)
```
balazsracz: typo

src/freertos_drivers/common/MCP23017GPIO.hxx
```diff
@@ -0,0 +1,325 @@
...
+ *
+ * @file MCP23017GPIO.hxx
+ * This file implements a GPIO driver for the MCP23017 I/O expander.
+ *
+ * @author Stuart W. Baker
+ * @date 7 October 2017
+ */
+
+#ifndef _FREERTOS_DRIVERS_COMMON_MCP23017GPIO_HXX_
+#define _FREERTOS_DRIVERS_COMMON_MCP23017GPIO_HXX_
```
balazsracz: this should probably be in a separate PR.

bakerstu: I agree.  If you look closely, there is already another branch for this.

src/freertos_drivers/common/MCP23017GPIO.cxx
bakerstu: I need to perform a little cleanup here.  I also want to support multiple MCP23017's on the same I2C bus with this driver.

PR #100, opened 2017-11-25
64bit compilation fix

balazsracz:
Fixes mismatching feedback key for railcom returns when running on a 64-bit processor.

### DIFF
--- a/src/openlcb/TractionCvSpace.cxx
+++ b/src/openlcb/TractionCvSpace.cxx
@@ -178,7 +178,7 @@ StateFlowBase::Action TractionCvSpace::fill_read1_packet()
         b->data()->add_dcc_address(dcc::DccShortAddress(dccAddress_));
     }
     b->data()->add_dcc_pom_read1(cvNumber_);
-    b->data()->feedback_key = reinterpret_cast<size_t>(this);
+    b->data()->feedback_key = reinterpret_cast<uintptr_t>(this);
     railcomHub_->register_port(this);
     errorCode_ = ERROR_PENDING;
     track_->send(b);
@@ -279,7 +279,7 @@ StateFlowBase::Action TractionCvSpace::fill_write1_packet()
         b->data()->add_dcc_address(dcc::DccShortAddress(dccAddress_));
     }
     b->data()->add_dcc_pom_write1(cvNumber_, cvData_);
-    b->data()->feedback_key = reinterpret_cast<size_t>(this);
+    b->data()->feedback_key = reinterpret_cast<uintptr_t>(this);
     railcomHub_->register_port(this);
     errorCode_ = ERROR_PENDING;
     track_->send(b);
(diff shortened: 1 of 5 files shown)


src/openlcb/TractionCvSpace.cxx
```diff
@@ -330,7 +330,7 @@ void TractionCvSpace::send(Buffer<dcc::RailcomHubData> *b, unsigned priority)
     if (errorCode_ != ERROR_PENDING)
         return;
     const dcc::Feedback &f = *b->data();
-    if (f.feedbackKey != reinterpret_cast<size_t>(this) || f.channel == 0xff)
+    if (f.feedbackKey != (uint32_t)(reinterpret_cast<size_t>(this)) || f.channel == 0xff)
```
bakerstu: It if it is a pointer, why would you not use type uintptr_t?  That would be portable between 32 and 64 bit.

src/openlcb/TractionCvSpace.cxxtest
```diff
@@ -62,9 +62,9 @@ protected:
         wait();
     }
 
-    size_t expected_feedback_key()
+    uint32_t expected_feedback_key()
     {
-        return reinterpret_cast<size_t>(&cv_space_);
+        return (size_t)(&cv_space_);
```
bakerstu: Why not use uintptr_t

src/openlcb/TractionCvSpace.cxx
balazsracz: feedbackKey is uint32_t. This is a member in the packet_t structure which is not internal to our application; this struct is used to communicate with the DCC driver. We are not free to modify the size of this struct or the members.

src/openlcb/TractionCvSpace.cxxtest
balazsracz: here we need to guess what the TractionCvSpace will send as feedback key in order to fabricate  the appropriate returned object from the DCC driver.

src/openlcb/TractionCvSpace.cxx
bakerstu: I'm a bit concerned about this on 64-bit systems.  As soon as you cast 64-bit pointer to a uint32_t, it is no longer guaranteed to be unique.  On a big endian system, it might be less likely to be unique.  Maybe it would be better to change the type of feedbackKey to unit64_t.  What is the reason that feedbackKey cannot be unitptr_t?  Is it because the instance can be passed between host architectures (PRU vs. ARM for example)?

balazsracz: yes it gets passed to the PRU

balazsracz: ok I switched everything to uintptr_t.

src/openlcb/TractionCvSpace.cxxtest
balazsracz: done

PR #102, opened 2017-11-25
Fixes to event registry bugs

balazsracz:
Fixes major bugs in event registry and DccAccyProducer:
- the SortedListMap helper class got into an invalid state after an erase() was called
- the event registry unregister_handler was using an inappropriate algorithm for iterating and erasing entries. This has resulted various side effects; sometimes multiple registrations were not all removed; sometimes the registration of someone else was removed.
- the DccAccyProducer had incorrect locking for the shared static data structures.

Added unit tests for the SortedListMap class and coverage for the unregistration from the event registry.

### DIFF
--- a/src/openlcb/EventHandlerContainer.cxx
+++ b/src/openlcb/EventHandlerContainer.cxx
@@ -34,6 +34,8 @@
 
 #include "openlcb/EventHandlerContainer.hxx"
 
+#include <algorithm>
+
 namespace openlcb
 {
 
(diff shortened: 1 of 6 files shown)


src/openlcb/EventHandlerContainer.cxx
```diff
@@ -49,17 +52,24 @@ void TreeEventHandlers::unregister_handler(EventHandler *handler)
...
+            });
+        if (erase_it != end_it)
+        {
+            r->second.erase(erase_it, end_it);
+            found = true;
         }
     }
-    if (found) return;
+    if (found)
+        return;
```
bakerstu: Add braces around if body.

PR #105, opened 2017-12-28
DCC debugging utils

balazsracz:
Adds debugging utilities for DCC packets and a test helper with gmock matchers.
Makes the user-triggered packet actually have higher repeat count.

### DIFF
--- a/src/dcc/DccDebug.cxx
+++ b/src/dcc/DccDebug.cxx
@@ -1,0 +1,202 @@
+/** \copyright
+ * Copyright (c) 2017, Balazs Racz
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are  permitted provided that the following conditions are met:
+ *
+ *  - Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ *  - Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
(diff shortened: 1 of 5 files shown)


src/dcc/DccDebug.cxx
```diff
@@ -0,0 +1,202 @@
...
+            options += " F[9-12]=";
+        }
+        options += StringPrintf("%d%d%d%d", (cmd >> 0) & 1, (cmd >> 1) & 1,
+            (cmd >> 2) & 1, (cmd >> 3) & 1);
+    }
+    else if ((cmd >> 5) == 0b110)
+    {
+        // expansion
+        uint8_t c = cmd & 0x1F;
+        if ((c & ~1) == 0b11110) {
```
bakerstu: Fix brace style in this section.

src/dcc/DccDebug.cxxtest
```diff
@@ -0,0 +1,157 @@
...
+    Packet pkt;
+    pkt.add_dcc_address(DccShortAddress(3));
+    pkt.add_dcc_function21_28(0b10010110);
+    EXPECT_EQ(
+        "[dcc] Short Address 3 F[21-28]=01101001", packet_to_string(pkt));
+}
+
+
+} // namespace
+} // namespace dcc
```
bakerstu: I don't see any tests for non-DCC protocol.

src/dcc/DccDebug.cxx
```diff
@@ -0,0 +1,202 @@
...
+ * @date 28 Dec 2017
+ */
+
+#include "dcc/DccDebug.hxx"
+#include "utils/StringPrintf.hxx"
+
+namespace dcc
+{
+
+string packet_to_string(const DCCPacket &pkt)
```
bakerstu: Think about how these strings will look scrolling across a terminal.  Does it make sense to put in some alignment so that regardless of the packet type, characters in the address, etc..., the "like fields" line up on vertically?

src/dcc/dcc_test_utils.hxx
```diff
@@ -0,0 +1,62 @@
+#ifndef _DCC_DCC_TEST_UTILS_HXX_
+#define _DCC_DCC_TEST_UTILS_HXX_
+
+#include "dcc/DccDebug.hxx"
+#include "dcc/Packet.hxx"
+
+void PrintTo(const DCCPacket& pkt, ::std::ostream* os) {
```
bakerstu: Fix brace style in this section

src/dcc/DccDebug.cxx
balazsracz: I'm not sure it's worth the effort now. If we get to a point where a huge terminal dump exists, we can come and change the code here.
For the moment I only used this to add more meaningful error messages to a unit test.

src/dcc/DccDebug.cxx
balazsracz: done

src/dcc/DccDebug.cxxtest
balazsracz: I have not written the decoding code yet, that's why there are no tests.

balazsracz: i did add a dummy test though so we can know we recognize them.

src/dcc/dcc_test_utils.hxx
balazsracz: done

PR #109, opened 2017-12-29
Bakerstu mcp2515 io

bakerstu:


src/freertos_drivers/common/MCP2515Can.hxx
```diff
@@ -570,7 +576,11 @@ private:
         ::write(spi, &rts, 1);
     }
 
-    unsigned txPending; /**< transmission in flight */
+    unsigned txPending : 2; /**< transmission in flight */
+    unsigned gpoData   : 2; /**< local copy of the I/O expantion output data */
```
balazsracz: typo (expansion)

src/freertos_drivers/common/MCP2515Can.hxx
```diff
@@ -584,7 +594,141 @@ private:
...
+    }
+
+    /** Sets the GPO pin to high.
+     */
+    void set() const override
+    {
+        portENTER_CRITICAL();
+        instance_->gpoData |= 0x1 << bit_;
+        portEXIT_CRITICAL();
+        instance_->sem.post();
```
balazsracz: i think we should dedup the semaphore notifications when the actual gpoData does not change. These are expensive enough that we don't want to spin on it.

src/freertos_drivers/common/MCP2515Can.hxx
```diff
@@ -584,7 +594,141 @@ private:
...
+    }
+
+    /** Clears the GPO pin to low.
+     */
+    void clr() const override
+    {
+        portENTER_CRITICAL();
+        instance_->gpoData &= ~(0x1 << bit_);
+        portEXIT_CRITICAL();
+        instance_->sem.post();
```
balazsracz: same here

src/freertos_drivers/common/MCP2515Can.hxx
```diff
@@ -584,7 +594,141 @@ private:
...
+    }
+
+    /** Sets the GPO direction (does nothing).
+     * @param dir @ref INPUT or @ref OUTPUT
+     */
+    void set_direction(Gpio::Direction dir) const override
+    {
+    }
```
balazsracz: assert dir == OUTPUT?

src/freertos_drivers/common/MCP2515Can.hxx
```diff
@@ -584,7 +594,141 @@ private:
...
+    {
+        return instance_->gpiData & (0x1 << bit_) ? Gpio::SET : Gpio::CLR;
+    }
+
+    /** Sets the GPI direction (does nothing).
+     * @param dir @ref INPUT or @ref OUTPUT
+     */
+    void set_direction(Gpio::Direction dir) const override
+    {
+    }
```
balazsracz: assert dir == INPUT?

src/freertos_drivers/common/MCP2515Can.hxx
```diff
@@ -584,7 +594,141 @@ private:
...
+    {
+        HASSERT(bit < 3);
+    }
+
+    /** Retrieves the current @ref Value of a GPI input pin.
+     * @return @ref SET if currently high, @ref CLR if currently low
+     */
+    Value read() const override
+    {
+        return instance_->gpiData & (0x1 << bit_) ? Gpio::SET : Gpio::CLR;
```
balazsracz: how about we post the semaphore once for each read? Or if the last read was older than 20 msec, we post?
This would allow to poll exactly at the frequency that the userspace code polls, they will just be always one tick late.
Including never polling if noone reads.
Also if we post the semaphore, and the priorities are appropriately set up, then our current thread will be interrupted and by the time we come back to the CPU the new value will be in gpiData.

src/freertos_drivers/common/MCP2515Can.hxx
balazsracz: I have another question. Can we prevent multiple posting of the semaphore by GPIO drivers?
The thread will spin so long as the semaphore is at nonzero, bu after each GPIO update we need to do only one spin.

balazsracz: we could do this with a dirty bit, for example: when the thread completes the spin, it clears the bit, and we only post the semaphore if the bit is clear.

src/freertos_drivers/common/MCP2515Can.hxx
```diff
@@ -570,7 +577,21 @@ private:
...
+            ioPending = true;
+            sem.post();
+        }
+    }
+
+    unsigned txPending : 2; /**< transmission in flight */
+    unsigned gpoData   : 2; /**< local copy of the I/O expansion output data */
+    unsigned gpiData   : 3; /**< local copy of the I/O expansion input data */
+    unsigned ioPending : 1; /**< true if an I/O update is pending */
+    unsigned reserved  : 24; /**< unused bits */
```
balazsracz: consider removing this line. It doesn't help (alignment is automatic in C) but can cause unnecessary memory usage if someone adds just one bit later without adjusting the count here.

src/freertos_drivers/common/MCP2515Can.hxx
```diff
@@ -570,7 +577,21 @@ private:
         ::write(spi, &rts, 1);
     }
 
-    unsigned txPending; /**< transmission in flight */
+    void request_io_operation()
```
balazsracz: consider: request_gpio_operation.

bakerstu: fixed

src/freertos_drivers/common/MCP2515Can.hxx
bakerstu: fixed

src/freertos_drivers/common/MCP2515Can.hxx
bakerstu: fixed

src/freertos_drivers/common/MCP2515Can.hxx
bakerstu: fixed

src/freertos_drivers/common/MCP2515Can.hxx
bakerstu: fixed

src/freertos_drivers/common/MCP2515Can.hxx
bakerstu: fixed

src/freertos_drivers/common/MCP2515Can.hxx
bakerstu: I added the post, but I did not add the timeout (for now).

src/freertos_drivers/common/MCP2515Can.hxx
bakerstu: fixed

src/freertos_drivers/common/MCP2515Can.hxx
bakerstu: Fixed

PR #111, opened 2018-01-06
Adding a release_train method to the traction interface

JohnSL:


src/openlcb/TractionThrottle.hxx
```diff
@@ -184,6 +184,10 @@ public:
     /// @return the controlled node (the train node) ID.
     /// @todo this function should not be here
     virtual openlcb::NodeID target_node() = 0;
+
+    /// Releases this throttles control over the train
+    /// @dispath is true to dispatch instead of release (WiThrottle specific)
+    virtual void release_train(bool dispath) = 0;
```
balazsracz: The problem is that I cannot implement this correctly for OpenLCB.
Releasing a train is a fundamentally asynchronous operation, and this API is a synchronous API. There has to be time given for the stack to perform the network call, wait for response, handle timeouts and return a status.

All of this is included in the RELEASE_TRAIN api that's just above, line 106. So we'll have to start from that API style.

JohnSL: OK, I'll switch to that approach.

PR #119, opened 2018-02-04
Bugfix for loopback memory config read

balazsracz:
Fixes a bug in MemoryConfig message routing client that prevented addressing localhost when the target space was one of the special space numbers and the command byte had its low bits modified.

### DIFF
--- a/src/openlcb/MemoryConfig.hxx
+++ b/src/openlcb/MemoryConfig.hxx
@@ -511,24 +511,27 @@ public:
         size_t len = message->data()->payload.size();
         const uint8_t *bytes = (const uint8_t *)message->data()->payload.data();
         uint8_t cmd = ((len >= 2) && (client_ != nullptr)) ? bytes[1] : 0;
+        bool is_client_command = false;
+        if ((cmd < 0x80) && (cmd & 0x10))
+        {
+            is_client_command = true;
+        }
         switch (cmd)
         {
-            case MemoryConfigDefs::COMMAND_WRITE_REPLY:
-            case MemoryConfigDefs::COMMAND_WRITE_FAILED:
-            case MemoryConfigDefs::COMMAND_WRITE_STREAM_REPLY:
-            case MemoryConfigDefs::COMMAND_WRITE_STREAM_FAILED:
-            case MemoryConfigDefs::COMMAND_READ_REPLY:
-            case MemoryConfigDefs::COMMAND_READ_FAILED:
             case MemoryConfigDefs::COMMAND_OPTIONS_REPLY:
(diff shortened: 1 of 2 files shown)


src/openlcb/MemoryConfig.hxx
```diff
@@ -511,25 +511,28 @@ public:
         size_t len = message->data()->payload.size();
         const uint8_t *bytes = (const uint8_t *)message->data()->payload.data();
         uint8_t cmd = ((len >= 2) && (client_ != nullptr)) ? bytes[1] : 0;
+        bool is_client_command = false;
+        if ((cmd < 0x80) && (cmd & 0x10))
```
JohnSL: Add enum values for the top and bottom of the client range directly into MemoryConfigDefs for two reasons:
1. So the values here are not magic numbers
2. To help ensure that anyone who might change MemoryConfigDefs knows about the grouping for client commands

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -145,4 +152,22 @@ TEST_F(MemoryConfigLocalClientTest, readfromlocal)
     EXPECT_EQ(0, memcmp(&dataContents_[0], b->data()->payload.data(), dataContents_.size()));
 }
 
+TEST_F(MemoryConfigLocalClientTest, readfromlocalspecialspace)
+{
+    // In this test we read from the local node, to test that the
+    // service-client interaction is without any deadlock. We use a special
+    // space number (0xFD to 0xFF) to regression test a bug in the request
```
JohnSL: What is the bug? It would be good for the comment to describe the bug or provide an issue number.

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -117,6 +117,13 @@ TEST_F(MemoryConfigClientTest, readsmallpart)
                      b->data()->payload.size()));
 }
 
+TEST_F(MemoryConfigClientTest, unsolicited)
```
JohnSL: Please add a comment to describe what required behavior this is testing. I wasn't able to figure that out looking at the test.

src/openlcb/MemoryConfig.hxx
balazsracz: done

src/openlcb/MemoryConfigClient.cxxtest
balazsracz: done

src/openlcb/MemoryConfigClient.cxxtest
balazsracz: There is no bug number. The bug is that responses for those space numbers were not routed to the client. I made a note of this.

PR #120, opened 2018-02-05
MemoryConfigClient: write support

balazsracz:
Adds implementation for Memory Config write commands to the MemoryConfigClient library.

### DIFF
--- a/src/openlcb/MemoryConfigClient.cxxtest
+++ b/src/openlcb/MemoryConfigClient.cxxtest
@@ -117,6 +117,40 @@ TEST_F(MemoryConfigClientTest, readsmallpart)
                      b->data()->payload.size()));
 }
 
+TEST_F(MemoryConfigClientTest, writelarge)
+{
+    expect_any_packet();
+    string test_payload;
+    for (int i = 56; i < 56 + 75; ++i)
+    {
+        test_payload.push_back(i);
+    }
+    EXPECT_NE(0,
+        memcmp(&dataContents_[34], test_payload.data(), test_payload.size()));
+    auto b = invoke_flow(&clientTwo_, MemoryConfigClientRequest::WRITE,
+        NodeHandle(TEST_NODE_ID), 0x51, 34, test_payload);
+    EXPECT_EQ(0, b->data()->resultCode);
+    EXPECT_EQ(0,
+        memcmp(&dataContents_[34], test_payload.data(), test_payload.size()));
+}
+
+TEST_F(MemoryConfigClientTest, writesmall)
+{
+    expect_any_packet();
+    string test_payload;
+    for (int i = 56; i < 56 + 13; ++i)
+    {
+        test_payload.push_back(i);
+    }
+    EXPECT_NE(0,
+        memcmp(&dataContents_[34], test_payload.data(), test_payload.size()));
+    auto b = invoke_flow(&clientTwo_, MemoryConfigClientRequest::WRITE,
+        NodeHandle(TEST_NODE_ID), 0x51, 34, test_payload);
+    EXPECT_EQ(0, b->data()->resultCode);
+    EXPECT_EQ(0,
+        memcmp(&dataContents_[34], test_payload.data(), test_payload.size()));
+}
+
 TEST_F(MemoryConfigClientTest, unsolicited)
 {
     expect_any_packet();
--- a/src/openlcb/MemoryConfigClient.cxxtest
+++ b/src/openlcb/MemoryConfigClient.cxxtest
@@ -117,6 +117,8 @@ TEST_F(MemoryConfigClientTest, readsmallpart)
                      b->data()->payload.size()));
 }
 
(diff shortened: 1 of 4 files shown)


src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -117,6 +117,40 @@ TEST_F(MemoryConfigClientTest, readsmallpart)
                      b->data()->payload.size()));
 }
 
+TEST_F(MemoryConfigClientTest, writelarge)
```
JohnSL: * Where are the magic numbers (56, 75, 34, and 0x51) coming from? Perhaps define named constants that describe what they mean
* Please add a comment to these test describing the required behavior they're validating. It's not at all easy to figure that out from the test itself.

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -270,7 +296,156 @@ private:
...
+        return allocate_and_call(
+            dg_service()->iface()->dispatcher(), STATE(send_write_datagram));
+    }
+
+    Action send_write_datagram()
+    {
+        auto *b = get_allocation_result(dg_service()->iface()->dispatcher());
+        b->set_done(bn_.reset(this));
+        unsigned sz = request()->payload.size() - payloadOffset_;
+        if (sz > 64)
```
JohnSL: What is the meaning of the value 64? Please replace with a named constant

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -270,7 +296,156 @@ private:
...
+            sz = 64;
```
JohnSL: Surround with { and }

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -270,7 +296,156 @@ private:
...
+
+    Action write_response_timeout()
+    {
+        if (responseCode_ & DatagramClient::OPERATION_PENDING)
+        {
+            return handle_write_error(Defs::OPENMRN_TIMEOUT);
+        }
+        size_t len = responsePayload_.size();
+        uint8_t *bytes = (uint8_t *)responsePayload_.data();
+        if (len < 6)
```
JohnSL: Magic number...

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -270,7 +296,156 @@ private:
...
+        {
+            LOG(INFO, "Memory Config client: response datagram payload not "
+                      "long enough");
+            return handle_write_error(
+                Defs::ERROR_INVALID_ARGS_MESSAGE_TOO_SHORT);
+        }
+        unsigned ofs;
+        unsigned a = bytes[2];
```
JohnSL: Does a refer to address? If so, please replace a with address.

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -270,7 +296,156 @@ private:
...
+        a |= bytes[4];
+        a <<= 8;
+        a |= bytes[5];
+        uint8_t space = 0;
+        uint8_t cmd = bytes[1];
+        if (a != offset_)
+        {
+            return handle_write_error(Defs::ERROR_OUT_OF_ORDER);
+        }
+        if (cmd & 3)
```
JohnSL: Magic...

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -270,7 +296,156 @@ private:
...
+            return sleep_and_call(
+                &timer_, SEC_TO_NSEC(3), STATE(write_response_timeout));
+        }
+        else
+        {
+            return call_immediately(STATE(write_response_timeout));
+        }
+    }
+
+    Action write_response_timeout()
```
JohnSL: There seems to be a lot of code repeated here from read_response_timeout(). It would be good to refactor to have methods that handle looking at the bytes. For example, to pull out the address. DRY!

src/openlcb/MemoryConfigClient.cxxtest
balazsracz: Those numbers are arbitrary, chosen for the test, and the consistency ofthose numbers are only within this test function. e.g. we register our space at segment 0x51, then we are sending write requests to segment 0x51 from the client.
I write at address 34 for a length of 75 bytes, then i verify that data in the backing space at offset 34 showed up with the bytes I wrote.

src/openlcb/MemoryConfigClient.hxx
balazsracz: I refactored some of the repeated code into helper functions, but it did not get all that much simpler.

src/openlcb/MemoryConfigClient.hxx
balazsracz: done

src/openlcb/MemoryConfigClient.hxx
balazsracz: gone

PR #122, opened 2018-02-09
Numeric CDI functions made

kaster14:
-Added functions that allow information about CDI numeric values to be retrieved (and therefore displayed in the numeric entry screen) such as min, max, and default value.

### DIFF
--- a/src/openlcb/CDIUtils.hxx
+++ b/src/openlcb/CDIUtils.hxx
@@ -118,7 +118,64 @@ public:
         return n->text;
     }
 
-    /// Clears out al luserinfo structure pointers. This is necessary to use
+    /// Finds the min value of the xml integer, if there isn't one it returns
+    /// a null pointer
+    /// @param "node" is a CDI element (segment, group, or config element).
+    /// @param "found" will be set to true if there was a minimum found, false if 
+    /// the XML did not contain the respective tag.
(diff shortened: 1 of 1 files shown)


src/openlcb/CDIUtils.hxx
```diff
@@ -118,7 +118,64 @@ public:
...
+
+    /// Finds the max value of the xml integer, if there isn't one it returns
+    /// a null pointer
+    /// @param "node" is a CDI element (segment, group, or config element).
+    /// @param "found" will be set to true if there was a maximum found, false if 
+    /// the XML did not contain the respective tag.
+    /// @return contents of the <max> tag, unspecified if !found. 
+    static int find_node_max(const XMLNode *node, bool* found)
+    {
+        auto *n = find_child_or_null(node, "max");
```
balazsracz: we should try to avoid copy-pasting code like this, because it makes it super hard to fix bugs later.

This should be refactored to use a shared helper function, like

static int find_numeric_element(const XMLNode* parent, const char* tag, bool* found) {
  // move implementation here, use `tag` instead of "min" "max" "default"
}

static int find_node_max(const XMLNode *node, bool* found) {
  return find_numeric_element(node, "max", found);
}

kaster14: modified, and tested manually, also merged other branch with validation testing and that tested good as well.

PR #128, opened 2018-02-17
Add option to provide a power mode on startup.

bakerstu:


src/freertos_drivers/net_cc32xx/CC32xxWiFi.hxx
```diff
@@ -87,6 +87,7 @@ public:
         WLAN_LOW_LATENCY_POLICY, /**< WLAN power policy low latency */
         WLAN_LOW_POWER_POLICY,   /**< WLAN power policy low power */
         WLAN_ALWAYS_ON_POLICY,   /**< WLAN power policy always on */
+        WLAN_DEFAULT_POLICY,     /**< WLAN power policy in default state */
```
balazsracz: This is somewhat confusing. There is no such thing as "default_policy" and this value is not accepted by wlan_power_policy_set. Maybe this constant should be called WLAN_POLICY_NO_CHANGE? That's what it seems to be doing.
There is no guarantee that the power policy is in factory default state when the device starts.

bakerstu: Fixed

PR #130, opened 2018-03-07
Bakerstu cc32xx wps

bakerstu:


src/freertos_drivers/net_cc32xx/CC32xxWiFi.hxx
```diff
@@ -130,6 +141,11 @@ public:
                       SecurityType security_type);
 
 
+    /** Initiate a WPS Push Button Control connection.
```
balazsracz: please document whether this blocks the current thread until the connection succeeds or not.
from the `@return` line it's not clear to me if it returns 0 when it successfully started or succesffully finished the WPS search.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -551,46 +578,21 @@ void CC32xxWiFi::wlan_setup_ap(const char *ssid, const char *security_key,
...
-    uint16_t len;
-    uint16_t config_opt;
+    uint8_t sec_type = security_type_to_simplelink(security_type);
 
-    char old_ssid[32];
-    len = 32;
-    config_opt = WLAN_AP_OPT_SSID;
-    sl_WlanGet(SL_WLAN_CFG_AP_ID, &config_opt , &len, (uint8_t*)old_ssid);
-    // simplify strcmp by null terminating the old_ssid
-    if (len != strlen(ssid) || strncmp(old_ssid, ssid, len))
```
balazsracz: how come you are removing these tests? I thought we need to protect the NWP flash from being written too often.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -542,6 +544,31 @@ void CC32xxWiFi::wlan_connect(const char *ssid, const char* security_key,
...
+
+    int result = sl_WlanConnect((signed char*)"WPS_AP", strlen("WPS_AP"), NULL,
+                                &sec_params, NULL);
+    HASSERT(result >= 0);
+
+    sleep(2);
+
+    for (int i = 0; i < 20 && !wlan_ready(); ++i)
+    {
+        connecting_update_blinker();
```
balazsracz: i think it would be a better API to have this busyloop be the responsibility of the caller. It's not clear whether we will be using the blinker to show the connection progress; on the throttle for example the LCD would be better. If we don't have any of the sleep commands in here, we can choose not to start a new thread for the WPS but call this directly from the executor or network thread.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.hxx
```diff
@@ -52,6 +52,9 @@ class CC32xxSocket;
 class CC32xxWiFi : public Singleton<CC32xxWiFi>
 {
 public:
+    /** the value passed to wlan_profile_del() to remove all profiles */
+    static constexpr int PROFILE_DELETE_ALL = 0xFF;
```
balazsracz: I don't see an implementation for this feature. Does this come from the CC32xx network API or would we have to write some code for this to work?

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
bakerstu: The reason is simple.  These values cannot be read from the NWP.  This is prevented for security reasons.  I'll put some commentary in the code.

balazsracz: is the non-read true only for the password or also the rest of the settings?

PR #131, opened 2018-03-14
traction node name and address helper function

balazsracz:
Adds a helper function to render traction node name and to parse traction node ID.

### DIFF
--- a/src/openlcb/TractionDefs.hxx
+++ b/src/openlcb/TractionDefs.hxx
@@ -98,6 +98,8 @@ struct TractionDefs {
     static const uint64_t NODE_ID_MARKLIN_MOTOROLA = 0x060300000000ULL;
     /// Node ID space allocated for the MTH DCS protocol.
     static const uint64_t NODE_ID_MTH_DCS = 0x060400000000ULL;
+    /// Node ID space mask.
+    static const uint64_t NODE_ID_MASK = 0xFFFF00000000ULL;
 
     enum
     {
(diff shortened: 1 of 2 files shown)


src/openlcb/TractionDefs.hxx
```diff
@@ -97,6 +98,8 @@ struct TractionDefs {
     static const uint64_t NODE_ID_MARKLIN_MOTOROLA = 0x060300000000ULL;
     /// Node ID space allocated for the MTH DCS protocol.
     static const uint64_t NODE_ID_MTH_DCS = 0x060400000000ULL;
+    /// Node ID space mask.
+    static const uint64_t NODE_ID_MASK = 0xFFFF00000000ULL;
```
bakerstu: Technically these node ID's are deprecated in the standard.  We originally agreed to do this so that there was no one-to-one mapping of DCC address to Node ID.  This is so we could have more than one Node ID representing more than one physical locomotive, but still use the same DCC address.  I know in our implementation, essentially hard coding node ID's to specific legacy addresses, but I'd like to get away from this and start assigning out DCC node ID's arbiltrarilly, as the standard and Yahoo group discussions had intended.

balazsracz: In the feedback for the throttle you requested me to make a guess at the node name based only on the node is. This parsing is necessary for that. If we arbitrarily assign DCC node IDs then we won't have much choice than to show mac address format to the user which you described as inappropriate user experience.

bakerstu: Yes, I understand the potential tradeoff, but right now we are implementing a system that the OpenLCB group (beyond just you and I) explicitly decided against.  I'm not necessarily opposed to "hard coding" legacy addresses, but I think procedurally, we should be going back to the OpenLCB list to state our intentions to diverge and why.  I wonder if we can have a compromise solution where both hard coded and arbitrary ID's are allowed.

It might be a good idea to circulate the discussion with Alex and David first before going public on the list.  That way they can help us keep it under control.  Also, we should put it on the Traction list, as it is less likely to attract mass comments into the weeds.

I'm okay with this change for now with the presumption that we can come to the compromise solution on the OpenLCB list.

bakerstu: An alternate suggestion...

As a way to minimize the effect on pagination, I wonder if it would be to okay to blank the name in the list for up to 1 second until the name comes in from the lookup request.  That way, if the name comes back, we don't see the Node ID "flash" briefly on the screen, but if it does not, it still gets displayed.

balazsracz: I understand your comment, but at that time in the mailing list there were a lot of derailed conversations also with incredible scope creep. There were proposals about how we could have multiple command stations on the same network and such, but no clear proposals of how the protocol should work. Then there was an entire different set of no-conclusion conversations about dynamic node ID assignment which eventually ended up with splitting of the group.
For this reason I regard that conversation as "inconclusive" rather than "decisive". For the moment we have the status quo of using those reserved node ID ranges in a fixed scheme. Having these helper functions in a single place allows us to centralize the functionality so that if and when we come up with an alternative scheme we can modify it in place.

Also note that the way the helper function's signature is presented this is forwards compatible; any node IDs that are not of this specific format will be treated as unknown (e.g. deadrail) locomotives and rendered correctly by the UI,

If we want we can choose to render an unknown name instead, but I'm not thrilled with an empty string. Having another timeout would add a lot of additional code complexity. As the user experience this fixed scheme is definitely pretty good, and we can rewrite the code when we have a workable method for multi-command station networks. I don't think we need to implement a solution for that at this point.

PR #132, opened 2018-03-15
Adds update complete support to the MemoryConfigClient

balazsracz:
Adds support for sending UPDATE_COMPLETE, REBOOT, FREEZE, and UNFREEZE commands from the MemoryConfigClient to a remote node.
Adds unittests including checking the case when the remote node actually reboots as part of these sequences.

Misc bugfixes:
- fixes a flakiness in the Bootlaoder client test
- makes sure that the datagram reboot command recognizes the initialization complete message even if dst only has the alias filled in and not the node ID
- performance improvement for message send flow to avoid a dynamic allocation and copy of the payload string.


src/openlcb/MemoryConfigClient.hxx
```diff
@@ -135,9 +220,19 @@ public:
...
+            Defs::OPENMRN_TIMEOUT, //< Timeout waiting for ack/nack.
+    };
+
 private:
     Action entry() override
     {
+        request()->resultCode = OPERATION_PENDING;
         switch (request()->cmd)
         {
         case MemoryConfigClientRequest::CMD_READ:
```
bakerstu: Not a big deal, but the style guide says that these case statements should be indented.

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -135,9 +220,19 @@ public:
     {
     }
 
+    enum ResultCodes
```
bakerstu: Add doxygen comment.

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -135,9 +220,19 @@ public:
...
+    {
+        // Internal error codes generated by the send flow
+        OPERATION_PENDING =
+            DatagramClient::OPERATION_PENDING, //< cleared when done is called.
```
bakerstu: These doxygen comments need another '/' to be valid.

src/openlcb/MemoryConfigClient.hxx
balazsracz: done

src/openlcb/MemoryConfigClient.hxx
balazsracz: done

src/openlcb/MemoryConfigClient.hxx
balazsracz: done

PR #139, opened 2018-05-01
changes made for railcom and CV programming changes

kaster14:
Had to make some adjustments for Wifi params, railcom timing, and CV programming menu on the throttle


src/freertos_drivers/ti/TivaDCC.hxx
```diff
@@ -945,13 +945,13 @@ TivaDCC<HW>::TivaDCC(const char *name, RailcomDriver* railcom_driver)
 
     unsigned h_deadband = 2 * (HW::H_DEADBAND_DELAY_NSEC / 1000);
     unsigned railcom_part = 0;
-    fill_timing(RAILCOM_CUTOUT_PRE, 6 - railcom_part, 0);
```
balazsracz: The code here has an invariant that railcom_part is always set to the absolute microsecond that the sum of the previous RAILCOM_CUTOUT_XXX segments represent. This is used to compute the length of next part by a simple subtraction.
You broke this invariant, and after your changes it is impossible to reason what this code does. Please restore this invariant so we can merge this code.

kaster14:
We should consider adding wifiparams.c to to the gitignore.

Also I should double check the timing I made, I think it only effects the tail end of the cutout, testing so far with current timing has had no adverse effects on Railcom performance however, it may just be that the packets I've used have not used the full channel 2 time window so it hasn't been an issue yet.
I will have to scope this up and take some measurements

PR #143, opened 2018-05-26
First crack at DMA SPI driver.

bakerstu:
Not working, quick review.


src/freertos_drivers/ti/CC32xxSPI.cxx
```diff
@@ -88,10 +97,111 @@ CC32xxSPI::CC32xxSPI(const char *name, unsigned long base, uint32_t interrupt,
...
+            MAP_SPIWordCountSet(base, dmaTransferSize_);
+            MAP_IntEnable(interrupt);
+
+            /* Enable channels & start DMA transfers */
+            MAP_uDMAChannelEnable(dmaChannelIndexTx_);
+            MAP_uDMAChannelEnable(dmaChannelIndexRx_);
+
+            portEXIT_CRITICAL();
+
+            MAP_SPIEnable(base);
```
balazsracz: why are you toggling the SPI enable bit?
It's especially strange to toggle it outside of locks.

previously we were fine keeping that bit enabled and only using the presence of data in the buffers for clocking.

PR #145, opened 2018-06-12
Core performance optimizations

balazsracz:
Optimizes the openmrn core executor and timer services to use less CPU for packet processing.
The optimizations are mostly around inner loops of the executor and the dispatcher flow.
- Amortizes the cost of timers the same way we do with select() calls.
- removes duplicate and unnecessary mutexes and locks, for example when we acquired nested mutexes for no good reason
- removes one lock-unlock sequence from the dispatcher's most common code path.
- makes the queue implementations be inlineable by the compiler

The timer poll resampling caused a lot of unittest failures because wait_for_main_executo() suddenly did not ensure that expired timers have run. This PR changes the underlying infrastructure ExecutorGuard to actually wait for the already expired timers by sleeping -1 time on the timer queue instead of getting added to the lowest priority executor queue.


src/executor/Dispatcher.hxx
```diff
@@ -408,6 +408,8 @@ StateFlowBase::Action DispatchFlowBase<NUM_PRIO>::iterate()
 {
     ID id = get_message_id();
     {
+        // @todo(balazs.racz) make the registered handlers structure for the
+        // dispatcher lock-free. This mutex here is very expensive.
```
bakerstu: Should we file an issue for this?

src/executor/Executor.cxx
```diff
@@ -286,14 +286,13 @@ void *ExecutorBase::entry()
     {
         Executable *msg = nullptr;
         unsigned priority = UINT_MAX;
-        long long wait_length = activeTimers_.get_next_timeout();
-        if (!selectPrescaler_ || empty()) {
+        if (!selectPrescaler_ || ((msg = next(&priority)) == nullptr)) {
```
bakerstu: Should we cleanup the style with hanging braces while we are making changes here?

src/executor/Timer.hxx
```diff
@@ -155,7 +155,7 @@ public:
      * the virtual method timeout(). */
     void run() override;
 
-    /** Special return values from the timeout function. */
+    /** Special return values frombtim the timeout function. */
```
bakerstu: typo

src/executor/Dispatcher.hxx
balazsracz: done

src/executor/Executor.cxx
balazsracz: done

src/executor/Timer.hxx
balazsracz: fixed

PR #153, opened 2018-07-28
Bakerstu wifi status callback

bakerstu:
I'm interested in your feedback.  What I have works great, but I wonder if it can be improved.  I could not use the existing callback structure because it is at the socket connect level and not at the IP acquired level.

### DIFF
--- a/src/freertos_drivers/common/MCP2515Can.hxx
+++ b/src/freertos_drivers/common/MCP2515Can.hxx
@@ -659,7 +659,7 @@ private:
 
     void (*interruptEnable_)(); /**< enable interrupt callback */
     void (*interruptDisable_)(); /**< disable interrupt callback */
-    uint32_t state_; /**< present bus state */
+    unsigned state_     : 4; /**< present bus state */
     unsigned txPending_ : 2; /**< transmission in flight */
     unsigned gpoData_   : 2; /**< local copy of the I/O expansion output data */
     unsigned gpiData_   : 3; /**< local copy of the I/O expansion input data */
--- a/src/freertos_drivers/common/MCP2515Can.hxx
+++ b/src/freertos_drivers/common/MCP2515Can.hxx
@@ -43,6 +43,8 @@
 #include "os/Gpio.hxx"
 #include "os/OS.hxx"
 
+#include "can_ioctl.h"
+
 #define MCP2515_DEBUG 0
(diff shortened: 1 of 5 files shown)


src/freertos_drivers/common/MCP2515Can.hxx
```diff
@@ -648,6 +659,7 @@ private:
 
     void (*interruptEnable_)(); /**< enable interrupt callback */
     void (*interruptDisable_)(); /**< disable interrupt callback */
+    uint32_t state_; /**< present bus state */
```
balazsracz: it would probably be sufficient to allocate just a few bits for this field.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -593,13 +594,16 @@ void CC32xxWiFi::wlan_setup_ap(const char *ssid, const char *security_key,
 
 void CC32xxWiFi::connecting_update_blinker()
 {
-    if (!connected)
+    if (ipAcquiredCallback_ != nullptr)
```
balazsracz: I don't understand this change. 
- if the caller does not want the blinker to be updated, they should not call connecting_update_blinker
- the logic of this if seems to be inverted?
- this change in logic should probably be represented in a change of the documentation comment

src/freertos_drivers/net_cc32xx/CC32xxWiFi.hxx
```diff
@@ -305,6 +305,11 @@ public:
         return rssi;
     }
 
+    void add_ip_acquired_callback(std::function<void(bool)> callback)
```
balazsracz: set_ip_acquired_callback
(normally an api called add_..._callback suggests there is a list of hooks, which is a common API, but you only have one)

please add documentation comment. Declare when this function will be called.
Declare which thread it will be called from!

src/freertos_drivers/common/MCP2515Can.hxx
bakerstu: Fixed

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
bakerstu: fixed

PR #158, opened 2018-09-01
Makes consist add order-preserving

balazsracz:
Updates the linked lists for consists to be order-preserving.
Updates consist add return value to not fail when a consist flag is being updated.


src/openlcb/TractionTrain.cxx
```diff
@@ -369,10 +369,14 @@ struct TrainService::Impl
                             Defs::ERROR_INVALID_ARGS_MESSAGE_TOO_SHORT);
                     NodeID target = data_to_node_id(payload() + 3);
                     uint8_t flags = payload()[2];
-                    bool resp = train_node()->add_consist(target, flags);
+                    if (target == train_node()->node_id()) {
```
bakerstu: style:  hanging brace.

src/openlcb/TractionTrain.hxx
```diff
@@ -130,15 +130,16 @@ public:
     {
         if (!tgt) return false;
         if (tgt == node_id()) return false;
```
bakerstu: style:  both if statements above require braces

src/openlcb/TractionTrain.cxx
balazsracz: fixed

src/openlcb/TractionTrain.hxx
balazsracz: done

PR #159, opened 2018-09-03
Add missing parenthesis

kennethlong:
Working on updating and adding info to the getting started docs.  I noticed this minor error.


balazsracz:
fascinating bug, thanks for catching! I'm wondering how this file even works with unbalanced parentheses.

PR #160, opened 2018-09-03
Bakerstu socket client self

bakerstu:
This helps to resolve the issue reported here: https://github.com/bakerstu/tcs_booster/issues/33

This fix gathers a list of all the interface addresses and checks to make sure the one we are trying to connect to is not our own.  We also verify that we have at least one valid IP address at the time of connection.

There is technically a race condition involved which I do not know how to resolve.  Since we typically use DHCP, it is possible (however extremely unlikely) that we would have our IP addressed changed between the time we validate our list to when we complete the connection.  This is an extremely unlikely case, but if anyone has ideas on how to resolve it, I'm open to suggestion.

### DIFF
--- a/boards/armv7m/default_handlers.h
+++ b/boards/armv7m/default_handlers.h
@@ -128,14 +128,29 @@ void hard_fault_handler_step_3(void);
 
 __attribute__((__naked__)) static void hard_fault_handler(void)
 {
+// set switch to 0 in order to alternatively recreate the previous stack frame
+// and hald the CPU
+#if 1
     __asm volatile
     (
         " tst   lr, #4               \n"
         " ite   eq                   \n"
         " mrseq r0, msp              \n"
         " mrsne r0, psp              \n"
-        " ldr r1, =hard_fault_handler_step_2 \n"
-        " bx r1 \n");
+        " ldr r1, [r0, #24]          \n"
+        " ldr r2, =hard_fault_handler_step_2 \n"
+        " bx r2 \n");
+#else
+    __asm volatile
+    (
+        " tst   lr, #4 \n"
+        " ite   eq     \n"
+        " mrseq r0, msp\n"
+        " mrsne r0, psp\n"
+        " mov   sp, r0 \n"
+        " bkpt  #1     \n"
+    );
+#endif
 #if 0
     // saves our return address
         " mov   r2, lr               \n"
--- a/src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
+++ b/src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
@@ -1086,7 +1086,7 @@ int getaddrinfo(const char *nodename, const char *servname,
     std::unique_ptr<struct sockaddr> sa(new struct sockaddr);
     if (sa.get() == nullptr)
     {
-        free(*res);
+        ai.reset();
         return EAI_MEMORY;
     }
     memset(sa.get(), 0, sizeof(struct sockaddr));
--- a/src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
+++ b/src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
@@ -43,6 +43,7 @@
 #include <netdb.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
+#include <ifaddrs.h>
 
 // Simplelink includes
 #include <ti/drivers/net/wifi/simplelink.h>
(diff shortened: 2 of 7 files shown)


dpharris:
Can you mark the connection in some manner, so that you can recognize it?

David

bakerstu:
David,

I presume you are referring to the race condition described in the opening of this pull request.  If so, one thing I had thought about doing, which would close the race window even further, would be to recheck the interface addresses again, after the connection is made, and reject (close) the connection if for some reason the address used is now local.  This test would be completed before the socket connection actually gets used by the application.

boards/armv7m/default_handlers.h
```diff
@@ -128,14 +128,29 @@ void hard_fault_handler_step_3(void);
...
+#if 1
     __asm volatile
     (
         " tst   lr, #4               \n"
         " ite   eq                   \n"
         " mrseq r0, msp              \n"
         " mrsne r0, psp              \n"
-        " ldr r1, =hard_fault_handler_step_2 \n"
-        " bx r1 \n");
+        " ldr r1, [r0, #24]          \n"
```
balazsracz: I don't understand this change. Where does this value go? hard fault handler only has one argument.

src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
```diff
@@ -1086,7 +1087,7 @@ int getaddrinfo(const char *nodename, const char *servname,
     std::unique_ptr<struct sockaddr> sa(new struct sockaddr);
     if (sa.get() == nullptr)
     {
-        free(*res);
+        ai.reset();
```
balazsracz: this line is not necessary -- will happen automatically upon return. (The previous version was a bug though, good catch.)

src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
```diff
@@ -1217,4 +1218,113 @@ const char *inet_ntop(int af, const void *src, char *dst, socklen_t size)
...
+    {
+        errno = ENOMEM;
+        return -1;
+    }
+    memset(ia.get(), 0, sizeof(struct ifaddrs));
+
+    std::unique_ptr<char[]> ifa_name(new char[6]);
+    if (ifa_name.get() == nullptr)
+    {
+        ia.reset();
```
balazsracz: this is unnecessary

src/freertos_drivers/net_cc32xx/CC32xxSocket.cxx
```diff
@@ -1217,4 +1218,113 @@ const char *inet_ntop(int af, const void *src, char *dst, socklen_t size)
...
+        errno = ENOMEM;
+        return -1;
+    }
+
+    std::unique_ptr<struct sockaddr> ifa_addr(new struct sockaddr);
+    if (ifa_addr == nullptr)
+    {
+        ifa_name.reset();
```
balazsracz: both of these are unnecessary

src/utils/SocketClient.cxxtest
```diff
@@ -162,3 +164,27 @@ TEST_F(SocketClientTest, DISABLED_connect_mdns_not_published)
...
+    EXPECT_CALL(*this, status_callback(SocketClient::Status::CONNECT_FAILED_SELF)).Times(AtLeast(1));
+
+    usleep(40000);
+    wait();
+    usleep(40000);
+    wait();
+
+    sl->shutdown();
+    sc->shutdown();
+    delete sl;
```
balazsracz: if you allocate these within the funciton then delete them before returning, it's easier to allocate them on the stack as a local variable.

src/utils/SocketClient.hxx
```diff
@@ -59,6 +60,7 @@ public:
         MDNS_CONNECT,
         STATIC_CONNECT,
         CONNECT_FAILED,
+        CONNECT_FAILED_SELF,
```
balazsracz: i'm not sure we need to have a separate status returned for this. It will make the UI more complex. We could report the same as CONNECT_FAILED.

src/utils/SocketClient.hxx
```diff
@@ -254,34 +259,100 @@ private:
 
             if (ai_ret == 0 && addr_)
             {
-                /* able to resolve the hostname */
-                fd_ = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
-                if (fd_ >= 0)
+                /* able to resolve the hostname to an address */
+                bool addr_okay = true;
+                if (addr_->ai_addr->sa_family != AF_INET)
```
balazsracz: what happened previously when you resolved the target host to an ipv6 address?

src/utils/SocketClient.hxx
```diff
@@ -254,34 +259,100 @@ private:
...
+                                        has_address = true;
+                                    }
+                                }
+                            }
+                            ifa = ifa->ifa_next;
+                        }
+                        freeifaddrs(ifa_free);
+                        if (!has_address)
+                        {
+                            /* none of the network interfaces have a valid
```
balazsracz: I think if we avoid checking has_address we should be fine as well.
The reason I suggest that is that having an implementation for getifaddr that just sets to nullptr is valid, without giving you protection.
Whereas has_address check will prevent all connections from happening with a really hard to debug error.
So if a new developer comes and implements a network driver for a new chip, they will be really puzzled.

src/utils/SocketClient.hxx
```diff
@@ -355,10 +426,18 @@ private:
...
+    /** number of seconds between retries */
     uint8_t retrySeconds_;
+
+    /** time in seconds that the connect is supposed to
+     *  timeout and look for a possible shutdown
+     */
     uint8_t timeoutSeconds_;
 
+    /** disallow local connections to one's self */
+    bool disallowLocal_;
```
balazsracz: `uint8_t   disallowLocal_ : 1` ?
  one bit would be enough for this and you have some unused alignment bytes here anyway.

boards/armv7m/default_handlers.h
bakerstu: I'm laying the ground work to merge in the SP restore later.

src/utils/SocketClient.hxx
bakerstu: This was added primarily for unit test purposes.

src/utils/SocketClient.hxx
bakerstu: Didn't work.  I suspect we want IPv6 support, but we don't have it today.

src/utils/SocketClient.hxx
bakerstu: Perhaps but it would not be portable to Linux and MacOS.

balazsracz: I don't get what would be the compatibility problem.
Just replace line 313 with ```addr_okay = true;```
This will work just fine on linux and mac. There is always a loopback interface anyway so you can never get into this codepath on linux and mac.

PR #161, opened 2018-09-05
DCC decoder application

balazsracz:
Adds a standalone application that prints DCC packets for debugging purposes. It connects to the track and uses the DCC decoder code and the DCC packet debug conversion routine to implement a DCC-to-USB sniffer.


applications/dcc_decoder/README.md
```diff
@@ -0,0 +1,57 @@
+# DCC decoder application
```
bakerstu: I would prefer this to be in Doxygen so that it can hyperlink to the relevant code as needed.  Maybe it is possible for this to remain in mark down and for Doxygen to pick it up.  https://www.stack.nl/~dimitri/doxygen/manual/markdown.html

applications/dcc_decoder/main.cxx
```diff
@@ -0,0 +1,83 @@
...
+
+/** Entry point to application.
+ * @param argc number of command line arguments
+ * @param argv array of command line arguments
+ * @return 0, should never return
+ */
+int appl_main(int argc, char *argv[])
+{
+    setblink(0);
+    int fd = ::open("/dev/nrz0", O_RDONLY);
```
bakerstu: I assume that nrz = "non return to zero"?  I don't really like this name, it is not exactly an accurate description of DCC.  Could we call something else?  Maybe /dev/dcc or something?

balazsracz: renamed to /dev/dcc_decoder0

applications/dcc_decoder/README.md
balazsracz: I added the applications directory but excluded all c and cxx sources from there to avoid the duplicate documentations/definitions.
I also figured out how to link this .md file into a meaningful place in our generated documentations.

PR #168, opened 2018-09-19
Getting started document updated to reflect current state of the union

rilull:
This document should let somebody go from zero to "it works!". This captures all the pieces that I used to get it working for me and I'm a single tick above total noob when it comes to development.

I think we might want to expand / split the documentation off a bit, but I'll leave that decision to the project brain trust. I can see where we want to make it streamlined and uncluttered, but I don't want to make a monster either.


rilull:
Updated file uploaded into my fork.

doc/getting_started.dox
```diff
@@ -110,9 +131,99 @@
...
+ * @code
+ * openmrn/applications/nucleo_io/targets$ ls -l
+total 12
+ * drwxr-xr-x 2 rick rick 4096 Sep 17 20:35 freertos.armv6m.st-stm32f091rc-nucleo-dev-board
+ * drwxr-xr-x 3 rick rick 4096 Sep 18 21:52 freertos.armv7m.st-stm32f303re-nucleo-dev-board
+ * -rw-r--r-- 1 rick rick  172 Sep  7 20:31 Makefile
+ * $cd freertos.armv7m.st-stm32f303re-nucleo-dev-board/
+ * $make
+ * @endcode
+ * As before, you will get a .elf file, nucleo_io.elf, so the objcopy command is called for here.
```
kennethlong: Rick, this looks great, I think you got everything we ran into.  Glad you remembered that note for the GCC build tools.  That one had me scratching my head for a few hours.  I have some stuff on setting up OpenLCB I will add after this gets merged.

PR #169, opened 2018-09-22
TCP interface

balazsracz:
Implements an interface using the draft Tcp protocol. 
The new interface operates in conjunction with a Hub<string>, where each TCP rendered packet is in exactly one buffer traversing the Hub. A select-aware state flow is added that segments incoming bytes from an fd into packets of the Tcp protocol format with as little memory fragmentation as possible. This needed a new HubDeviceSelect (which is now templated on the read segmenting flow) and the equivalent code of create_gc_port_for_can_hub to handle graceful closing of sockets by the remote side, socket errors or teardown of the interface locally (in unit tests).
This scheme supports having multiple clients connected to a single hub, which is necessary for TCP servers. There is no filtering, i.e., all messages are passed on to every client.

The tcp interface itself converts between the Buffer<string> having the binary (wire) representation and the Buffer<GenMessage> which is the inmemory struct representation, taking care of loopback of global and addressed messages, as well as filtering incoming addressed messages.

Fragmented messages are not yet supported. The datagram service is not yet supported. Routing between a CAN interface and a TCP interface is not yet supported.

Unit tests are added for the following schemes:
- static parsing and rendering functions; sequence generator
- the transmit flow communicating with an FD.
- the receive flow reading from an FD using select; with various alignment cases as it is reassembling a single string from the FD.
- the interface, for sending and receiving addressed and global messages, including tests for loopback and filtering noninteresting addressed messages
- the interface handling of VerifyNodeID messages.
- a TcpIf and a loopback socket connected as a client, sending and receiving wire payload directly on the fd.
- a TcpIf as a hub, with three loopback sockets connected as clients and each having its own TcpIf instance, sending GenMessages between each other using the high-level APIs.


src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+        {
+            LOG(WARNING, "Incomplete or incorrectly formatted TCP message.");
+            return false;
+        }
+        uint16_t flags = data_to_error(&src[0]);
+        if ((flags & FLAGS_OPENLCB_MSG) == 0)
+        {
+            return false;
+        }
+        HASSERT((flags & FLAGS_CHAINING) == 0);
```
bakerstu: Should we have an @todo comment here since we will be removing chaining from the standards?

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+        {
+            tgt->set_flag_dst(GenMessage::DSTFLAG_NOT_LAST_MESSAGE);
+        }
+        // Contents of the size field in the header.
+        uint32_t sz = (uint8_t)src[HDR_SIZE_OFS];
+        sz <<= 8;
+        sz |= (uint8_t)src[HDR_SIZE_OFS + 1];
+        sz <<= 8;
+        sz |= (uint8_t)src[HDR_SIZE_OFS + 2];
+        if ((sz + HDR_SIZE_OFS) > src.size())
```
bakerstu: Is this sequence not a repeat of the get_tcp_message_len() operation?  Don't write the same code twice?

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+    {
+        if (tcp_payload.size() < ABS_MTI_OFS + 2)
+        {
+            return UINT_MAX;
+        }
+        auto mti = (Defs::MTI)data_to_error(tcp_payload.data() + ABS_MTI_OFS);
+        return Defs::mti_priority(mti);
+    }
+
+    enum
```
bakerstu: Consider commenting the members of this enum

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+    };
+
+private:
+    /// No usable constructor; this is a static-only class.
+    TcpDefs();
+};
+
+/**
+   Virtual clock interface. Implementations are not required to be thread-safe.
+*/
```
bakerstu: Change to C++ style comment or put an '*' preceding the text line.

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+{
+public:
+    /** Returns the next strictly monotonic sequence number. */
+    virtual long long get_sequence_number() = 0;
+};
+
+/**
+   Implementation of sequence number generator that uses the real clock. Not
+   thread-safe.
+ */
```
bakerstu: same.

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+class ClockBaseSequenceNumberGenerator : public SequenceNumberGenerator
+{
+public:
+    long long get_sequence_number() override
```
bakerstu: Add method header comment.

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+};
+
+/**
+ * This flow renders outgoing OpenLCB messages to their TCP stream
+ * representation.
+ */
+class TcpSendFlow : public MessageStateFlowBase
+{
+public:
+    TcpSendFlow(If *service, NodeID gateway_node_id,
```
bakerstu: Add method header comment here and below state flow methods.

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+private:
+    Action entry() override
+    {
+        auto id = nmsg()->dst.id;
+        if (id)
+        {
+            auto *n = iface()->lookup_local_node(id);
+            if (n)
+            {
+                // Addressed message loopback.
```
bakerstu: This is interesting.  I would have expected loopback messages to happen at the general level and not the TCP specific level.  If that is the case, should this be an assert instead?

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+    HubPortInterface *skipMember_;
+    /// Populated into the source gateway field of the outgoing messages.
+    NodeID gatewayId_;
+    /// Responsible for generating the sequence numbers of the outgoing
+    /// messages.
+    SequenceNumberGenerator *sequenceNumberGenerator_;
+};
+
+/// This flow is listening to data from a TCP connection, segments the incoming
+/// data into TcpMessages based on the incoing size, and forwards packets
```
bakerstu: "incoming"

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+/// containing the TCP message as string payload.
+class FdToTcpParser : public StateFlowBase
+{
+public:
+    FdToTcpParser(FdHubPortService *s, HubPortInterface *dst,
```
bakerstu: Add method comment block here and state flow methods below.

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+    void shutdown()
+    {
+        auto *e = this->service()->executor();
+        if (e->is_selected(&helper_))
+        {
+            e->unselect(&helper_);
+            helper_.remaining_ = 0;
+        }
+        set_terminated();
+        notify_barrier();
```
bakerstu: This is an interesting pattern.  I wonder if it could be platformized and extended to support sleep_and_wait() states as well.  Just a thought.

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+    /// Assembly buffer. Holds the captured message so far.
+    string msg_;
+    /// Where to send parsed messages to.
+    HubPortInterface *dst_;
+    /// Parsed messages will be initialized to this skipMember_.
+    HubPortInterface *skipMember_;
+    StateFlowSelectHelper helper_{this};
+};
+
+using TcpHubDeviceSelect = HubDeviceSelect<HubFlow, FdToTcpParser>;
```
bakerstu: Is it good practice to have "using" in the global scope of a header file?

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+{
+public:
+    /// @param target is where to send the parsed messages. Usually the
+    /// interface's dispatcher flow.
+    TcpRecvFlow(MessageHandler *target)
+        : target_(target)
+    {
+    }
+
+    void send(Buffer<HubData> *data, unsigned prio) override
```
bakerstu: Add comment header.

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+        auto dst = get_buffer_deleter(target_->alloc());
+        dst->set_done(data->new_child());
+        if (TcpDefs::parse_tcp_message(*src->data(), dst->data()))
+        {
+            target_->send(dst.release(), prio);
+        }
+    }
+
+private:
+    MessageHandler *target_;
```
bakerstu: Add comment.

src/utils/FdUtils.hxx
```diff
@@ -0,0 +1,87 @@
...
+#include "utils/macros.h"
+
+struct FdUtils
+{
+    /// Performs a reliable read from the given FD. Crashes if the read fails.
+    ///
+    /// @param fd the file to read data from
+    /// @param buf the location to write data to
+    /// @param size how many bytes to read
+    ///
```
bakerstu: optionally remove unnecessary empty comment line here and below.

src/utils/FdUtils.hxx
```diff
@@ -0,0 +1,87 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * \file fd_utils.hxx
```
bakerstu: Does not match file name.

src/utils/Hub.hxx
```diff
@@ -232,4 +232,34 @@ protected:
...
+
+    /// Callback from the readflow when it encounters an error.
+    virtual void report_read_error() = 0;
+
+protected:
+    // For barrier_.
+    template <class HFlow> friend class HubDeviceSelectReadFlow;
+    friend class openlcb::FdToTcpParser;
+
+    FdHubPortService(ExecutorBase *exec, int fd)
```
bakerstu: Add comment header.

src/utils/HubDeviceSelect.hxx
```diff
@@ -127,34 +127,141 @@ struct SelectBufferInfo<Buffer<CanHubData>> {
...
+        }
+        else
+        {
+            return this->read_single(&selectHelper_, device()->fd(),
+                (void *)b_->data()->data(), b_->data()->size(),
+                STATE(read_done), 0);
+        }
+    }
+
+    Action read_done()
```
bakerstu: Add comment header.

src/utils/HubDeviceSelect.hxx
```diff
@@ -344,7 +363,12 @@ protected:
         {
         }
 
-        /// Unregisters this oebject from the flows.
+        ~WriteFlow()
```
bakerstu: Add comment header.  A simple "/// Destructor." will do.

balazsracz: done

src/utils/Hub.hxx
balazsracz: done

src/utils/HubDeviceSelect.hxx
balazsracz: done

src/utils/FdUtils.hxx
balazsracz: done

src/utils/FdUtils.hxx
balazsracz: done

src/openlcb/IfTcpImpl.hxx
balazsracz: done

src/openlcb/IfTcpImpl.hxx
balazsracz: done

src/openlcb/IfTcpImpl.hxx
balazsracz: using namespace foobar; is forbidden.
the syntax here is a typedef, just formatted as a more readable equivalent since c++11.

src/openlcb/IfTcpImpl.hxx
balazsracz: I've spent countless hours trying to make tests shut down cleanly. In the world of hubs and smarter gateways we have to bring down more and more complex objects in an appropriate manner. So framework support would be definitely nice.
We could use the otherwise currently empty Service concept to implement an ownership/guardian concept for a set of flows.
We can probably not make it happen automatically though; it is not possible to interrupt a state flow without giving it a chance to exit gracefully. The problem is that we never know if the flow is holding on to some resource or not or when there is a notification pointer in somebody else's hands that will come in later.

src/openlcb/IfTcpImpl.hxx
balazsracz: done

src/openlcb/IfTcpImpl.hxx
balazsracz: done

src/openlcb/IfTcpImpl.hxx
balazsracz: We don't have a general level send flow at this point. The API of `If` gives you a send flow pointer, but that is a flow specific to the interface today, and there is no shared implementation (at this point) between different interfaces.
Maybe when we get to do the big refactoring where local interface and router/gateway are separated, then this functionality will go to the local interface instead.

src/openlcb/IfTcpImpl.hxx
balazsracz: done

src/openlcb/IfTcpImpl.hxx
balazsracz: actually it's worse. It's also buggy, and since we have already called get_tcp_message_len earlier in this function this code is not even needed.
Deleted.

src/openlcb/IfTcpImpl.hxx
balazsracz: this code assumes chaining has been removed, so we don't need to have a todo here in the code. I filed https://github.com/openlcb/documents/issues/10

src/openlcb/IfTcpImpl.hxx
```diff
@@ -0,0 +1,553 @@
...
+     * message source (last sending node ID).
+     * @param sequence is a 48-bit millisecond value that's monotonic.
+     * @param target is the buffer into which to render the message. */
+    static void render_tcp_message(const GenMessage &msg,
+        NodeID gateway_node_id, long long sequence, string *tgt)
+    {
+        bool has_dst = Defs::get_mti_address(msg.mti);
+        HASSERT(tgt);
+        string &target = *tgt;
+        target.assign(HDR_LEN + msg.payload.size() +
```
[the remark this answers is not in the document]
balazsracz: That's different. resize will add new zero bytes only if the new length is longer. Assign will clear out any bytes that might exist from before.

src/openlcb/IfTcp.hxx
```diff
@@ -0,0 +1,105 @@
...
+    /// Rendered string for the message payload. Includes the header (flags and
+    /// length as well).
+    string data;
+};
+
+class ClockBaseSequenceNumberGenerator;
+class TcpSendFlow;
+class TcpRecvFlow;
+
+class IfTcp : public If
```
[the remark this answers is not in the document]
balazsracz: done

src/openlcb/IfTcp.hxx
```diff
@@ -0,0 +1,105 @@
...
+    void add_owned_flow(Executable *e) override;
+    void add_owned_flow(Destructable *e)
+    {
+        ownedFlows_.emplace_back(e);
+    }
+    /// Finds a given pointer in the owned flows list, deletes it and removes
+    /// it from the list.
+    void delete_owned_flow(Destructable *d);
+    void delete_local_node(Node *node) override;
+    bool matching_node(NodeHandle expected, NodeHandle actual) override;
```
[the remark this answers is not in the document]
balazsracz: I copied the documentation from the parent class. Normally doxygen does this, but the warning does get emitted.

src/openlcb/IfTcp.hxx
```diff
@@ -0,0 +1,105 @@
...
+    /// from. The interface will register itself into this hub. The executor of
+    /// this device will be used for processing the packets in this interface.
+    /// @param local_nodes_count is the maximum number of virtual nodes that
+    /// this interface will support.
+    IfTcp(NodeID gateway_node_id, HubFlow *device, int local_nodes_count);
+
+    ~IfTcp();
+
+    void add_owned_flow(Executable *e) override;
+    void add_owned_flow(Destructable *e)
```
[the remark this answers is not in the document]
balazsracz: done

src/openlcb/IfTcp.cxx
```diff
@@ -0,0 +1,207 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * \file IfTcp.cxx
+ *
+ * OpenLCB interface implementation for native TCP connection.
+ *
+ * @author Balazs Racz
+ * @date 16 Apr 2017
```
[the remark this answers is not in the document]
balazsracz: Actually it is. That day I was flying to the US and had 11 hours of uninterrupted time to work on it. You can see that the earliest commits from this PR are from that time.

src/openlcb/IfTcp.cxx
```diff
@@ -0,0 +1,207 @@
+/** \copyright
+ * Copyright (c) 2017, Balazs Racz
```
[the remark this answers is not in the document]
balazsracz: see below.

PR #176, opened 2018-10-29
Adds 23017 expander to Nucleo IO boards

balazsracz:
- Adds an I2C driver for the STM32 MCUs
- adds a generic MCP23017 Gpio expander driver, and an Mcp23017Gpio object
- adds a MultiConfiguredPC class to somewhat efficiently export a large number of pins that can be both input or outputs, configured by the CDI
- instantiates all of this in the nucleo_io application for the F303.


dpharris:
Hey, great news.  Do the outputs have special effects?

David

balazsracz:
The 23017 is an IO extender that does binary output states (or inputs). No PWM. This means that we can do blinking effects and probably a few grades of constant dimming, but no servos or things like ditchlights. Not sure what lighting effects you were thinking of. None of these are currently implemented -- just on and off like on the RailStars board.
For PWM (lighting effects, servos, arbitrary dimming etc), you can use the PCA9685, which is already supported with a driver.

PR #181, opened 2018-12-01
Bakerstu broadcast time

bakerstu:


src/openlcb/BroadcastTime.hxx
```diff
@@ -0,0 +1,448 @@
...
+                               BroadcastTimeDefs::STOP_EVENT_SUFFIX;
+                    break;
+                default:
+                    // should never get here.
+                    return delete_this();
+            }
+
+            if (!clock_->node()->is_initialized())
+            {
+                // since the node is not yet initialized, events get thrown on
```
balazsracz: This is a state flow, so we could just as well delay until the node reaches initialized state.

src/openlcb/BroadcastTimeDefs.hxx
```diff
@@ -0,0 +1,346 @@
...
+    {
+        EVENT_ID_MASK      = 0xFFFFFFFFFFFF0000, ///< Unique ID mask
+        EVENT_SUFFIX_MASK  = 0x000000000000FFFF, ///< suffix mask
+        EVENT_TYPE_MASK    = 0x000000000000F000, ///< type mask
+        EVENT_HOURS_MASK   = 0x0000000000001F00, ///< hours mask
+        EVENT_MINUTES_MASK = 0x00000000000000FF, ///< minutes mask
+        EVENT_MONTH_MASK   = 0x0000000000000F00, ///< month mask
+        EVENT_DAY_MASK     = 0x00000000000000FF, ///< day mask
+        EVENT_YEAR_MASK    = 0x0000000000000FFF, ///< rate mask
+        EVENT_RATE_MASK    = 0x0000000000000FFF, ///< rate mask
```
balazsracz: I think it might make sense to stop the anonymous enum after this and start again for the remaining entries, because above here we have 64-bit entries and below only 16 to 32 bit entries. I would guess now the symbols below also have 64-bit storage class due to being in the same enum.

src/openlcb/BroadcastTimeDefs.hxx
```diff
@@ -0,0 +1,346 @@
...
+
+        EVENT_HOURS_SHIFT   = 8, ///< hours mask
+        EVENT_MINUTES_SHIFT = 0, ///< minutes mask
+        EVENT_MONTH_SHIFT   = 8, ///< month mask
+        EVENT_DAY_SHIFT     = 0, ///< day mask
+        EVENT_YEAR_SHIFT    = 0, ///< day mask
```
balazsracz: typoe in comment

src/openlcb/BroadcastTime.hxx
```diff
@@ -0,0 +1,448 @@
...
+                case START:
+                    event_id = clock_->clock_id() +
+                               BroadcastTimeDefs::START_EVENT_SUFFIX;
+                    break;
+                case STOP:
+                     event_id = clock_->clock_id() +
+                               BroadcastTimeDefs::STOP_EVENT_SUFFIX;
+                    break;
+                default:
+                    // should never get here.
```
balazsracz: log error ?

src/openlcb/BroadcastTime.hxx
```diff
@@ -0,0 +1,448 @@
...
+
+    long long timestamp_; ///< monotonic timestamp from last server update
+    time_t seconds_; ///< clock time in seconds from last server update
+    int16_t rate_; ///< effective clock rate
+    int16_t rateRequested_; ///< pending clock rate
+
+    uint16_t started_ : 1; ///< true if clock is started
+
+private:
+    /// Reset our process local timezone environment to GMT0.
```
balazsracz: is it GMT0 or UTC? Or is that the same

src/openlcb/BroadcastTime.hxx
```diff
@@ -0,0 +1,448 @@
...
+namespace openlcb
+{
+
+/// Implementation of a Broadcast Time Protocol Interface.
+class BroadcastTime : public SimpleEventHandler
+                    , public StateFlowBase
+                    , protected Atomic
+{
+public:
+    /// Constructor.
```
balazsracz: Is this supposed to be public or rather protected? i.e. is it expected that applications are creating these directly (to operate as a time client)?

src/openlcb/BroadcastTimeAlarm.hxx
```diff
@@ -0,0 +1,435 @@
...
+
+        return BroadcastTimeAlarm::entry();
+    }
+
+    /// callback for when the alarm expires
+    void expired_callback()
+    {
+        if (clock_->rate() > 0)
+        {
+            set(clock_->time() + 60);
```
balazsracz: drift

PR #184, opened 2018-12-26
Move nucleo_io servo-related objects to main.cxx to avoid static init data race.

szabodabo:


balazsracz:
please see https://github.com/bakerstu/openmrn/pull/185 instead.

PR #185, opened 2018-12-26
Linker init for servo channels.

balazsracz:
Moves the servo PWM initialization to a constexpr function to allow the servo_channels[] array be initialized by the linker instead of at static initialization time.

This fixes bug when the static initialization order is not appropriate.


balazsracz:
@szabodabo  please review

szabodabo:
Works for me

szabodabo:
thank you for the fix

PR #186, opened 2018-12-27
Add servo min/max rotation options to nucleo_io CDI.

szabodabo:
* Since PWMGPO isn't a singleton, GPIOBit wasn't directly compatible (because of the HW::instance() default argument in the second constructor). Not sure if there's a cleaner way to do this.

* "online" config changes appear not to work, though I *think* this may be an issue with ConfiguredConsumer-based consumers as well. I always need to reset the MCU for any config change to take effect. Not a big issue, but maybe you have ideas on how to fix.

* ServoConsumer's PWMGPO instance is heap-allocated because I want it to be nullptr until first config. Not sure if there's a better way to do it.

* Removed references to servos 5-8 (I don't think they're currently hooked up (I remember a discussion on software PWM)). I could bring them back, but the CDI gets really huge.


applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
```diff
@@ -0,0 +1,135 @@
...
+
+const unsigned servo_ticks_0 = configCPU_CLOCK_HZ * 1 / 1000;   // 48k
+const unsigned servo_ticks_180 = configCPU_CLOCK_HZ * 2 / 1000; // 96k
+
+uint8_t dummy_var[1] = {0x0};
+constexpr const MmapGpio8 DUMMY_GPIO(
+    dummy_var, /*offset=*/0, /*is_output=*/false);
+
+// GPIOBit has a <template HW> ctor that won't work if
+// supplied Gpio* is a PWMGPO; no idea why.
```
balazsracz: I tested this and have a fix merged into master.

I removed the custom gpio implementation if you want to cherrypick the commit 
https://github.com/bakerstu/openmrn/commit/ee860add363ec46af28bed56e894b9c97772f972

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    UpdateAction apply_configuration(
+        int fd, bool initial_load, BarrierNotifiable *done) OVERRIDE
+    {
+        AutoNotify n(done);
+
+        const openlcb::EventId cfg_event_min = cfg_.event_rotate_min().read(fd);
+        const openlcb::EventId cfg_event_max = cfg_.event_rotate_max().read(fd);
+        const uint8_t cfg_servo_min_pct = cfg_.servo_min_percent().read(fd);
```
balazsracz: You want to have the ability to overdrive and underdrive servos. Some servos are able to rotate a bit more if you send them out-of-bounds signal. For this reason you want to interpret this as a signed integer, probably increase the size to 16 bit.
A good way to do this is to set minimum to be -99, maximum to be 200, and normal range be 0 to 100.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
```diff
@@ -0,0 +1,135 @@
...
+        const uint8_t cfg_servo_max_pct = cfg_.servo_max_percent().read(fd);
+
+        const unsigned servo_range_ticks = servo_ticks_180 - servo_ticks_0;
+
+        const unsigned cfg_srv_ticks_min =
+            servo_ticks_0 + (servo_range_ticks * (cfg_servo_min_pct / 100.0));
```
balazsracz: This is going to link in floating point libraries which is somewhat undesirable.
Please check if you can get equivalent results using Fixed16 in src/utils/Fixed16.hxx. It also supports division.

Another alternative is to rewrite this entire expression to a numerically more stable weighted average:
(min_pct * ticks_0 + (100-min_pct) * ticks_180) / 100

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
```diff
@@ -0,0 +1,135 @@
...
+            cfg_srv_ticks_min != pwm_gpo_->get_off_counts() ||
+            cfg_srv_ticks_max != pwm_gpo_->get_on_counts())
+        {
+            auto saved_node = gpio_impl_.node();
+
+            consumer_.~BitEventConsumer();
+            gpio_impl_.~MyGPIOBit();
+            if (pwm_gpo_)
+            {
+                delete pwm_gpo_;
```
balazsracz: if you need to write this, then you most likely want to replace this with a std::unique_ptr<PWMGPO> and use the reset() function.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+            pwm_gpo_ = new PWMGPO(pwm_,
```
balazsracz: Instead of this you might want to consider adding setters for the values you want to change.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
```diff
@@ -0,0 +1,135 @@
...
+            return REINIT_NEEDED;
+        }
+
+        return UPDATED;
+    }
+
+    void factory_reset(int fd) OVERRIDE
+    {
+        cfg_.description().write(fd, "");
+        cfg_.servo_min_percent().write(fd, 0);
```
balazsracz: you could try `CDI_FACTORY_RESET(cfg.servo_min_percent);`

precondition that there be a default value attribute defined on the config structure.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
```diff
@@ -0,0 +1,135 @@
...
+        cfg_.servo_max_percent().write(fd, 100);
+    }
+
+private:
+    // not owned; lives forever
+    PWM *pwm_; // timer channel
+
+    // all the rest are owned and must be reset on config change.
+    // pwm_gpo_ heap-allocated because it's nullptr until first config.
+    PWMGPO *pwm_gpo_; // has PWM* and on/off counts
```
balazsracz: The style guide says this should be named `pwmGpo_`

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
```diff
@@ -0,0 +1,135 @@
...
+
+    MyGPIOBit gpio_impl_;                // has on/off events, Node*, and Gpio*
```
balazsracz: `gpioImpl_`

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
```diff
@@ -0,0 +1,135 @@
...
+#define _APPLICATIONS_IO_BOARD_TARGET_SERVOCONSUMER_HXX_
+
+#include "config.hxx"
+#include "os/MmapGpio.hxx"
+
+const unsigned servo_ticks_0 = configCPU_CLOCK_HZ * 1 / 1000;   // 48k
+const unsigned servo_ticks_180 = configCPU_CLOCK_HZ * 2 / 1000; // 96k
+
+uint8_t dummy_var[1] = {0x0};
+constexpr const MmapGpio8 DUMMY_GPIO(
```
balazsracz: There is freertos_drivers/common/DummyGPIO.hxx
to see how to use https://github.com/bakerstu/openmrn/commit/ee860add363ec46af28bed56e894b9c97772f972

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
```diff
@@ -0,0 +1,135 @@
+#ifndef _APPLICATIONS_IO_BOARD_TARGET_SERVOCONSUMER_HXX_
+#define _APPLICATIONS_IO_BOARD_TARGET_SERVOCONSUMER_HXX_
+
+#include "config.hxx"
```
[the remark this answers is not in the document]
balazsracz: Three reasons:
- it behaves more like a cxx file in that certain global constants it is actually initializing. This was a tradeoff that was made so that application developers would not have to carry around a config.hxx and a config.cxx file.
- For the compile CDI binary config.hxx needs to be compiled for the host which would be difficult to achieve correctly in the makefile with .o files lying around for different targets. Keeping it header-only allows compile_cdi to be created as a single compilation unit.
- Anything that includes config.hxx can by definition only live in the application tree (because that's where config.hxx exists) so it's not very portable. This makes the need for including config.hxx very low.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
szabodabo: nice, thanks

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
szabodabo: way better, thanks.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
szabodabo: I tested this with JMRI setting the field to "-10". OpenLCB Monitor log shows this datagram sent when I press "Write":

`
13:11:24.487: [[1a5654a5] 20 01 00 00 02 EC FF F6]  S: 02.01.12.FE.08.D3 - 05.01.01.01.18.15 Datagram: (8) 20.1.0.0.2.EC.FF.F6
`

and then the field [updates to show value "65526"](https://imgur.com/NBbIDlQ), which matches `FF F6`. When I try -11, the field shows 65525, and `FF F5` gets sent. This looks like a JMRI bug to me.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
szabodabo: Done. I used this strategy, but I switched `min_pct` with `100-min_pct`. Otherwise, what's calculated is the ticks at `100-percent` .

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
szabodabo: It would reduce memory churn, but it'd be more lines of code. I already need code that initializes `pwmGpo_` on initial config based on the calculated min/max ticks. 

current:
```
apply_config() {
  if (anything_changed) {
    pwmGpo_.reset(new PWMGPO(...));
  }
}
```

with setters:
```
apply_config() {
  if (!pwmGpo_) {
    pwmGpo_.reset(new PWMGPO(...));
  } else if (anything_changed) {
    pwmGpo_.set_max_ticks(...);
    pwmGpo_.set_min_ticks(...);
  }
}
```

is there a good reason to prefer the more verbose case?

balazsracz: I think if you have the setters you would not need to have pwmGpo_ be a pointer at all. You can allocate at construction time and it be a regular class member. Then your code does become simpler.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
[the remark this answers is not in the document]
szabodabo: I will move ServoConsumer and its ServoConsumerConfig to src/openlcb/. Because the PWM object takes in ticks (not time... why?), ServoConsumer uses configCPU_CLOCK_HZ to compute the ticks. Is this freertos-only (defined in FreeRTOSConfig.h)? What about ServoConsumer's deps on DummyGPIO and PWM.hxx, which are also freertos-specific?

For the specific problem at hand, would it be best to pass configCPU_CLOCK_HZ as a constructor parameter to ServoConsumer?

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
szabodabo: I wanted to keep pwmGpo_ nullptr so that the PWM* wouldn't be referenced anywhere until first config. maybe that's silly since the PWMGPO and Gpio ctors don't do anything.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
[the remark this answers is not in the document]
balazsracz: yes that sounds like a good idea

balazsracz:
Probably the best construction argument would be pwm count per millisecond.

--
Sent from my mobile device

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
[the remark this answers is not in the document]
szabodabo: in `HwInit.cxx`, I see

```
Stm32PWMGroup servo_timer(TIM3,
    /*prescaler=*/ (configCPU_CLOCK_HZ * 6 / 1000 + 65535) / 65536,
    /*period_counts=*/ configCPU_CLOCK_HZ * 6 / 1000);
```

does that mean the period, as currently configured, is 6ms? shouldn't the period be 20ms for a standard servo?

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/ServoConsumer.hxx
szabodabo: also PWMGPO's functions depend on having sane values for onCounts_ and offCounts_. I could add handling for sentinel "unset" values in those functions, but... seems like more work than it's worth.

szabodabo:
ok, I think this is getting closer now. thanks for all the help.

src/openlcb/ServoConsumer.hxx
```diff
@@ -0,0 +1,110 @@
...
+#include "freertos_drivers/common/DummyGPIO.hxx"
+#include "freertos_drivers/common/PWM.hxx"
+#include "openlcb/ServoConsumerConfig.hxx"
+#include "os/MmapGpio.hxx"
+#include <memory>
+
+namespace openlcb
+{
+
+// Basically a specialized ConfiguredConsumer.
```
balazsracz: please use doxygen style comments, i.e. `///` for field and structure descriptions. Make sure the comment would read meaningful in the documentation as well.

src/openlcb/ServoConsumer.hxx
```diff
@@ -0,0 +1,110 @@
...
+
+    void factory_reset(int fd) OVERRIDE
+    {
+        cfg_.description().write(fd, "");
+        CDI_FACTORY_RESET(cfg_.servo_min_percent);
+        CDI_FACTORY_RESET(cfg_.servo_max_percent);
+    }
+
+private:
+    // Used to compute PWM ticks for max/min servo rotation.
```
balazsracz: `///`

szabodabo:
looks like this is ready. thanks for all the help and information.

PR #190, opened 2019-01-12
Refactor Tcp connection client

balazsracz:
Refactors TCP SocketClient to provide more flexibility:
- the parameters are extracted into a virtual base structure
- convenience implementation is added for host:port and mdns connection for legacy compatibility
- adds a third connection stage called "reconnect" which is an IP address:port pair
- makes the sequence of the connection attempts configurable
- parallelizes the mDNS lookup and the first connection, if an additional thread is available
- switches from dedicated thread to stateflow + executor model to allow reusing threads between different socket clients.

Some misc infra improvements:
- adds a lambda based executable
- adds implementation to MDNS class to shut down an export.
- adds support in the configuration descriptor to hide a group from the user (while keeping the memory allocated and reachable from code) for internal storage purposes.


src/utils/SocketClient.hxx
```diff
@@ -150,12 +183,23 @@ public:
...
+                timer_.ensure_triggered();
+            }
+            if (isConnected_)
+            {
+                isConnected_ = false;
+                notify();
             }
         }
-        sem_.post();
+        /// @todo wake up any pending tasks
```
bakerstu: Does this todo comment need to be resolved before merging back?

balazsracz: Today tasks are pending because we are in a blocking system call, such as in connect() or a getaddrinfo(). I don't have a way to wake those up so we'll have to just wait for them to finish and timeout.

balazsracz: added comment to the code.

PR #191, opened 2019-01-12
[WIP] initial pass for ESP32 compilation, more work TBD

atanisoft:


arduino/OpenMRN.h
```diff
@@ -0,0 +1,140 @@
+#ifndef _ARDUINO_OPENMRN_H_
```
bakerstu: Missing a BSD license header.  Same comment for all of the new source files.

arduino/readme.md
```diff
@@ -0,0 +1,37 @@
+# Arduino Lib notes
```
bakerstu: Can we add some commentary here (or perhaps another appropriate readme) that indicate the purpose of the arduino tree and that it is not part of a native (non Arduino tools) build?

include/can_frame.h
```diff
@@ -63,7 +63,7 @@
         (_frame).can_id += ((_value) & CAN_SFF_MASK);   \
     }
 
-#elif defined (__nuttx__) || defined (__FreeRTOS__) || defined (__MACH__) || defined (__WIN32__) || defined(__EMSCRIPTEN__) || defined(ESP_NONOS)
+#elif defined (__nuttx__) || defined (__FreeRTOS__) || defined (__MACH__) || defined (__WIN32__) || defined(__EMSCRIPTEN__) || defined(ESP_NONOS) || defined(ESP32)
```
bakerstu: Will we not define __FreeRTOS__ as part of the Arduino build?  Would there not alternatively be a better Arduino specific define instead of an ESP32 specific define?

include/freertos/stropts.h
```diff
@@ -38,12 +38,14 @@
 extern "C" {
 #endif
 
+#ifndef ESP32
```
bakerstu: #ifndef Arduino, ...or something?  Or possibly the right action is to remove this file dependency for non POSIX usage.

src/executor/Executor.hxx
```diff
@@ -87,7 +87,7 @@ public:
      * the execution is completed. @param fn is the closure to run. */
     void sync_run(std::function<void()> fn);
 
-#ifdef __FreeRTOS__
+#if defined(__FreeRTOS__) && !defined(ESP32)
```
bakerstu: We should revisit this.  Should we define __FreeRTOS__ as part of the Arduino build?  ...same comment for those that follow.

src/freertos_drivers/common/Devtab.hxx
```diff
@@ -46,6 +46,9 @@ class FileSystem;
 class Notifiable;
 class DeviceBufferBase;
 
+#ifdef ESP32
```
bakerstu: This seems like a "make it build" hack.  If we don't use POSIX, we should find out why it is getting pulled into the build.

src/openlcb/DatagramDefs.hxx
```diff
@@ -36,6 +36,10 @@
 
 #include <stdint.h>
 
+#if defined(ESP32)
+#undef DISPLAY
```
bakerstu: Is there a better way to address this namespace pollution?  Can we change our OpenMRN away from our usage of keyword "DISPLAY", or can we eliminate this from the ESP32 namespace?

src/openlcb/DatagramDefs.hxx
```diff
@@ -128,7 +132,15 @@ struct DatagramDefs
     {
         return error & BUFFER_UNAVAILABLE;
     }
-    
+
+    /** Constants used by the protocol_size function. */
```
bakerstu: This change is suspicious.  Is it really required?  Why?

src/os/Gpio.hxx
```diff
@@ -37,6 +37,13 @@
 
 #include "utils/macros.h"
 
+#if defined(ESP32)
+#undef LOW
```
bakerstu: Hmm...  These are unfortunate.  Should we change the OpenMRN side?

arduino/OpenMRN.h
atanisoft: These are clones from @balazsracz repo, before merge they will certainly need to be stamped.

arduino/readme.md
atanisoft: Yes, this is currently a dumping ground for notes as I'm working on porting this over. It is really ugly still..

include/can_frame.h
atanisoft: I had started with FreeRTOS being defined but couldn't get it to compile cleanly. I'll take another pass at it.

src/executor/Executor.hxx
atanisoft: yes

src/openlcb/DatagramDefs.hxx
atanisoft: its worse than namespace pollution, on the arduino-esp32 side it is a #define, I had opted originally to rename the two references to DISPLAY in this file but it didn't seem right. I also don't like this option either.

src/openlcb/DatagramDefs.hxx
atanisoft: this was necessary due to compilation failure with it in private. I'd like to revisit this further.

src/os/Gpio.hxx
atanisoft: Perhaps, it is the same issue as observed in DISPLAY. It isn't really being done correctly on the arduino-esp32 side (or ArduinoAvr-Core for that matter likely)

src/freertos_drivers/common/Devtab.hxx
atanisoft: yes, it was in a lot of ways. I'll be dropping this once I solve the includes.

include/can_frame.h
atanisoft: @bakerstu I've got it mostly compiling cleanly now with __FreeRTOS__ defined in platformio.ini but that doesn't solve it for Arduino IDE which I've not found a good solution for yet. I suspect we may need to add a few checks for ARDUINO_ESP32 at the top of certain files and define __FreeRTOS__ in those files.

include/freertos/stropts.h
atanisoft: Ideally yes it would be good to drop the dependency. At this time though, it compiles with this declaration dropped as it is already declared in the ESP-IDF framework with a different datatype for key (int vs unsigned long int here). I'm also not 100% convinced that we can't use the POSIX approach on the ESP32 as the VFS layer supports it as does LwIP. That will only leave the HW CAN controller which has its own API which may be possible to wrap but I don't think that is entirely necessary at this stage.

balazsracz:
I merged the arduino-core branch into master.
Please merge master into your branch and resolve the conflicts you get.
A number of your changes will not be necessary anymore as those were already fixed in arduino-core. Some open questions should be closed by this (e.g. around DISPLAY, OUTPUT/INPUT, HIGH/LOW).

You will find that Can.hxx is replaced by arduino/Can.hxx. This does not use Devtab.hxx anymore. There should be nobody including devtab.hxx -- if somebody does, that's probably undesired, because devtab is specific to how we implemented posix fd's and we are not using the posix layer here.

I replaced the use of portENTER_CRITICAL with an Atomic for arduino/Can.hxx. Your implementation for Atomic seems sound (using the spinlock).

I'm not sure if there are other uses of portENTER_CRITICAL outside of the drivers tree which we should not be compiling. If yes, I can see two options:
- either locally it can be replaced with an atomic
- or we could create a global static spinlock for openmrn that we use in these cases.

atanisoft:
Sounds good, I'll pull the latest master to rebase my updates on top. I'll likely close this PR in the process and resubmit

PR #194, opened 2019-01-19
Externalizes the wifi name and password parameters

balazsracz:
Externalizes the wifi name and password parameters and adds .gitignore file to avoid them accidentally being checked in.


balazsracz:
@atanisoft please take a look.

PR #195, opened 2019-01-19
[WIP] ESP32 updates

atanisoft:
More work on the ESP32 version, more work TBD.

### DIFF
--- a/applications/hub/targets/linux.x86/AvaHiMDNS.cxx
+++ b/applications/hub/targets/linux.x86/AvaHiMDNS.cxx
@@ -126,7 +126,8 @@ void mdns_publish(const char *name, uint16_t port)
 
     int result = avahi_entry_group_add_service(group, AVAHI_IF_UNSPEC,
         AVAHI_PROTO_UNSPEC, (AvahiPublishFlags)0, name,
-        openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN, NULL, NULL, port,
+        openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN,
+        openlcb::TcpDefs::MDNS_PROTOCOL_TCP, NULL, port,
         "platform=linux-x86-openmrn", NULL);
 
     if (result != 0)
--- a/applications/hub/targets/linux.x86/AvaHiMDNS.cxx
+++ b/applications/hub/targets/linux.x86/AvaHiMDNS.cxx
@@ -126,8 +126,7 @@ void mdns_publish(const char *name, uint16_t port)
 
     int result = avahi_entry_group_add_service(group, AVAHI_IF_UNSPEC,
         AVAHI_PROTO_UNSPEC, (AvahiPublishFlags)0, name,
-        openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN,
-        openlcb::TcpDefs::MDNS_PROTOCOL_TCP, NULL, port,
+        openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN_TCP, NULL, NULL, port,
         "platform=linux-x86-openmrn", NULL);
 
     if (result != 0)
--- a/applications/modbus_controller/targets/custom-cc3220sf/main.cxx
+++ b/applications/modbus_controller/targets/custom-cc3220sf/main.cxx
@@ -549,11 +549,8 @@ int appl_main(int argc, char *argv[])
     }
 
     resetblink(WIFI_BLINK_CONNECTING);
-    string mDNSServiceName = openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN;
-    mDNSServiceName.append(".");
-    mDNSServiceName.append(openlcb::TcpDefs::MDNS_PROTOCOL_TCP);
     g_socket_client = new SocketClient(stack.service(),
-        mDNSServiceName.c_str(), WIFI_HUB_HOSTNAME,
+        openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN_TCP, WIFI_HUB_HOSTNAME,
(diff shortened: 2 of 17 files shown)


applications/modbus_controller/targets/custom-cc3220sf/main.cxx
```diff
@@ -549,8 +549,11 @@ int appl_main(int argc, char *argv[])
     }
 
     resetblink(WIFI_BLINK_CONNECTING);
+    string mDNSServiceName = openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN;
+    mDNSServiceName.append(".");
+    mDNSServiceName.append(openlcb::TcpDefs::MDNS_PROTOCOL_TCP);
     g_socket_client = new SocketClient(stack.service(),
-        openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN, WIFI_HUB_HOSTNAME,
+        mDNSServiceName.c_str(), WIFI_HUB_HOSTNAME,
```
balazsracz: this won't work because the lifetime of the new pointer you pass in is stack-lived whereas the SocketClient will probably hold on to the const char* passed in.
Maybe we could have two static symbols, one with the ._tcp suffix and another without? like basename and full name.
Any binary will only use one of the two.

arduino/examples/ESP32WiFiClientBridge/ESP32WiFiClientBridge.ino
```diff
@@ -105,7 +105,8 @@ void setup() {
     openMRNServer.setNoDelay(true);
     openMRNServer.begin();
     MDNS.begin(hostname);
-    MDNS.addService(openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN, "tcp", OPENMRN_TCP_PORT);
+    MDNS.addService(openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN,
+        openlcb::TcpDefs::MDNS_PROTOCOL_TCP, OPENMRN_TCP_PORT);
```
balazsracz: previously this was "tcp" and now it is "_tcp". Is that going to be a problem?

applications/hub/targets/linux.x86/AvaHiMDNS.cxx
```diff
@@ -126,7 +126,8 @@ void mdns_publish(const char *name, uint16_t port)
 
     int result = avahi_entry_group_add_service(group, AVAHI_IF_UNSPEC,
         AVAHI_PROTO_UNSPEC, (AvahiPublishFlags)0, name,
-        openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN, NULL, NULL, port,
+        openlcb::TcpDefs::MDNS_SERVICE_NAME_GRIDCONNECT_CAN,
+        openlcb::TcpDefs::MDNS_PROTOCOL_TCP, NULL, port,
```
balazsracz: I'm not convinced this is equivalent. The documentation says
`/**< The domain to register this domain in. We recommend to pass NULL here, to let the daemon decide */,`
and normally the domain is ".local" I think.

arduino/examples/ESP32WiFiClientBridge/ESP32WiFiClientBridge.ino
atanisoft: no issues, the mDNS system prepends the _ only if missing

applications/hub/targets/linux.x86/AvaHiMDNS.cxx
atanisoft: you are right, I missed that note. It was supposed to go into type and not domain. I'm thinking it may make sense to go with the option you suggested above of having two defines.

applications/modbus_controller/targets/custom-cc3220sf/main.cxx
atanisoft: Yes, I'm thinking that makes sense, looking at that AvaHi docs again it looks like it expects the name to be in one string same as the SocketClient does. I'll add defines for both.

atanisoft:
@TrainzLuvr you may want to check out this PR as it is building on the existing esp32 support code. I'm working on adding more examples and documentation as well as extending support for the rest of the stack. For the IO example make sure to comment out the consumer/producer lines otherwise the ESP32 will crash when JMRI probes the device state.

TrainzLuvr:
Thanks for the heads-up, I'm just setting up my VM to cross-compile code for ESP32 atm. It's already setup for RPi and STM32 so shouldn't be too much effort.

balazsracz:
@TrainzLuvr these examples use the Arduino IDE to compile which can be installed under all three major operating systems.

TrainzLuvr:
Oh, didn't think of that. I had to setup a cross-compile for RPi when I was compiling OpenMRN for it.

I'll use Arduino IDE instead...thanks!

src/freertos_drivers/arduino/Esp32HardwareCanAdapter.hxx
```diff
@@ -0,0 +1,209 @@
...
+            msg.data[7] = can_frame->data[7];
+            if(can_frame->can_eff)
+            {
+                msg.flags |= CAN_MSG_FLAG_EXTD;
+            }
+            if(can_frame->can_rtr)
+            {
+                msg.flags |= CAN_MSG_FLAG_RTR;
+            }
+            esp_err_t tx_result = can_transmit(&msg, pdMS_TO_TICKS(10));
```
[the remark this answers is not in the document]
atanisoft: this is the only API at this time but I'll add a queue and send it to the queue for TX on a diff task, similar to RX. This code will instead just be a "convert and queue" and if it sends to the queue it will be considered "sent" and the TX task will need to handle retries for sending to "can_transmit" if there is a failure in queuing for TX.

also TX / RX buffers are set to 5 packets currently, is this sufficient?

atanisoft:
This needs a bit more work before being considered for merge.

PR #196, opened 2019-01-19
Minor ESP32 fixes.

balazsracz:
Fixes a bug that caused the ESP32 to echo all packets sent to it.
Disables sleep mode to get better responsiveness of the ESP32.

### DIFF
--- a/arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
+++ b/arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
@@ -112,6 +112,7 @@ void setup() {
         delay(500);
         Serial.print(".");
     }
+    // This makes the wifi much more responsive. Since we are plugged in we don't care about the increased power usage. Disable when on battery.
     WiFi.setSleep(false);
 
     printf("\nWiFi connected, IP address: %s\n", WiFi.localIP().toString().c_str());


TrainzLuvr:
I wanted to comment on the WiFi externalization of parameters, but that PR was closed and merged already, so I'll comment here.

Could you please add an option to disable WiFi?

Matter a fact that should be a default behaviour, if no external parameters are provided (blank), WiFi should not be enabled.

This will prevent potential OTA attacks, AP sniffing, etc. IMHO, there's no reason to introduce another wireless IoT device to the world automatically. Device owners should instead configure it first, thus acknowledging they want WiFi on, understanding the consequences.

atanisoft:
@TrainzLuvr It can be disabled by simply not connecting to WiFi. I'll add an example of that once I have a chance to test the hardware CAN adapter using an MCP2515 chip.

TrainzLuvr:
Do you mean not connecting to the AP, or not instantiating a WiFi class?

What I was asking is to *disable* WiFi by default, so it does not turn on at all.

atanisoft:
@TrainzLuvr The ESP32 is a WiFi based device primarily, but yes the entire WiFi (and Bluetooth) stack can be disabled easily. But that would defeat the primary use case of the device, a WiFi LCC interface. But with the built in hardware CAN controller it is possible to use the ESP32 in the same fashion as an Arduino, STM32, or virtually any other MCU that can interface with the LCC bus.

TrainzLuvr:
Personally, I prefer to keep WiFi off until needed, and not enable it just because it is there, for many reasons.

Regarding CAN, since ESP32 has the CAN controller built-in, and only a transceiver is needed, how come you are attaching MCP2515 to it? Are you going to convert the logic levels to 3.3V?
I bought couple of ESP32 DevKits and SN65HVD230 transceivers to setup and test OpenMRN with, hopefully it all works good so I could apply it on my layout.

atanisoft:
I'm considering moving the WiFi pieces into the OpenMRN.h/cxx code which will allow turning on the WiFi stack if you opt into that model. I'll also be adding code for the CAN adapter into the same which can be setup to optionally turn off WiFi. I'll add examples for the same as well.

atanisoft:
also I gave the wrong part number, I have a few transceivers to work with: MCP2551 and SN65HVD230. Both should work just fine once I have the CAN adapter code in place.

atanisoft:
@TrainzLuvr to disable WiFI/BLE add this to the setup() method:
```
WiFi.mode( WIFI_MODE_NULL );
btStop();
```

PR #200, opened 2019-01-20
Arduino: dynamically create XML file

balazsracz:
Adds code to the arduino helper library to generate the XML representation of the CDI in the target, and save it to a local filesystem such as SPIFFS.
Extends the base library to allow dynamicaly created CDI event offsets.

### DIFF
--- a/arduino/OpenMRN.cpp
+++ b/arduino/OpenMRN.cpp
@@ -34,6 +34,9 @@
 
 #include <OpenMRN.h>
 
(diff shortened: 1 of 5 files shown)


balazsracz:
@atanisoft This PR makes the configuration options show up in JMRI correctly. (Need your PR as well in order to create the writable config file.)

atanisoft:
I copied this over to my branch and I can now see the data showing up in JMRI, there are a few performance issues but things are functional.

PR #203, opened 2019-01-20
Rewrites libify to allow creating symbolic links

balazsracz:
Rewrites libify to allow creating symbolic links for the exported library.


balazsracz:
@atanisoft PTAL

PR #205, opened 2019-01-20
esp32 write performance

balazsracz:
Optimizes write performance for the esp32 sockets:
- Adds a more efficient implementation for availableForWrite() that uses select() to identify whether the socket is writeable or not.
- Adds configuration overrides for gridconnect params to buffer entire datagram sends in one packet.
- Increases the number of Executor loops per arduino loops.

### DIFF
--- a/src/freertos_drivers/arduino/Esp32WiFiClientAdapter.hxx
+++ b/src/freertos_drivers/arduino/Esp32WiFiClientAdapter.hxx
@@ -44,13 +44,47 @@ public:
     Esp32WiFiClientAdapter(WiFiClient client) : client_(client){
         client_.setNoDelay(true);
     }
+
+    /// This is how many bytes we return as writeable when select says the
+    /// socket is write active.
+    static constexpr unsigned WRITE_PACKET_SIZE = 512;
+
     // on the ESP32 there is no TX limit method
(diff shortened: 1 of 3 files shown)


src/freertos_drivers/arduino/Esp32WiFiClientAdapter.hxx
```diff
@@ -44,13 +44,45 @@ public:
     Esp32WiFiClientAdapter(WiFiClient client) : client_(client){
         client_.setNoDelay(true);
     }
+
+    /// This is how many bytes we return as writeable when select says the
+    /// socket is write active.
+    static constexpr unsigned WRITE_PACKET_SIZE = 512;
```
atanisoft: this doesn't appear to be used in availableForWrite()

balazsracz:
> considerably, it does still take some time for the config panel to open but there is a lot of data being transmitted.

Actually the time it takes to download the CDI now is on par with what I see for other network-connected devices. Because there is no queueing or batching of requests (strictly one outstanding request per target), the throughput is defined by the roundtrip latency. That's really good for CAN-bus connected devices, but usually 10x to 100x worse for wifi, depending on the power saving mode.

PR #206, opened 2019-01-24
ESP32 hardware CAN support and more documentation

atanisoft:
This PR provides the ESP32 hardware CAN driver support as well as documentation for most of the ESP32 code. Still to be done is provide a "how to" guide for the Arduino library creation.

ESP32 hardware CAN support and more documentation.
- Adds hardware CAN driver
- Updates IO board to show how to use hardware CAN driver
- Removes start_background_task from the OpenMRN class because it does not work well yet.
- Adds WifiCan sketch
- adds readme.md for arduino directory

Misc fixes:
- IO board GPIO pins and producers polling are fixed
- Adds factory reset routine to sketches to show node name
- Node hardware name will now contain can/wifi depending on what is enabled
- Reformat code
- Adds verbose option to libify.sh
- Fixes SNIP not displaying node user name /description on the ESP32


atanisoft:
@bakerstu @balazsracz PTAL

atanisoft:
@TrainzLuvr these are only examples, but I can update them to include ESP32 in the name field, it is also included in the Hardware field today, you can see an example of this here: https://photos.app.goo.gl/vWKPpHTJguQ4FsEj8

TrainzLuvr:
I guess I was being a bit picky. :)

atanisoft:
> I guess I was being a bit picky. :)

Picky isn't always a bad thing. As you encountered a few issues and reported them via the OpenLCB list I'm going to start writing more documentation on the connections for this so we can avoid some common pitfalls. Expect that in a PR at some point in the near future.

TrainzLuvr:
I got one more thing, not necessarily related to CAN but ESP32, and not sure if it's a bug or I'm doing something wrong.

I went and changed the "User name" in Configuration Dialog for this node in JMRI, yet it does not show up in the Network Tree next to the NodeID.

Reloading the Configuration Dialog I see the set User name and description though.

atanisoft:
That sounds like a bug but I'm not sure if it is in the esp32 side or jmri...

@balazsracz, thoughts?

TrainzLuvr:
Actually, that's easy to check (from JMRI OpenLCB Traffic Monitor) :

R: 05.01.01.01.xx.23 - 02.01.xx.xx.xx.FC Simple Node Ident Info with content '4,OpenMRN,ESP32 Arduino IO Board,esp32,1.00,2,,,'

So the 'User name' is missing - it's the parameter after the '2' in '4,OpenMRN,ESP32 Arduino IO Boardesp32,1.00,2,(missing User name),,'

(I took the liberty to change value in the model_name field in config.h to 'ESP32 Arduino IO Board'. Never the less it can't be that causing it since it's defined as _const char model_name[41]_)

atanisoft:
The esp32 code is just a transport layer built with OpenMRN API. What I suspect is happening is the SNIP data is not being updated in memory or reported after update. Not really sure which. Are you connected via CAN or WiFi?

TrainzLuvr:
I'm using CAN with VP230.

Looking at some of the other boards' main.cxx files from OpenMRN, do you need to define the classes below for ESP32, as they deal with cfg.userinfo().name().read/write:

```
class CustomFactoryReset : public DefaultConfigUpdateListener {
...
}
```
and
```
class ConfigApply : public DefaultConfigUpdateListener {
...
}
```

atanisoft:
Currently these are not defined in the examples, I'll take a look at them and add them to them. We want to have these examples be as near complete as possible so users can use them with few (if any) issues.

TrainzLuvr:
Another thought I had was to set in the model_name (config.h) of the node whether the firmware is running WiFi or CAN. Ideally that would come from USE_CAN or USE_WIFI.

The model-name would look something like: 
_ESP32 Arduino IO Board (WiFI)_
or
_ESP32 Arduino IO Board (CAN)_

atanisoft:
The only problem I see with that is you can enable both simultaneously and not knowing which to add to the SNIP.

TrainzLuvr:
Then add both? :)

It goes back to me drumming about UX/UI. A user should be able to identify the way each node is connected to the network.

This could come helpful for troubleshooting, for example, or just to understand the topography of the network, and segmentation between wired and wireless portions.

atanisoft:
I can do that as long as the length is under the length limit.

If a node is WiFi is possibly a hub as well, I don't think that is displayed anywhere today either.

TrainzLuvr:
I'll add this as a separate issue as well then.

Perhaps they will consider adding a few more characters to the model-name, or even better extending the SNIP struct by adding a single byte variable whose bits define networking options (CAN, WiFi, HUB,...) allowing for multiple bits to be set.

atanisoft:
@TrainzLuvr I've added the FactoryResetHelper to the examples and updated the SNIP model_name to display connection types.

TrainzLuvr:
Thanks, I see it now, under model it says (CAN) at the end for me! And all the garbage characters are also gone from the I/O segment.

Unfortunately, the User name still does not show up next to the Node ID in the network tree.

atanisoft:
@TrainzLuvr you will need to erase the flash / spiffs to force the regeneration of the config, you can also do this via updating the CANONICAL_VERSION in config.h

TrainzLuvr:
Ok done that, still no change though.

atanisoft:
I'll have a fix in a few mins :)

TrainzLuvr:
It's working now !!

Is it normal that both 'user name' and 'description' show up, I thought only 'user name' was supposed to show up next to the Node ID in the network tree?

atanisoft:
If there is a description JMRI will show it I believe.

arduino/OpenMRN.h
```diff
@@ -465,25 +476,36 @@ class OpenMRN : private Executable
     }
 
 #if defined(ESP32)
+
+    /// Creates a FreeRTOS task on the ESP32 to automatically call the @ref loop
+    /// method. If the ESP32 hardware CAN device is in use the background task
+    /// is strongly suggested to be used. @ref start_background_task.
```
balazsracz: The implementation of this is still broken (uses a fixed delay without taking into account whether there is work to be done). I would not recommend to use this now.

arduino/OpenMRN.h
```diff
@@ -465,25 +476,36 @@ class OpenMRN : private Executable
...
+
+        // Add this task to the WDT, NULL is passed in to have the
+        // WDT code find the task handle automatically
+        esp_task_wdt_add(NULL);
+
+        while (true)
         {
-            // Feed the watchdog so it doesn't reset the ESP32
-            esp_task_wdt_reset();
+            // it is not necessary to feed the WDT here as it is done internally
```
balazsracz: i think this comment can be safely removed.

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -25,69 +25,129 @@
...
+#include <WiFi.h>
 #include <vector>
 
 #include <OpenMRN.h>
 #include <openlcb/TcpDefs.hxx>
+
+#include "openlcb/MultiConfiguredConsumer.hxx"
+
+// Pick one mode below, if you select USE_WIFI it will expose this node on WIFI
+// if you select USE_CAN, this node will be available on CAN.
```
balazsracz: Can I pick both? Consider documenting.

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -135,7 +178,33 @@ openlcb::ConfiguredProducer IO14_producer(
...
+                                     BarrierNotifiable *done) OVERRIDE {
+        AutoNotify n(done);
+        return UPDATED;
+    }
+
+    void factory_reset(int fd) override
+    {
+        cfg.userinfo().name().write(fd, openlcb::SNIP_STATIC_DATA.model_name);
+        cfg.userinfo().description().write(
+            fd, "OpenLCB DevKit + Arduino-ESP32 on an ESP32 board.");
```
balazsracz: This does not have the devkit board underneath -- please remove that reference.

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -135,7 +178,33 @@ openlcb::ConfiguredProducer IO14_producer(
...
+        for(int i = 0; i < openlcb::NUM_OUTPUTS; i++)
+        {
+            cfg.seg().consumers().entry(i).description().write(fd, "");
+        }
+        for(int i = 0; i < openlcb::NUM_INPUTS; i++)
+        {
+            cfg.seg().producers().entry(i).description().write(fd, "");
+            cfg.seg().producers().entry(i).debounce().write(fd, 3);
```
balazsracz: CDI_FACTORY_RESET(cfg.seg().producers().entry(i).debounce);

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -215,30 +299,45 @@ void setup() {
     // have performance impact.
     // openmrn.stack()->print_all_packets();
 
-    // Start the TCP/IP listener
-    openMRNServer.setNoDelay(true);
-    openMRNServer.begin();
+#if defined(USE_CAN)
+    // When using the hardware can device it is recommended to also utilize the
```
balazsracz: no, it is most definitely not recommended.

arduino/OpenMRN.h
atanisoft: Is there a way to detect if work is to be done? Or do you think the loop method can always be called?

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
atanisoft: The problem with not calling the by task is that loop() runs at priority 0 and rx/tx at ~20. Loop() will be called much less often in this case, by using the BG task the OpenMRN loopethod will be called as often or more than RX task.

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
atanisoft: Yes, I'll update the comment as it does work for both.

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
atanisoft: I'll drop the BG task and do some further testing, if it is stable (likely) and not starving the rest of the stack of cycles (some possibility) then the BG task can go away.

arduino/OpenMRN.h
atanisoft: This won't be necessary as the BG task is going away. I'll have an update shortly with its removal.

atanisoft: task has been removed.

arduino/OpenMRN.h
atanisoft: removed

arduino/OpenMRN.h
atanisoft: no longer relevant as the bg task has been removed.

arduino/OpenMRN.h
atanisoft: removed

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
atanisoft: updated.

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
atanisoft: fixed

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
atanisoft: fixed for all examples

arduino/examples/ESP32WifiCanBridge/ESP32WifiCanBridge.ino
```diff
@@ -0,0 +1,243 @@
...
+
+    // Dump all packets as they are sent/received.
+    // Note: This should not be enabled in deployed nodes as it will
+    // have performance impact.
+    // openmrn.stack()->print_all_packets();
+
+    // When using the hardware can device it is recommended to also utilize the
+    // background task on the ESP32 rather than call openmrn.loop() from the
+    // loop method.
+    openmrn.start_background_task();
```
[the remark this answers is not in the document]
atanisoft: oops, you are right, this should be dropped, will clean up.

arduino/examples/ESP32SerialBridge/ESP32SerialBridge.ino
```diff
@@ -108,8 +151,17 @@ void setup() {
 
     // Add Serial1 as a bridge
     openmrn.add_gridconnect_port(new Esp32HardwareSerialAdapter(Serial1));
+
+    // When using the hardware can device it is recommended to also utilize the
+    // background task on the ESP32 rather than call openmrn.loop() from the
+    // loop method.
+    openmrn.start_background_task();
```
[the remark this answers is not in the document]
atanisoft: right, missed in cleanup.

arduino/OpenMRN.h
```diff
@@ -36,27 +36,34 @@
...
 #if defined(ESP32)
-/// Default stack size to use for the OpenMRN background task on the ESP32 platform.
-constexpr uint32_t OPENMRN_STACK_SIZE = 4096L;
 
-/// Default thread priority for the OpenMRN background task on the ESP32 platform.
-constexpr UBaseType_t OPENMRN_TASK_PRIORITY = tskIDLE_PRIORITY + 1;
+#include <esp_task.h>
+#include <esp_task_wdt.h>
+
+/// Default stack size to use for all OpenMRN tasks on the ESP32 platform.
```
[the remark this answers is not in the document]
atanisoft: this is in bytes, will make a note of it and get it documented with a ref to the idf docs for this.

arduino/readme.md
```diff
@@ -0,0 +1,125 @@
...
+1. 3v3 / 5v to the VCC/VDD pin on the transceiver. Make sure to use the right
+voltage for the transceiver.
+2. GND to the GND pin on the transceiver.
+3. GPIO for RX to the RX pin on the transceiver. This can be any unused GPIO
+pin.
+4. GPIO for TX to the TX pin on the transceiver. This pin must be usable as an
+output pin, GPIO 34-39 on the ESP32 are input only.
+
+## Powering the ESP32
+It is not recommended to power the ESP32 from the CAN PWR_POS (7) / PWR_NEG (8)
```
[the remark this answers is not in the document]
atanisoft: I'll reword this paragraph to include your comment as you are right, the spec allows for up to 500mA and the ESP32 may exceed this under some conditions and the PWR_POS/NEG per the spec are considerably higher than the typical ESP32 on-board regulator will tolerate.

arduino/OpenMRN.h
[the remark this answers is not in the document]
atanisoft: https://github.com/espressif/esp-idf/issues/2042 which references https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/freertos-smp.html#tasks-and-task-creation

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -25,73 +25,141 @@
...
-const char* ssid     = DEFAULT_WIFI_NAME;
+const char *ssid = DEFAULT_WIFI_NAME;
+
 /// Password of the wifi network.
-const char* password = DEFAULT_PASSWORD;
-const char* hostname = "esp32mrn";
+const char *password = DEFAULT_PASSWORD;
 
-static constexpr uint64_t NODE_ID = UINT64_C(0x050101011423);
+/// This is the hostname which the ESP32 will advertise via mDNS, it should be
```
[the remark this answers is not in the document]
atanisoft: Perhaps using this as the base name and appending the node alias like "esp32mrn-{alias}"?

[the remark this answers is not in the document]
atanisoft: I'd also like to move wifi into its own manager type class accessible via OpenMRN.h which can then handle this and possibly promoting the node as a hub if there are none available.

arduino/OpenMRN.h
```diff
@@ -36,27 +36,34 @@
...
+constexpr uint32_t OPENMRN_STACK_SIZE = 4096L;
 
-constexpr TickType_t OPENMRN_TASK_TICK_DELAY = pdMS_TO_TICKS(1);
+/// Default thread priority for any OpenMRN owned tasks on the ESP32
+/// platform. ESP32 hardware CAN RX and TX tasks run at lower priority
+/// (-1 and -2 respectively) of this default priority to ensure timely
```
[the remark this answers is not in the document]
atanisoft: https://github.com/bakerstu/openmrn/issues/212

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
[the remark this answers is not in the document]
atanisoft: https://github.com/bakerstu/openmrn/issues/213

arduino/OpenMRN.h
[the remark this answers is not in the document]
atanisoft: https://github.com/bakerstu/openmrn/issues/214

arduino/libify.sh
```diff
@@ -50,24 +63,35 @@ done
...
     REL_DIR=$1
     shift
+    if [ "x$VERBOSE" != "x" ]; then
+        echo "Creating ${TARGET_LIB_DIR}/${REL_DIR}"
+    fi
     mkdir -p ${TARGET_LIB_DIR}/${REL_DIR}
     pushd ${TARGET_LIB_DIR}/${REL_DIR} >/dev/null
     while [ "x$1" != "x" ] ; do
+        if [ "x$VERBOSE" != "x" ]; then
+            echo "cp -fax ${USE_LINK} ${OPENMRNPATH}/${1} ${TARGET_LIB_DIR}/${REL_DIR}"
```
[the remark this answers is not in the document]
atanisoft: fixed

arduino/readme.md
```diff
@@ -0,0 +1,125 @@
+# OpenMRN-lite Arduino Library
+The OpenMRN-lite Arduino Library is a packaged version of the OpenMRN code that
+has been designed to run in the Arduino environment. The primary features of the
+OpenMRN library are available through this library.
+
+## Supported platforms
+At this time the only supported platform for execution is the ESP32 using the
```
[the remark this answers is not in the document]
atanisoft: added a note for this, some examples for how to use this would be good to add but I'm not sure what would be good to pull in.

arduino/readme.md
```diff
@@ -0,0 +1,125 @@
+# OpenMRN-lite Arduino Library
+The OpenMRN-lite Arduino Library is a packaged version of the OpenMRN code that
+has been designed to run in the Arduino environment. The primary features of the
```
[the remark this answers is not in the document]
atanisoft: fixed

arduino/readme.md
```diff
@@ -0,0 +1,125 @@
...
+    [D][WiFiGeneric.cpp:342] _eventCallback(): Event: 0 - WIFI_READY
+    [D][WiFiGeneric.cpp:342] _eventCallback(): Event: 2 - STA_START
+    [D][WiFiGeneric.cpp:342] _eventCallback(): Event: 2 - STA_START
+    [D][WiFiGeneric.cpp:342] _eventCallback(): Event: 5 - STA_DISCONNECTED
+```
+
+If you observe this output this generally means there was a timeout condition
+where the ESP32 did not receive a response from the access point. This timeout
+is unfortuntely not configurable at this time. It is not known if this is due
+to a poor signal quality or an underlying bug in the ESP-IDF WiFi driver. The
```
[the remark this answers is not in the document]
atanisoft: changed to strength, it does make more sense than quality.

3dluvr:
Thank you for all the work on this.

TrainzLuvr:
Thank you for all the work on this.

balazsracz:
> It's working now !!
> 
> Is it normal that both 'user name' and 'description' show up, I thought only 'user name' was supposed to show up next to the Node ID in the network tree?

yes they both show up if not empty.

TrainzLuvr:
I think something got broken at the end, I cannot see my ESP32 node anymore in the network tree?!

I pulled the latest OpenMRN master and compiled the example sketch, changing the define to USE_CAN.

And the other branch (atanisoft:hwcan) is now gone, deleted in hasted I must say.

atanisoft:
I pulled down the latest master and it was working. Do you have the console output? How are you connected to the node?

balazsracz:
The branches' merge points are permanently recorded in the git history as
the parents of the merge commit.

TrainzLuvr:
I'm connected via RR-CirKits LCC Repeater and have LCC Buffer-USB going into JMRI. My usual test setup I've been using past week or so.

This is the console after a reset:
```
rst:0x1 (POWERON_RESET),boot:0x17 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0018,len:4
load:0x3fff001c,len:1100
load:0x40078000,len:10088
load:0x40080400,len:6380
entry 0x400806a4
ESP32-CAN: rx-q:0, tx-q:0, rx-err:0, tx-err:0, arb-lost:0, bus-err:0, state: STOPPED
Allocating new alias 861 for node 05010101xx23
ESP32-CAN: rx-q:0, tx-q:0, rx-err:0, tx-err:0, arb-lost:0, bus-err:0, state: RUNNING
ESP32-CAN: rx-q:0, tx-q:0, rx-err:0, tx-err:0, arb-lost:0, bus-err:0, state: RUNNING
```

TrainzLuvr:
I'll try to find the right HEAD then and see if it makes any difference.

atanisoft:
Latest master on this repo should be fine. The console output looks ok. We may need to adjust log levels, @balazs is there an easy way to set log level to verbose without editing the hxx file?

balazsracz:
Log level is compilation time option in order to optimize away the log
statements if compiling without logging. So for logging you need to
recompile and reflash.
I would recommend enabling the packet printer first.

TrainzLuvr:
Actually I can't go back in this master, because I was using the atanisoft:hwcan branch to test. :(

Did not change anything in my setup - neither software or hardware. Only been pulling changes from the hwcan branch until this PR got merged, then pulled OpenMRN master and libified it.

That's where I am now.

atanisoft:
Something may be amiss then. I'd say download the latest master from here as it has everything. I tested WiFi and CAN a short bit ago and didn't see issues with it, but that doesn't mean there aren't any.

Packet printing would be a good start, the extra log levels are mostly diagnostics in the can adapter code.

TrainzLuvr:
Meh, so I went and added a Tower-LCC to the network and now the ESP32 node showed up.

Both are connected to the same side of the LCC Repeater (C+D). The other side (A+B) is the LCC Buffer-USB and the Railcom IO board (which doesn't register as it needs to be flashed but is terminated).

atanisoft:
Ok that's weird. And termination is in place on both sides as needed?

TrainzLuvr:
Yes, I terminated the ESP32 with a resistor originally, did not change it since.

TrainzLuvr:
ARGH! I'm sorry guys for raising an alarm. It's Windows 8.1 acting up. :-1: 

I switched the USB cables around my front panel ports and now I see the ESP32 node as well as the Tower-LCC.

I have a DCC++ connected (which also powers the Railcom board), then there's the ESP32 node, and lastly LCC Buffer-USB. It seems if I don't plug each USB cable back to a specific ports, whole hell breaks lose.

On top of that, I must make sure that the TIVA on the Railcom board is not plugged into USB, unless I disconnect Railcom board from DCC++ power, else the Arduino Motor Bridge goes crazy.

dpharris:
Yikes ...

On Fri, Feb 1, 2019, 20:12 TrainzLuvr <notifications@github.com wrote:

> On top of that, I must make sure that the TIVA on the Railcom board is not
> plugged into USB, unless I disconnect Railcom board from DCC++ power, else
> the Arduino Motor Bridge goes crazy.
>
Danger, Will Robinson!  Usb shouln't be plugged into Railcom board ... its
ground is special!  Use isolated power.

At least how I understand it.

David

atanisoft:
I have the same issue with the USB devices setting up new com ports, I think I'm up to com18 as my LCC buffer.... I need to get a USB hub and a few more cables so I can debug a full network of nodes at some point ...

TrainzLuvr:
Railcom board is a creature on its own, for sure. I am having problems getting it to flash and accept my Node ID, so it's dead in the water right now.

Doing "make ADDRESS=0xXY flash" works but then "make ADDRESS=0xXY rflash" fails...and I don't know what's the catch or what kind of a setup do I need to have. Doing all this from a VM.

PR #208, opened 2019-01-27
Adding defines for EMERGENCY_STOP_EVENT and CLEAR_EMERGENCY_STOP_EVENT.

atanisoft:
These are defined in the well-known events spec but are not defined
currently in the defs, I've added these and a short description (may
need edits). These will primarily be used by a command station that
supports LCC.

### DIFF
--- a/src/openlcb/Defs.hxx
+++ b/src/openlcb/Defs.hxx
@@ -221,7 +221,12 @@ struct Defs
     /// track power.
     static constexpr uint64_t CLEAR_EMERGENCY_OFF_EVENT = 0x010000000000FFFEULL;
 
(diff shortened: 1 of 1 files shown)


src/openlcb/Defs.hxx
```diff
@@ -221,6 +221,15 @@ struct Defs
     /// track power.
     static constexpr uint64_t CLEAR_EMERGENCY_OFF_EVENT = 0x010000000000FFFEULL;
 
+    /// Producing this event causes a node to revert to safe output defaults.
```
balazsracz: please also put in the standard's text (in quote if you'd like to)

PR #218, opened 2019-02-01
Disable mDNS broadcasting for the ESP32IOBoard by default.

atanisoft:
By default the ESP32IOBoard should not broadcast over mDNS as this
should only be done for nodes which are acting as a hub. The
ESP32IOBoard can be a hub for other nodes to connect to via the
advertised mDNS service details.

### DIFF
--- a/arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
+++ b/arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
@@ -53,9 +53,9 @@
 #define USE_WIFI
 // #define USE_CAN
 
-// uncomment the line below to have this node advertise itself via mDNS as a
+// Uncomment the line below to have this node advertise itself via mDNS as a
 // hub. When this is enabled, other devices can find and connect to this node
-// via mDNS treating it as a hub. Note this requires USE_WIFI to be enabled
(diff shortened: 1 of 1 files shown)


arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -344,7 +350,7 @@ void setup()
         new Esp32HardwareCan("esp32can", CAN_RX_PIN, CAN_TX_PIN));
 #endif // USE_CAN
 
-#if defined(USE_WIFI)
+#if defined(USE_WIFI) && defined(BROADCAST_MDNS)
```
balazsracz: I would think that MDNS.begin would be inside this if def?

atanisoft: It is outside of this section to allow the node's hostname to be registered for lookup. This is also the first step towards an updated version which will connect the client nodes to a hub.

balazsracz: So MDNS.begin exports some other service (in addition to the once in this block)?

atanisoft: Yes, it starts the mDNS resolver which will be needed for clients to find the hub nodes that are broadcasting.

PR #219, opened 2019-02-01
macOS and os_get_time_monotonic changes 

kiwi64ajs:
added macOS specific support to the libify.sh shell script
changed os_get_time_monotonic() to just call the Arduino micros() function to improve precision and added a prototype to suppress the warning

### DIFF
--- a/arduino/libify.sh
+++ b/arduino/libify.sh
@@ -3,6 +3,8 @@
 # 
 #
 
+#set -x
+
 function usage() {
     echo
     echo 'usage: libify.sh path/to/arduino/library/output path/to/openmrn [-f] [-l]'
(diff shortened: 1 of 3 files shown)


src/os/os.c
```diff
@@ -618,17 +618,10 @@ long long os_get_time_monotonic(void)
...
-        ++overflow_millis;
-    }
-    time = overflow_millis;
-    time <<= 32;
-    time += new_millis;
-    time *= 1000000;
+	// redeclare micros() prototype to keep the compiler happy
+	unsigned long micros();
+	
+	time = micros();
```
balazsracz: 1) you definitely need to keep the overflow handling for microseconds because those will overflow rather sooner than millis.
2) time is in nanoseconds so you need to do time *= 1000.

arduino/libify.sh
```diff
@@ -92,7 +119,13 @@ function copy_dir() {
     if [ "x$VERBOSE" != "x" ]; then
         echo "${OPENMRNPATH}/${2} ==> ${TARGET_LIB_DIR}/$1"
     fi
-    cp -faxr ${USE_LINK} ${OPENMRNPATH}/$2 .
+
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+	    cp -fa ${USE_LINK} ${OPENMRNPATH}/$2 .
+	else
```
balazsracz: please use spaces instead of tabs, because the indentation will be messed up.

kiwi64ajs:
Ok, I've:
- changed the tags to 4 spaces
- reinstated the micros() overflow detection
- converted micros() to uint_32
- updated last_micros after comparison to new_micros as it wasn't actually being saved

PR #227, opened 2019-02-10
Make the bootloader behave better when doing just CRC.

balazsracz:


TrainzLuvr:
I guess this answers my question why a connection to the board would be needed, and what are those bytes in the diff between railcom.bin and railcom.upload. :)

Thanks for the fix.

balazsracz:
Yes, these crc bytes ensure that we only boot the firmware if the download has been completed. If the power is interrupted during download, the bootloader will never vector to the new firmware.

PR #231, opened 2019-02-19
add tristate to API.

bakerstu:


balazsracz:
is there a reason yo use a different state enum than EventState ?

bakerstu:
Yes, EventState is inside namespace openlcb, however ::NetworkState is not, it is agnostic of namespace openlcb.

PR #234, opened 2019-03-02
Executor implementation for esp32

balazsracz:
- Refactor os.h to have the pieces of code conditionally compiled using feature bits instead of list of operating systems.
- Enable mutex and semaphore implementation for ESP32.
- Adds a ::select based executor sleep method (which needs custom wakeup)
- Adds wakeup code for esp32 using the esp_vfs's select implementation that their device drivers use.
- Adds start_background_task for the arduino API when using the ESP32.
- Adds workaround for endian.h bug in ESP32 SDK.


balazsracz:
> you can see my version of this here [ff6ef77](https://github.com/bakerstu/openmrn/commit/ff6ef772f41d5ed889770bf2cb2eb25ea2043337)

I merged that commit to master.

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -37,6 +37,7 @@
 #include <SPIFFS.h>
 #include <WiFi.h>
 #include <vector>
+#include "../vfs/esp_vfs_dev.h"
```
atanisoft: This include can likely be dropped

balazsracz: done

src/os/TempFile.cxx
```diff
@@ -43,7 +43,8 @@
 
 #include "os/TempFile.hxx"
 
-#ifndef __FreeRTOS__
+/// @todo mingw does not seem to have an mkdtemp call.
+#if !defined(__FreeRTOS__) && !defined(__WINNT__)
```
bakerstu: Could we handle this through openmrn_features.h?

src/os/TempFile.hxx
```diff
@@ -129,7 +129,9 @@ public:
       HASSERT(ret >= 0);
       ofs += ret;
     }
+#ifndef __WINNT__
```
bakerstu: same.

src/utils/macros.h
```diff
@@ -225,5 +226,13 @@ extern const char* g_death_file;
 /// Macro to signal a function that the result must be used.
 #define MUST_USE_RESULT __attribute__((__warn_unused_result__))
 
+#ifdef ESP32
```
bakerstu: Could we handle this through openmrn_features.h

arduino/OpenMRN.h
```diff
@@ -355,6 +354,23 @@ class OpenMRN : private Executable
...
 
+    void start_executor_thread()
+    {
+        haveExecutorThread_ = true;
+        xTaskCreate(openmrn_background_task, "OpenMRN", OPENMRN_STACK_SIZE,
+                    this, OPENMRN_TASK_PRIORITY, nullptr);
+        /// @todo: replace the above with the below once os_thread_create gets
+        /// implemented for the esp32.
+        // stack_->executor()->start_thread(
+        //     "OpenMRN", OPENMRN_TASK_PRIORITY, OPENMRN_STACK_SIZE);
```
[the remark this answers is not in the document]
atanisoft: This method should be locked to only FreeRTOS supported envs likely. In general Arduino doesn't support multi-threaded code except for a few usages.

src/os/OS.hxx
```diff
@@ -138,10 +138,24 @@ public:
...
+    os_thread_t get_handle()
+    {
         return handle;
     }
 
+    /// Sets the thread handle to the current calling thread's. Needed for
+    /// multiplexing environments like Arduino.
+    void lock_to_thread()
+    {
+        handle = os_thread_self();
```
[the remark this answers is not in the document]
atanisoft: Maybe, but it may trigger unnecessarily on the esp32 depending on call flow.

src/utils/macros.h
atanisoft: Perhaps, this is more of a temp fix until esp-idf updates for Balazs' PR. Perhaps remove the esp32 check and instead check for definition of the method only instead?

src/os/TempFile.cxx
atanisoft: Where is this used? We may want to exclude it entirely for some envs beyond the two above

balazsracz:
> Overall looks pretty good. Couple of general pieces of feedback.
> 
> Should we do a grep of the source for lingering conditional compiles that are NOT part of the new features defines and make sure that each one is fully justified?

Eventually yes, but it will be a ton of work and a lot of PRs. This is just scratching the surface.

> 
> Could we consider not making class OpenMRN in arduino/OpenMRN.h part of the global namespace. Suggestions include renaming the class or transparently wrapping in a namespace followed by "using namespace ...".

will make a separate PR for that.

src/utils/macros.h
balazsracz: This is a specific workaround for a specific bug in the ESP32 SDK. I am not aware of this bug existing anywhere else.

src/freertos_drivers/common/SPI.hxx
```diff
@@ -34,8 +34,10 @@
 #ifndef _FREERTOS_DRIVERS_COMMON_SPI_HXX_
 #define _FREERTOS_DRIVERS_COMMON_SPI_HXX_
 
+#ifndef _DEFAULT_SOURCE
```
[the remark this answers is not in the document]
balazsracz: this file currently does not compile on certain targets.

filed https://github.com/bakerstu/openmrn/issues/236 for the cleanup task.

src/os/OS.hxx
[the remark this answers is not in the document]
balazsracz: done

src/os/TempFile.cxx
balazsracz: i moved this change out of this PR because it is unrelated.

src/os/TempFile.hxx
balazsracz: i moved this change out of this PR because it is unrelated.

src/freertos_drivers/common/SPI.hxx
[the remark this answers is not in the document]
balazsracz: i moved this change out of this PR because it is unrelated.

arduino/OpenMRN.h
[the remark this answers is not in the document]
balazsracz: added missing `#ifndef OPENMRN_FEATURE_SINGLE_THREADED`

PR #235, opened 2019-03-02
Simplify task creation code in os.c

balazsracz:
Removes a lot of code duplication from this file.
Removes support for STATIC_ALLOCATION because openmrn is using malloc and cannot live without it.


balazsracz:
> This looks okay, but it will need to be reconciled with the cleanup done with TI SimpleLink. I would have rather we merged that back first before taking this on.

What we could do with TI simplelink is to switch from OPENMRN_FEATURE_FREERTOS_THREAD to OPENMRN_FEATURE_PTHREAD and then none of this code will compile.

bakerstu:
That is not really the problem.  The problem is that I already cleaned up the same code lines on my branch that brings in the TI SimpleLink Pthread.

PR #239, opened 2019-03-03
Wraps arduino objects in namespace openmrn_arduino {}.

balazsracz:
@atanisoft if you could review that would be sufficient as well.

### DIFF
--- a/arduino/OpenMRN.cpp
+++ b/arduino/OpenMRN.cpp
@@ -41,8 +41,11 @@
 extern const char DEFAULT_WIFI_NAME[] __attribute__((weak)) = "defaultap";
 extern const char DEFAULT_PASSWORD[] __attribute__((weak)) = "defaultpw";
 
+namespace openmrn_arduino {
 
 OpenMRN::OpenMRN(openlcb::NodeID node_id)
 {
     init(node_id);
 }
+
+} // namespace openmrn_arduino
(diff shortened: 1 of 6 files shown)


balazsracz:
> Overall it looks good. Two points (for diff PRs)
> 
> 1. I think the Esp32HardwareSerialAdapter may be able to be promoted to HardwareSerialAdapter as it is mostly generic already. This could also extend to supporting SoftwareSerial for board types that require that. Not sure if this is necessary though.

that sounds fine. Don't we already have a serialadapter in OpenMRN.h?

> 2. DEFAULT_WIFI_NAME and DEFAULT_PASSWORD should likely be moved inside the ESP32 block since it is only used there. We may also want to consider a rename of these two vars to include OPENMRN in the name and change DEFAULT_PASSWORD to DEFAULT_WIFI_PASSWORD (consistency in naming).

Adding _WIFI for consistency should be fine; I would not however add OPENMRN into the name to keep it simpler for users. (e.g. avoid a possible typoe)

atanisoft:
> that sounds fine. Don't we already have a serialadapter in OpenMRN.h?

there is SerialBridge that takes a SerialType, the Esp32HardwareSerialAdapter is used when calling OpenMRN::add_gridconnect_port(). I want to say the reason this is used on the esp32 was due to type conflicts in the code, I'll need to double check though. It would be better to drop the adapter class if possible so we can do OpenMRN::add_gridconnect_port(&Serial) or OpenMRN::add_gridconnect_port(softSerial);

PR #240, opened 2019-03-03
Remove #pragma -Wunused-private-field override.

balazsracz:
@atanisoft PTAL

### DIFF
--- a/src/openlcb/SimpleInfoProtocol.hxx
+++ b/src/openlcb/SimpleInfoProtocol.hxx
@@ -396,23 +396,21 @@ private:
     }
 
     /** Configuration option. See constructor. */
-    unsigned maxBytesPerMessage_ : 8;
+    uint8_t maxBytesPerMessage_;
     /** Configuration option. See constructor. */
-    unsigned useContinueBits_ : 1;
+    uint8_t useContinueBits_ : 1;
 
     /** Whether this is the first reply message we are sending out. Used with
      * the continuation feature. */
-    unsigned isFirstMessage_ : 1;
+    uint8_t isFirstMessage_ : 1;
     /** Tells which descriptor entry we are processing. */
-    unsigned entryOffset_ : 5;
-#pragma clang diagnostic ignored "-Wunused-private-field"
-    unsigned reserved_ : 1; // for alignment
+    uint8_t entryOffset_ : 5;
 
     /** Byte offset within a descriptor entry. */
-    unsigned byteOffset_ : 8;
+    uint8_t byteOffset_;
     /** Total / max length of the current block. This is typically strlen() + 1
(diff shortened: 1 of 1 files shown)


src/openlcb/SimpleInfoProtocol.hxx
```diff
@@ -396,23 +396,21 @@ private:
...
 
     /** Whether this is the first reply message we are sending out. Used with
      * the continuation feature. */
-    unsigned isFirstMessage_ : 1;
+    uint8_t isFirstMessage_ : 1;
     /** Tells which descriptor entry we are processing. */
-    unsigned entryOffset_ : 5;
-#pragma clang diagnostic ignored "-Wunused-private-field"
-    unsigned reserved_ : 1; // for alignment
+    uint8_t entryOffset_ : 5;
```
atanisoft: Is the alignment issue not really an issue?

src/openlcb/SimpleInfoProtocol.hxx
```diff
@@ -396,23 +396,21 @@ private:
     }
 
     /** Configuration option. See constructor. */
-    unsigned maxBytesPerMessage_ : 8;
+    uint8_t maxBytesPerMessage_;
     /** Configuration option. See constructor. */
-    unsigned useContinueBits_ : 1;
+    uint8_t useContinueBits_ : 1;
```
atanisoft: would using bool type be better for this single bit usage?

src/openlcb/SimpleInfoProtocol.hxx
atanisoft: we could also drop : 5 if switching to bool for the other two single bit usages

src/openlcb/SimpleInfoProtocol.hxx
balazsracz: bool is 32 bits.

src/openlcb/SimpleInfoProtocol.hxx
balazsracz: bool is 32 bits.
alignment is provided by the next element being a "full" type instead of a bitfield.

src/openlcb/SimpleInfoProtocol.hxx
atanisoft: ahh, good point. that is also likely platform specific so best to keep as uint8_t, perhaps drop the single bit usage and keep as uint8_t in that case?

balazsracz: a single bit unsigned is perfectly fine for this use case (that's what it used to be before this PR).
I don't see a reason to change.

atanisoft: that works for me.

Issue #237, opened 2019-03-03
generalize vTaskSuspendAll()


balazsracz:
Currently we use in a couple of key locations a mutex-like construct:
```
   vTaskSuspendAll();
   // ...some code...
  vTaskResumeAll();
```

This behaves essentially like a single global mutex or critical section. The goal is to prevent any other task to get swapped in while this code is running.

We need to refactor this construct into a utility class that looks like an OSMutex or an Atomic. (Also while at it, should create a RAII class to hold it).

The main reason is that this construct is incompatible with multi-processor FreeRTOS and right now we are running the risk of race conditions involving these constructs.

PR #241, opened 2019-03-03
Thread creation for ESP32

balazsracz:
- Adds selection feature macro for which os_thread_create implementation should be compiled.
- Updates arduino esp32 version to use os_thread_create for creating the background thread.

### DIFF
--- a/arduino/OpenMRN.h
+++ b/arduino/OpenMRN.h
@@ -364,18 +364,8 @@ public:
     void start_executor_thread()
     {
         haveExecutorThread_ = true;
-        xTaskCreate(openmrn_background_task, "OpenMRN", OPENMRN_STACK_SIZE,
-                    this, OPENMRN_TASK_PRIORITY, nullptr);
-        /// @todo: replace the above with the below once os_thread_create gets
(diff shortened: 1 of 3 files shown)


balazsracz:
> For the esp32 we can't suspend the task manager, it could lead to a deadlock. Can we add a guard for TaskList management to disable it for esp32?

Right now it seems to be working fine. This code should be replaced with a MUX under the esp32, and I filed an issue to do that: https://github.com/bakerstu/openmrn/issues/237

> 
> Task tag is also not guaranteed to exist on esp32,

The task tag code is conditioned with the FREERTOS_DEVICE_SELECT which is not set in esp32, so it will not get compiled.

> it wasn't when I was doing this update previously. I had to add an esp32 specific call for it. Perhaps we can abstract this out of os_thread_start and os_thread_create so the logic can be in one place (including the FreeRTOS version specific ifdef). Or can we leverage pthread storage instead?

unnecessary

atanisoft:
Ok let me pull this in and do some tests as well. With your updates above it may just work. The task manager suspension may also be tied to only one core, that would likely need to be confirmed as well.

Issue #242, opened 2019-03-03
arduino: including library in sketch includes unnecessary header includes


atanisoft:
When adding the libify.sh generated library to the Arduino IDE and creating a new sketch which includes the library the following is generated as the sketch:
```c++
#include <can_frame.h>
#include <endian.h>
#include <freertos_includes.h>
#include <nmranet_config.h>
#include <OpenMRN.h>
#include <openmrn_features.h>

void setup() {
  // put your setup code here, to run once:

}

void loop() {
  // put your main code here, to run repeatedly:

}
```
We should add the following line to the library.properties file:
```
includes=OpenMRN.h
```

With this line added the Arduino IDE will only add this one include.

atanisoft:
fixed in #243

PR #243, opened 2019-03-03
Arduino Lite library

balazsracz:
renames the arduino library to OpenMRNLite.
- Renames includes to OpenMRNLite.h 
- moves default Node IDs to the 0501010118 range which is reserved for OpenMRN. Makes the Node IDs disjoint in the example sketches.
- updates symbol name to DEFAULT_WIFI_PASSWORD for consistency.
- sets version to 0.1.0, authors and maintainers in the library.properties file.


atanisoft:
Note this also fixes https://github.com/bakerstu/openmrn/issues/242.

Issue #244, opened 2019-03-03
Esp32: node reboot from jmri doesn't reboot the node


atanisoft:
When in the CDI pane for an esp32 node clicking more then reboot does nothing.

Most likely this just needs a platform specific hook to be put in place but I'm unsure where exactly to put it.

balazsracz:
This is done by an extern "C" function called `reboot()`
https://github.com/bakerstu/openmrn/blob/master/boards/ti-bracz-acc3/HwInit.cxx#L111

balazsracz:
If there is a standard arduino call to reboot the MCU, then we could put that function into OpenMRNLite.cpp.

atanisoft:
I'm not aware of a generic reboot function for all Arduino platforms, on the ESP32 it can be done through a couple approaches: ESP.restart() or esp_restart() (IDF). I'm leaning towards the IDF function since that is what the ESP.restart() calls and we would avoid an include of Arduino.h.

balazsracz:
sounds like it needs to be conditionally compiled for the esp32 then.
I don't see a problem in including Arduino.h into OpenMRNLite.cpp. That does not poison anybody's namespace.

atanisoft:
will namespace play any factor for the C function?

balazsracz:
instead of a namespace you have to wrap it in `extern "C" {` like here:
https://github.com/bakerstu/openmrn/blob/master/boards/ti-bracz-acc3/HwInit.cxx#L89

atanisoft:
Spotted the need for the extern "C" block and tested within the openmrn_arduino namespace and it works as expected now. See PR #247.

PR #245, opened 2019-03-04
Move FileUtils to a cxx file

balazsracz:
This file contains implementation of functions so by the ODR it should be a cxx file.


atanisoft:
with this being merged we should update https://github.com/bakerstu/openmrn/blob/master/arduino/OpenMRNLite.h#L75 to not define these functions here but instead include the header file.

balazsracz:
go for it

atanisoft:
I'll take a look at that and get #244 into a single PR while working on the wifi piece for a diff PR

PR #249, opened 2019-03-05
ESP32 WiFi Manager

atanisoft:
This PR implements an ESP32 specific WiFi manager which simplifies the WiFi/hub/uplink management on the ESP32. With this PR the Esp32WiFiClientAdapter.h is being removed from OpenMRNLite.h as it is not necessary to include. Esp32WiFiClientAdapter.h should likely be removed from the tree prior to merge.

@balazsracz can you review as there seems to be an issue with the esp32 acting as a hub and CDI load failing consistently at 512 bytes. I did not see anything in the serial console to explain the hang up but JMRI fails to load at this byte mark when connected to the ESP32 as a hub over TCP/IP.


src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    bool are_ssid_profiles_empty();
+
+    /// Retrieves a list of available access points.
+    ///
+    /// @param entries will contain a list of available network entries.
+    /// @param count is the number of network entries to retrieve, max of 20.
+    /// @return number of network entries retrieved.
+    int get_network_list(NetworkListEntry *entries, size_t count);
+
+    /// Initiate scanning for available access points.
```
balazsracz: consider moving this above the "get list" entry.

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    ///
+    /// @param mac is a 6 byte array which will hold the resulting MAC address.
+    /// @param mode is the WiFi interface to retrieve the MAC address for, if
+    /// an interface does not support the passed mode it will set the mac to
+    /// all zeros.
+    /// @return zero for succesful retrieval of the MAC address, non-zero for
+    /// failure (ie: mode not supported).
+    int get_mac(uint8_t *mac, WiFiMode mode = WiFiMode::STATION);
+
+    /// Returns the IPv4 address assigned to the WiFi iterface.
```
balazsracz: document endianness

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    /// Note: This method can return zero if the passed mode is not supported
+    /// or the WiFi interface does not have an IPv4 address for a supported
+    /// WiFi WiFiMode.
+    ///
+    /// @param mode is the WiFi interface to return the IPv4 address for.
+    /// @return the IPv4 address for the WiFi interface for the mode provided.
+    uint32_t get_ip(WiFiMode mode = WiFiMode::STATION);
+
+    /// @return SSID of the access point we are connected to, can return
+    /// nullptr if not currently connected to an access point. When the
```
balazsracz: nullptr -> empty

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    /// interface is operating as an ACCESS POINT only this can return the 
+    /// SSID of the ACCESS POINT operating on the WiFi interface.
+    const std::string get_ssid();
+
+    /// @return RSSI value of the currently connected access point, can return
```
balazsracz: (usually a negative number in dB)

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+        WiFiSecurity security;
+        /// Receive Signal Strength Indicator of the access point.
+        int rssi;
+    };
+
+    /// Startup the WiFi interface.
+    ///
+    /// @param mode is the requested mode of the WiFi interface.
+    /// @return zero upon successfully starting in the provided mode, non-zero
+    /// otherwise.
```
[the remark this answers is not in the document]
atanisoft: I was thinking of making it an enum but wasn't sure what values would make sense.

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+        IDLE,
+        /// The interface is currently in an initializing state. It is not
+        /// expected to be usable at this point.
+        INITIALIZING,
+        /// The interface is currently connecting to an access point.
+        CONNECTING,
+        /// Interface is not associated with an access point. This state can be
+        /// used in the event the interface has lost it's connection to the
+        /// access point.
+        NOT_ASSOCIATED,
```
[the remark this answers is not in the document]
atanisoft: Will reorder

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+        /// Receive Signal Strength Indicator of the access point.
+        int rssi;
+    };
+
+    /// Startup the WiFi interface.
+    ///
+    /// @param mode is the requested mode of the WiFi interface.
+    /// @return zero upon successfully starting in the provided mode, non-zero
+    /// otherwise.
+    int start(WiFiMode mode = WiFiMode::STATION);
```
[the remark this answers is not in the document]
atanisoft: Yes this should be a pure virtual, I'll update accordingly

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    /// Stops the WiFi interface in preparation for a reboot.
+    void stop();
+
+    /// @return true if WiFi interface has been started, false otherwise.
+    bool is_started();
+
+    /// Connect to access point.
+    ///
+    /// @param ssid access point ssid.
+    /// @param security_key access point security key.
```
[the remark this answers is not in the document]
atanisoft: Will update accordingly

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    /// the RSSI value for the access point. Note that a successful connection
+    /// is not guaranteed by this method.
+    void connect();
+
+    /// Creates an access point on this WiFi interface.
+    /// 
+    /// @param ssid access point ssid.
+    /// @param security_key access point security key.
+    /// @param security_type specifies security type.
+    /// @param ap_ip is the IP address for the access point.
```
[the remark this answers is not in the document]
atanisoft: For IP?

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    /// @param dns_ip is the DNS server IP address for the access point to
+    /// provide to stations.
+    void setup_access_point(const std::string &ssid,
+                            const std::string &security_key = "",
+                            WiFiSecurity security_type = WiFiSecurity::OPEN,
+                            uint32_t ap_ip = 0, uint32_t dns_ip = 0);
+
+    /// Initiate the WPS Push Button Control connection process.
```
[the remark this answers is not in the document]
atanisoft: Good question, I'm not certain on either as I have not used this yet. Most likely it is sync, I copied this from the CC32xx declaration and it is supported by esp32 as well

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    WiFiMode get_mode();
+
+    /// @return the current WiFi interface status.
+    WiFiState get_state();
+
+    /// Adds a saved SSID profile.
+    ///
+    /// @param ssid SSID of the profile to save.
+    /// @param sec_type @ref WiFiSecurity of the profile to be saved.
+    /// @param key password of the SSID, nullptr allowed if sec_type is OPEN.
```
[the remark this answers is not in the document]
atanisoft: Yes, forgot to update across the board with the shift to string vs char *

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    int del_ssid_profile(const int index);
+
+    /// Retrieves a saved SSID profile by index.
+    ///
+    /// @param index of the stored profile to retrieve.
+    /// @param ssid is a 33 byte array containing the ssid of the profile.
+    /// @param sec_type is the security type of the profile.
+    /// @param priority is the priority of the profile.
+    /// @return 0 upon success, else -1 on error.
+    int get_ssid_profile(const int index, const std::string &ssid,
```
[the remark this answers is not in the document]
atanisoft: Will fix for output params

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+
+    /// @return true if there are no saved SSID profiles, false otherwise.
+    bool are_ssid_profiles_empty();
+
+    /// Retrieves a list of available access points.
+    ///
+    /// @param entries will contain a list of available network entries.
+    /// @param count is the number of network entries to retrieve, max of 20.
+    /// @return number of network entries retrieved.
+    int get_network_list(NetworkListEntry *entries, size_t count);
```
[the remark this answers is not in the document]
atanisoft: I was thinking the same but modelled after CC32xx API. I'm almost thinking just return the vector instead as it can be used for iteration in that case

src/freertos_drivers/common/WiFiInterface.hxx
atanisoft: I'd expect it always to be negative values. Perhaps zero for not connected and positive for not supported/known?

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    /// otherwise.
+    int start(WiFiMode mode = WiFiMode::STATION);
+
+    /// Stops the WiFi interface in preparation for a reboot.
+    void stop();
+
+    /// @return true if WiFi interface has been started, false otherwise.
+    bool is_started();
+
+    /// Connect to access point.
```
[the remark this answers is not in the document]
atanisoft: adding a timeout parameter to make this clear that it is a synchronous (blocking) call. Though, it may be up to the implementation weather or not it is truly blocking. For code simplicity though blocking makes more sense for this API.

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,250 @@
...
+    ///
+    /// @param ssid access point ssid.
+    /// @param security_key access point security key.
+    /// @param security_type specifies security type.
+    void connect(const std::string &ssid, const std::string &security_key,
+                 WiFiSecurity security_type);
+
+    /// Connects to an access point based on stored SSID profiles.
```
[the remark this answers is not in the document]
atanisoft: sync, adding timeout param and comments to the same.

src/freertos_drivers/common/WiFiInterface.hxx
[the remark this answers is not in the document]
atanisoft: declaring byte order as NETWORK byte order (big endian)

src/freertos_drivers/common/WiFiInterface.hxx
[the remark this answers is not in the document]
atanisoft: Adding an enum which we can use for all APIs to return more meaningful error codes than zero/non-zero.

src/freertos_drivers/common/WiFiInterface.hxx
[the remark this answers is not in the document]
atanisoft: Adjusting to std::vector<NetworkEntryList> *

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,377 @@
...
+    /// Initiate the WPS Push Button Control connection process.
+    ///
+    /// @param blocking will cause this method to block until the WPS process
+    /// completes. Default is non-blocking.
+    /// @param timeout is the number of milliseconds to wait for the WPS
+    /// process to complete. A value of zero indicates that the interface
+    /// default timeout should be used.
+    ///
+    /// @return @ref WiFiErrorCode as the result of the connection attempt.
+    virtual WiFiErrorCode initiate_wps_connect(bool blocking = false,
```
bakerstu: WPS has multiple modes of operation.  For example, there is push button connect (PBC) as well as PIN mode.  This is why the CC32xxWiFi::wlan_wps_pbc_initiate() with "pbc" suffix was used.  Consider adding the "pbc" suffix.

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,377 @@
...
+    /// completes. Default is non-blocking.
+    /// @param timeout is the number of milliseconds to allow for the network
+    /// scan to complete. This is only valid when @ref blocking is true.
+    /// @param max_results is the maximum number of results to scan for. If the
+    /// WiFi interface supports a lower number than requested and in such case
+    /// the WiFi interface supported limit will be used instead.
+    ///
+    /// @return the number of networks found via scan, this value will be zero
+    /// when @ref blocking is false.
+    virtual int initiate_network_scan(bool blocking = false,
```
bakerstu: The CC32xx supports both one-shot and continuous scan.  On platforms that do not support continuous scan, it can be emulated with a periodic one-shot scan.  Should we add an additional API for continuous scan enable/disable?

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,377 @@
...
+
+        /// The WiFi interface does not have an IPv4 address available.
+        IP_UNAVAILABLE,
+
+        /// RSSI value is not known or is otherwise unavailable.
+        RSSI_UNKNOWN
+    };
+
+    /// Contains the metadata for an access point.
+    struct NetworkListEntry
```
bakerstu: Should we add bssid and channel members to this structure.  The CC32xx supports these (and more).  bssid, in case you don't know, is basically the MAC address of the ssid.

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,377 @@
...
+    /// @param priority will contain the priority of the profile.
+    ///
+    /// @return @ref WiFiErrorCode as the result of the retrieval of the
+    /// profile.
+    virtual WiFiErrorCode get_ssid_profile(const int index, std::string *ssid,
+                                           WiFiSecurity *security,
+                                           uint8_t *priority) = 0;
+
+    /// @return true if there are no saved SSID profiles, false otherwise.
+    virtual bool are_ssid_profiles_empty() = 0;
```
bakerstu: Maybe instead of a boolean API, how about a num_ssid_profiles().  It can then be used in a loop counter.  This would require some abstraction in the CC32xx because indexes can be "skipped", but it would be much more user friendly at the application layer.

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,377 @@
+/** \copyright
```
bakerstu: This is an overall comment.  I think the add_ssid_profile() and setup_access_point() forms are less elegant than ssid_profile_add() and access_point_setup.  In other words, lead with the subject, follow with the action.  We did not consistently follow this with the CC32xx, but looking back, we probably should have.

src/freertos_drivers/common/WiFiInterface.hxx
atanisoft: Yes, switching to ssid_profile_count

src/freertos_drivers/common/WiFiInterface.hxx
atanisoft: I'll add a second set of APIs for continuous.

balazsracz:
I would like to move forward with some part of the PR, essentially trying to break it into some smaller pieces. I have a couple of reasons for that:
1) smaller PRs are easier to review and allow faster progress
2) the current wifi implementation is performing extremely poorly under load conditions. (crashes between 50 and 100 packets/sec; needed performance is 800 packets/sec). However due to the significant changes pending in this PR it is not meaningful to debug its performance.

@atanisoft do you have some suggestions on how we could proceed?
For example:
- small fixes you have here and there should be able to go in
- I would be very interested in merging the socket-fd-based connection instead of the Esp32WifiClientAdapter to test its performance.

src/utils/socket_listener.cxx
```diff
@@ -102,30 +118,30 @@ void SocketListener::AcceptThreadBody() {
   ERRNOCHECK("bind",
              ::bind(listenfd, (struct sockaddr *) &addr, sizeof(addr)));
 
-#ifndef __FreeRTOS__  // no getsockname support
+#if OPENMRN_HAVE_BSD_SOCKETS_GETSOCKNAME
```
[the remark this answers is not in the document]
atanisoft: Yeah, I'm not sure who uses it, my thinking is that it was intended as a check for the bind to make sure it did bind to the right port but I don't think that is a valid use case as ERRNOCHECK should also handle that sort of scenario as well. This entire block can likely be dropped.

atanisoft:
> 2\. the current wifi implementation is performing extremely poorly under load conditions. (crashes between 50 and 100 packets/sec; needed performance is 800 packets/sec). However due to the significant changes pending in this PR it is not meaningful to debug its performance.

It is very likely something very subtle in the underlying LwIP stack. I remember you added some code to the write method in Esp32WiFiClientAdapter to increase the buffer size artificially to 512 by default. We may need to do something similar here but I'm unsure how or where that might be doable.

> 1. smaller PRs are easier to review and allow faster progress

Yes, I agree. I think this PR is at a point where it is "ready" to go in and the WiFiInterface.hxx piece can (and should) be dropped so it can be done *AFTER* this has been merged.

Should I submit a "clean" PR with the WiFiInterface.hxx pieces dropped or do we have an easy way to prune them from this PR since they are isolated to their own commits (we could go back to 5995d7defa485f22ecf9290b348ffcea1139484b and cherry-pick 0eaa84ccb4c1fbeb9f18bb34a37b00f39dca72c6 which fixes a missed htons call in the mDNS code.

balazsracz:
OK in this case I recommend you to create another branch at the same head,
and clean this branch with a rebase/cherrypick and force-push approach.
Then I'll review and merge it.
We can then work separately on the wifiinterface stuff as a follow-up.

atanisoft:
WiFiInterface.hxx dropped and will submit a new PR for that when ready.

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -137,6 +126,10 @@ string dummystring("abcdef");
 /// layout. The argument of offset zero is ignored and will be removed later.
 static constexpr openlcb::ConfigDef cfg(0);
 
+#if defined(USE_WIFI)
+Esp32WiFiManager wifiMgr(ssid, password, openmrn.stack(), cfg.seg().wifi());
```
balazsracz: what is the desired style for global variables for you?
After this change there will be both lower_case_underscore and camelCase in the same file.

arduino/examples/ESP32IOBoard/config.h
```diff
@@ -1,10 +1,12 @@
 #ifndef _ARDUINO_EXAMPLE_ESP32IOBOARD_CONFIG_H_
 #define _ARDUINO_EXAMPLE_ESP32IOBOARD_CONFIG_H_
 
-#include "openlcb/ConfiguredConsumer.hxx"
-#include "openlcb/ConfiguredProducer.hxx"
-#include "openlcb/ConfigRepresentation.hxx"
-#include "openlcb/MemoryConfig.hxx"
+#include <openlcb/ConfiguredConsumer.hxx>
```
balazsracz: I usually use <...> includes for system headers and "..." for headers in our own codebase.

arduino/examples/ESP32WifiCanBridge/ESP32WifiCanBridge.ino
```diff
@@ -108,6 +104,8 @@ string dummystring("abcdef");
 // layout. The argument of offset zero is ignored and will be removed later.
 static constexpr openlcb::ConfigDef cfg(0);
 
+Esp32WiFiManager wifiMgr(ssid, password, openmrn.stack(), cfg.seg().wifi());
```
balazsracz: see above

include/openmrn_features.h
```diff
@@ -101,6 +103,39 @@
...
+/// when listening for incoming connections.
+#define OPENMRN_HAVE_BSD_SOCKETS_GETSOCKNAME 1
+#endif
+
+#if defined(__linux__)
+/// Compiles suport for IPv6 address types
+#define OPENMRN_HAVE_BSD_SOCKETS_IPV6 1
+#endif
+
+#if defined(__linux__)
```
balazsracz: also add `__MACH__` here

include/openmrn_features.h
```diff
@@ -101,6 +103,39 @@
...
+/// Compiles suport for ignoring SIGPIPE write failures on sockets.
```
balazsracz: /// Ignores SIGPIPE signals to avoid write failures crashing the program.

include/openmrn_features.h
```diff
@@ -101,6 +103,39 @@
...
+#define OPENMRN_BSD_SOCKETS_IGNORE_SIGPIPE 1
+#endif
+
+#if defined(__linux__) || defined(ESP32)
```
balazsracz: `__MACH__` ?

include/openmrn_features.h
```diff
@@ -101,6 +103,39 @@
...
+#endif
+
+#if defined(__linux__)
+/// Compiles suport for IPv6 address types
+#define OPENMRN_HAVE_BSD_SOCKETS_IPV6 1
+#endif
+
+#if defined(__linux__)
+/// Compiles suport for ignoring SIGPIPE write failures on sockets.
+#define OPENMRN_BSD_SOCKETS_IGNORE_SIGPIPE 1
```
balazsracz: `_FEATURE_`

include/openmrn_features.h
```diff
@@ -101,6 +103,39 @@
...
+#endif
+
+#if defined(__linux__) || defined(ESP32)
+/// Compiles support for reporting EOF as an error for read/write.
+#define OPENMRN_BSD_SOCKETS_REPORT_EOF_ERROR 1
```
balazsracz: `_FEATURE_`

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
```diff
@@ -0,0 +1,139 @@
...
+    Min(0), Max(1), Default(0),
+    MapValues(Esp32WiFiConfigurationParams::BOOLEAN_MAP));
+/// Specifies the port which should be used by the hub.
+CDI_GROUP_ENTRY(port, openlcb::Uint16ConfigEntry,
+    Name(Esp32WiFiConfigurationParams::HUB_LISTENER_PORT_NAME),
+    Description(Esp32WiFiConfigurationParams::HUB_LISTENER_PORT_DESC),
+    Min(1), Max(65535),
+    Default(openlcb::TcpClientDefaultParams::DEFAULT_PORT))
+/// Specifies the mDNS service name to advertise for the hub.
+CDI_GROUP_ENTRY(service_name, openlcb::StringConfigEntry<48>,
```
balazsracz: Consider adding a few bytes (4-6) of reserved space before this entry. Will help to avoid needing to reset settings in case of a later update.

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
```diff
@@ -0,0 +1,139 @@
...
+    Name(Esp32WiFiConfigurationParams::WIFI_POWER_SAVE_NAME),
+    Description(Esp32WiFiConfigurationParams::WIFI_POWER_SAVE_DESC),
+    Min(0), Max(1), Default(0),
+    MapValues(Esp32WiFiConfigurationParams::BOOLEAN_MAP));
+/// CDI Configuration to enable this node to be a hub.
+CDI_GROUP_ENTRY(hub, HubConfiguration,
+    Name(Esp32WiFiConfigurationParams::HUB_NAME),
+    Description(Esp32WiFiConfigurationParams::HUB_DESC));
+/// CDI Configuration for this node's connection to an uplink hub.
+CDI_GROUP_ENTRY(uplink,
```
balazsracz: Is there an enable/disable setting for the uplink?
Or should there be a single 3/4-way setting for wifi instead:
- disabled
- client
- server
- client + server

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
```diff
@@ -0,0 +1,139 @@
...
+    openlcb::TcpClientConfig<openlcb::TcpClientDefaultParams>,
+    Name(Esp32WiFiConfigurationParams::UPLINK_NAME),
+    Description(Esp32WiFiConfigurationParams::UPLINK_DESC));
+CDI_GROUP_END();
+
+} // namespace openmrn_arduino
+
+using openmrn_arduino::WiFiConfiguration;
```
balazsracz: I don't think this is necessary.
We should only use using in .hxx files for stuff that the end-user will have to write in their sketch in a place where a namespace designation would be surprising or ugly.

src/utils/HubDevice.hxx
```diff
@@ -47,10 +48,18 @@ template <class Data> class FdHubWriteFlow;
 class FdHubPortBase : public FdHubPortInterface, private Atomic
 {
 public:
+#ifdef ESP32 // TODO: shrink these if possible
```
balazsracz: If we need these ifdefs then maybe we should make these from constants into DECLARE_CONST type parameters. You can file an issue against me and I can migrate them.

We already have a parameter called main_thread_stack_size or similar.

src/utils/socket_listener.cxx
```diff
@@ -57,18 +63,28 @@ static void* accept_thread_start(void* arg) {
   return NULL;
 }
 
+#ifdef ESP32
+/// Stack size to use for the accept_thread_.
+static constexpr size_t listener_stack_size = 1536;
```
balazsracz: same as above

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -84,24 +75,22 @@ constexpr uint16_t OPENMRN_TCP_PORT = 12021L;
...
 //     #include <OpenMRNLite.h>
 //
-//     const char DEFAULT_WIFI_NAME[] = "linksys";
-//     const char DEFAULT_WIFI_PASSWORD[] = "theTRUEsupers3cr3t";
+//     char WIFI_SSID[] = "linksys";
+//     char WIFI_PASS[] = "theTRUEsupers3cr3t";
 
 /// This is the name of the WiFi network (access point) to connect to.
-const char *ssid = DEFAULT_WIFI_NAME;
+const char *ssid = WIFI_SSID;
```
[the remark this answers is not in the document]
atanisoft: Both WIFI_SSID and WIFI_PASS are defined in freertos_drivers/common/WiFiDefs.hxx/cxx which are added to libify.sh to put them under the arduino tree like other files were. The weak declaration is already in place in this file so there should be no problems (and haven't been from my testing).

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
atanisoft: Either is fine, I'll update the code to match existing standards.

arduino/examples/ESP32IOBoard/config.h
atanisoft: Even in applications? From an application standpoint this would be considered a "system" include since it lives outside the application code base.

include/openmrn_features.h
atanisoft: I think it makes sense but I don't have a MacOS available to test with. I'll add it here and we can remove it if we find it causes issues.

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
atanisoft: I copied the size from TcpAutoAddress / service_name, should we increase this in all places to a uniform size?

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
atanisoft: will remove, I wasn't sure if it was needed and erred on the side of caution.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+
+    // Configure the SSID details for the station based on the SSID and
+    // password provided to the Esp32WiFiManager constructor.
+    wifi_config_t conf;
+    memset(&conf, 0, sizeof(wifi_config_t));
+    strcpy(reinterpret_cast<char*>(conf.sta.ssid), ssid_);
+    // if we have a password check if it is a PSK or regular password
+    if(password_)
+    {
+        if (strlen(password_) == 64) // it's the PSK and not a password.
```
balazsracz: What is the difference between the two branches? They both seem to copy until the first null.
Seems like this could be replaced with a single strncpy

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
atanisoft: Right now there are two types of nodes configurable with this CDI:
1) hub + uplink
2) uplink only

I think it would make sense to add the third option of:
3) hub only

Right now I'm just exposing uplink as TcpClientConfig, would it make sense to add an "enable" option to that config or should I add "uplink_enabled" here? Or should there be a "mode" entry that uses a map (uplink only, hub only, hub+uplink)?

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+        if(bits & DHCP_GOTIP_BIT)
+        {
+            break;
+        }
+    }
+
+    // Check if we successfully connected or not. If not, force a reboot.
+    if((bits & WIFI_CONNECTED_BIT) != WIFI_CONNECTED_BIT)
+    {
+        LOG(FATAL, "[WiFi] Failed to connect to SSID.");
```
balazsracz: please also print the actual ssid we were trying to connect to

src/utils/HubDevice.hxx
atanisoft: https://github.com/bakerstu/openmrn/issues/255

src/utils/socket_listener.cxx
atanisoft: https://github.com/bakerstu/openmrn/issues/256

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+void Esp32WiFiManager::start_wifi_task()
+{
+    LOG(INFO, "[WiFiMgr] Starting WiFi Manager task");
+    os_thread_create(&wifiTaskHandle_, "OpenMRN-WiFiMgr", WIFI_TASK_PRIORITY,
+        WIFI_TASK_STACK_SIZE, wifi_manager_task, this);
+}
+
+// Background task for the Esp32WiFiManager. This handles all outbound
+// connection attempts, configuration loading and making this node as a hub.
+void *Esp32WiFiManager::wifi_manager_task(void *param)
```
balazsracz: this code is fine, it's idiomatic FreeRTOS code.

However, we are typically not writing low-level freertos code in OpenMRN, as it is not generally portable. Whenever possible we should try to use the abstractions in OS.hxx and Executor.
We have OSThread for infinite loops, StateFlow for sleeps-most-of-the-time use-cases, Notifiable and OSSem for wake-me-up-sometime callbacks, etc.

With moving towards WifiInterface it would be nice to make this code portable, because ultimately there is nothing freertos specific in here. In other words, all of this code should be able to run on a Raspberry Pi Zero W, or an ESP8266 NONOS, provided that an appropriate WifiInterface implementation is created.

This is clearly not a blocking comment, because we decided to take the WifiInterface refactoring to a separate effort.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+
+        // Sleep until we are woken up again for configuration update or WiFi
+        // event.
+        ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
+    }
+
+    return nullptr;
+}
+
+// Shuts down the hub listener (if enabled and running) for this node.
```
balazsracz: ///
consider copying this comment to the class header.
Please document that this is a blocking call.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+    hubServiceName_ = cfg_.hub().service_name().read(configFd_);
+    uint16_t hub_port = CDI_READ_TRIMMED(cfg_.hub().port, configFd_);
+
+    LOG(INFO, "[HUB] Starting TCP/IP listener on port %d", hub_port);
+    hub_.reset(new GcTcpHub(stack_->can_hub(), hub_port));
+
+    // wait for the hub to complete it's startup tasks
+    while(!hub_->is_started())
+    {
+        vTaskDelay(HUB_STARTUP_DELAY);
```
balazsracz: this should be usleep instead (eventually to migrate to nanosleep)

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+    if(hub_)
+    {
+        mdns_unpublish(hubServiceName_.c_str());
+        LOG(INFO, "[HUB] Shutting down TCP/IP listener");
+        //hub_->shutdown();
+        hub_.reset(nullptr);
+    }
+}
+
+// Creates a hub listener for this node after loading configuration details.
```
balazsracz: see above

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+
+/// Maximum number of milliseconds to wait for mDNS query responses.
+static constexpr uint32_t MDNS_QUERY_TIMEOUT = 2000;
+
+/// Maximum number of results to capture for mDNS query requests.
+static constexpr size_t MDNS_MAX_RESULTS = 10;
+
+// macro for splitting a service name since the ESPmDNS library requires the
+// service name and service protocol to be split.
+#define SPLIT_MDNS_SERVICE_NAME(service_name, service_protocol) \
```
balazsracz: why is this a macro instead of a function?

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+    { \
+        string::size_type split_loc = service_name.find('.', 0); \
+        service_protocol = service_name.substr(split_loc + 1); \
+        service_name.resize(split_loc); \
+    }
+
+// Advertises an mDNS service name for this node.
+void mdns_publish(const char *name, const char *service, uint16_t port)
+{
+    string service_name = service, protocol_name;
```
balazsracz: we don't usually declare more than one variable on a single line.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+int getifaddrs(struct ifaddrs **ifap)
+{
+    tcpip_adapter_ip_info_t ip_info;
+
+    /* start with something "safe" in case we bail out early */
+    *ifap = nullptr;
+
+    if(!tcpip_adapter_is_netif_up(TCPIP_ADAPTER_IF_STA))
+    {
+        // Station TCP/IP interface is not up
```
balazsracz: what about 127.0.0.1, should that be included in the interfaces?

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+{
+    while (ifa)
+    {
+        struct ifaddrs *next = ifa->ifa_next;
+
+        HASSERT(ifa->ifa_data == nullptr);
+        HASSERT(ifa->ifa_ifu.ifu_broadaddr == nullptr);
+        HASSERT(ifa->ifa_netmask == nullptr);
+
+        if (ifa->ifa_addr)
```
balazsracz: delete nullptr is fine.

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
balazsracz: the point is not the byte size of service name, but to have a few reserved bytes in the parent structure so we can add more options later.

arduino/examples/ESP32IOBoard/config.h
balazsracz: applications we have in the openmrn tree are using " "
Is there a specific style that is common to the arduino people? Let's exclude the trivial cases where library includes are triggered.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: oops, I missed adding that, let me add those before this merges.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: I'll move it to a function, it originally was a macro for inline impl but with it being used in a few places it makes more sense as a function.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: will split.

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
atanisoft: ahh, I'll add a hidden block of bytes so we can inject them if needed in the future.

src/freertos_drivers/esp32/Esp32WiFiManager.hxx
```diff
@@ -0,0 +1,208 @@
...
+    /// configuration for the first time, otherwise it is an update to the
+    /// configuration and may require a restart.
+    /// @param done is the control used by the caller to track when all config
+    /// consumers have completed their updates.
+    ///
+    /// @return UPDATED when the configuration has been successfully updated,
+    /// or REBOOT_NEEDED if the node needs to reboot for configuration to take
+    /// effect.
+    ConfigUpdateListener::UpdateAction apply_configuration(int fd,
+        bool initial_load, BarrierNotifiable *done) OVERRIDE;
```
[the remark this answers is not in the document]
atanisoft: ok, I'll scrub the new code for that and fix it up.

src/freertos_drivers/esp32/Esp32WiFiManager.hxx
```diff
@@ -0,0 +1,208 @@
...
+    /// can react to WiFi events to cleanup or recreate the hub or uplink
+    /// connections as required.
+    ///
+    /// @param event_id is the system_event_t.event_id value.
+    void process_wifi_event(int event_id);
+private:
+    /// Default constructor.
+    Esp32WiFiManager();
+
+    /// Starts the WiFi system and initiates the SSID connection process.
```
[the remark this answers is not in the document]
atanisoft: Yes, that is also true, I'll add it and document the time limit.

src/freertos_drivers/esp32/Esp32WiFiManager.hxx
```diff
@@ -0,0 +1,208 @@
...
+    void stop_hub();
+    void start_hub();
+    void stop_uplink();
+    void start_uplink();
+
+    void on_uplink_created(int fd, Notifiable *on_exit);
+
+    /// Handle for the wifi_manager_task that manages the WiFi stack, including
+    /// periodic health checks of the connected hubs or clients.
+    TaskHandle_t wifiTaskHandle_;
```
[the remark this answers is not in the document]
atanisoft: fixed

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+/// exposed in the MDNS class but is supported on the ESP32.
+void mdns_unpublish(const char *service);
+
+// End of global namespace block.
+
+namespace openmrn_arduino {
+
+/// Priority to use for the wifi_manager_task. This is currently set to
+/// the same priority level as the arduino-esp32 loopTask. The task will
+/// be in a sleep state until woken up by Esp32WiFiManager::process_wifi_event
```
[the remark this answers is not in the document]
atanisoft: fixed

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+    // underscore.
+    hostname_.reserve(TCPIP_HOSTNAME_MAX_SIZE);
+
+    // Generate the hostname for the ESP32 based on the provided node id.
+    // node_id : 0x050101011425
+    // hostname_ : esp32_050101011425
+    NodeID node_id = stack_->node()->node_id();
+    hostname_.append(uint64_to_string_hex(node_id, 0));
+
+    // Release any extra capacity allocated for the hostname.
```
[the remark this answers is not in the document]
atanisoft: fixed

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: you're right, it doesn't make sense as written. The code was copied as-is from the arduino code and doesn't really make sense there either. will consolidate to just the else case instead.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: fixed

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: Yes, it does make sense to move this to a more generic approach using OSThread/etc. We just need a way to wake up the task when an "interesting" event occurs so it can take care of cleanup or start running, etc.

With the WiFiInterface code I think a common implementation can be achieved between CC32xx, ESP, RPi, ESP32 with a minimal implementation layer for the hardware drivers etc.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: will switch and update HUB_STARTUP_DELAY to use MSEC_TO_USEC instead.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: originally it was declared as a macro for inline purposes but it no longer makes sense as such so it will move to a function.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: will fix

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: copied from CC32xx, will drop condition checks since they will always be defined.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: I'm not certain that the esp32 even has an equivalent to localhost. From the tcpadapter code there is only STA, AP and ETH interfaces. I don't should support ETH in this code, I'm also uncertain what components would be needed to support that if we decided to via a more generic interface (NetworkInterface, WiFiInterface extending NetworkingInterface, Esp32WiFi extending WiFiInterface, EspEthernet extending NetworkInterface, etc)

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,985 @@
...
+private:
+    const int configFd_;
+    const TcpClientConfig<TcpClientDefaultParams> cfg_;
+};
+
+// With this constructor being used the Esp32WiFiManager will manage the
+// WiFi connection, mDNS system and the hostname of the ESP32.
+Esp32WiFiManager::Esp32WiFiManager(const char *ssid, const char *password,
+    SimpleCanStack *stack, const WiFiConfiguration &cfg) :
+    DefaultConfigUpdateListener(), ssid_(ssid), password_(password),
```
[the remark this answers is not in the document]
atanisoft: will have it reformatted on next commit.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: fixed

arduino/examples/ESP32IOBoard/config.h
atanisoft: Most libraries and examples I've seen use < > with only a few exceptions. This may be one of those that could go either way really.

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
atanisoft: Turns out this is required after all. With the updates being done for the remaining comments below this wouldn't compile without the using declaration as it is used in the config.h.

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
atanisoft: What do you envision the usecase for "disabled" to be?

arduino/examples/ESP32IOBoard/config.h
balazsracz: I would prefer keeping to quotes.

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
balazsracz: CAN-bus only.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -0,0 +1,993 @@
...
+static constexpr UBaseType_t WIFI_TASK_PRIORITY = 2;
+
+/// Stack size for the wifi_manager_task.
+static constexpr uint32_t WIFI_TASK_STACK_SIZE = 2560L;
+
+/// Interval at which to check the WiFi connection status.
+static constexpr TickType_t WIFI_CONNECT_CHECK_INTERVAL = pdMS_TO_TICKS(5000);
+
+/// Interval at which to check if the GcTcpHub has started or not.
+static constexpr TickType_t HUB_STARTUP_DELAY = MSEC_TO_USEC(50);
```
balazsracz: uint32_t instead of ticktype

src/freertos_drivers/esp32/Esp32WiFiConfiguration.hxx
atanisoft: Ahh, that does make sense. A can node that might make use of TCP/IP. I don't think that will be a common use case but it can be supported. Let me move hub enabled up to the top level as an enum like entry containing: uplink enabled, hub enabled, uplink and hub enabled, disabled. That should cover the options above and when we move to the WiFiInterface model we could possibly add a third mode of operation for this code to use a wifi profile instead of passing SSID/password.

arduino/examples/ESP32IOBoard/config.h
atanisoft: Ok, will update.

I believe this is one of those sort of issues that falls into compiler behavior. " Usually means look in same directory or local include path and < > means system include path. Or something along those lines.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
balazsracz: i've already fixed it

PR #252, opened 2019-03-09
Javascript client upgrades

balazsracz:
- Updates js_clent application to be module compatible
- Adds encapsulation class for Node ID
- Adds PIP client
- Adds NodeBrowser client
- Adds Bootloader Client
- Adds progress callback to Bootloader Client
- Adds an example node.js script that upgrades all nodes that show up in firmware upgrade mode.
- moves connection establishment (server/client for TCP/websocket) to separate calls.


TCSDCC-Jordan:
This looks great. I read through the changes and this looks like a totally doable project here. Thanks I will keep you posted as pieces come together. I am holding off merging the pull (not sure if I have the authority to make that call), but it all looks good.

PR #254, opened 2019-03-12
Loadtest for esp32

balazsracz:
- Adds documentation .md file for loadtesting why and how
- Adapts the CPuLoad component to work on the ESP32 including multi-core freertos
- Adds an arduino example sketch with CpuLoad enabled
- Adds a bit more logging into cpuload, specifically for free heap and buffer pool size


applications/load_test/README.md
```diff
@@ -0,0 +1,164 @@
...
+- CPU usage by individual functions (hotspot detection)
+- jitter of available CPU -- maximum burst length where no CPU is available.
+
+## CPU load detection for FreeRTOS
+
+### Overview
+
+There is a shared driver for computing CPU load under FreeRTOS using a sampling
+technique. The sampler shall be a hardware timer resource that ticks
+independently of the FreeRTOS system tick, and upon each tick we sample whatthe
```
atanisoft: missing space at end of line.

arduino/OpenMRNLite.h
```diff
@@ -362,6 +362,11 @@ class OpenMRN : private Executable
         haveExecutorThread_ = true;
         stack_->executor()->start_thread(
             "OpenMRN", OPENMRN_TASK_PRIORITY, OPENMRN_STACK_SIZE);
+#ifdef ESP32
+        // Remove IDLE0 task watchdog, because the openmrn task sometimes gets
+        // scheduled on CPU0 and can use 100% cpu.
+        esp_task_wdt_delete(xTaskGetIdleTaskHandleForCPU(0));
```
atanisoft: switch to ```disableCore0WDT();``` which is the API from arduino-esp32 that is recommended for this sort of usage. we can alternatively do this for CORE 1 if the executor ends up there instead.

arduino/examples/ESP32CanLoadTest/ESP32CanLoadTest.ino
```diff
@@ -0,0 +1,401 @@
...
+    extern const size_t CONFIG_FILE_SIZE = cfg.seg().size() + cfg.seg().offset();
+
+    // Default to store the dynamic SNIP data is stored in the same persistant
+    // data file as general configuration data.
+    extern const char *const SNIP_DYNAMIC_FILENAME = CONFIG_FILENAME;
+}
+
+CpuLoad cpu_load;
+hw_timer_t * timer = nullptr;
+CpuLoadLog* cpu_log = nullptr;
```
atanisoft: we should also evaluate using a pair of GPIO pins for LEDs to rule out the serial IO as an influencer in the timing. The serial IO is blocking in a few scenarios (ie: rapidly sending serial data), it is not clear that this will influence it but it would be good to rule that out as well.

arduino/examples/ESP32CanLoadTest/ESP32CanLoadTest.ino
```diff
@@ -0,0 +1,401 @@
...
+  }
+}
+
+void setup()
+{
+    Serial.begin(115200L);
+
+    timer = timerBegin(3, 80, true); // timer_id = 3; divider=80; countUp = true;
+    timerAttachInterrupt(timer, &onTimer, true); // edge = true
+    timerAlarmWrite(timer, 1000000/163, true);  //1000 ms
```
atanisoft: comment does not reflect intention here. We can increase the divider from 80 to 80000 and replace this math with 1000/163 (or ~6).

arduino/OpenMRNLite.h
balazsracz: there is no WDT on core 1 by default in arduino (since the loop task is there the idle is never run by default)

arduino/examples/ESP32CanLoadTest/ESP32CanLoadTest.ino
balazsracz: We only print one line per two seconds.
There exists a component for displaying via two LEDs (targeted at RG leds: https://github.com/bakerstu/openmrn/blob/master/src/utils/CpuDisplay.hxx ) but for loadtests it is too little information.

arduino/examples/ESP32CanLoadTest/ESP32CanLoadTest.ino
balazsracz: fixed.
1000/6 is not good because that will run the timer in sync with the kernel ticks. We need the two timers to be running at a relative prime interval to avoid systematic error in the sampling.

arduino/examples/ESP32CanLoadTest/ESP32CanLoadTest.ino
atanisoft: Good point, the Serial calls may block while writing to internal buffers but at a once every 2s shouldn't influence it too significantly I'd imagine.

arduino/OpenMRNLite.h
atanisoft: WDT can be enabled on core 1 by the user. You are correct though that it is off by default. The idle task will still run regardless though for general housekeeping tasks but it should be minimal impact.

arduino/examples/ESP32CanLoadTest/ESP32CanLoadTest.ino
atanisoft: I wasn't meaning 1000/6 but instead 6uS by itself which is close to the 1M/163 that is used now. The updated commit though clarifies the intent enough though.

PR #260, opened 2019-03-19
[WIP] Esp32 Wifi event handling improvements

atanisoft:
Originally the Esp32WiFiManager would only react to the SYSTEM_EVENT_STA_DISCONNECTED if we had previously connected. This unfortunately could result in the esp32 never connecting to the SSID due to timeouts which would not be logged. The updated code now processes the event regardless of connected state and will trigger a reconnect unless the disconnect reason is SSID not found which will be logged and the esp32 restarted (via FATAL log level handling)


balazsracz:
There is a case when this code shuts down the wifi stack and never starts it again.
It typically happens to me when i connect successfully, have IP, then later lose the connection to the AP.

```
I (292338) wifi: reason: assoc expire(4)
I (292341) wifi: Send disconnect event, reason=4, AP number=0
[WiFi] Connection to SSID "opia24" failed, reason: 4 (ASSOC_EXPIRE)
[WiFi] Shutting down WiFi stack...
I (292356) wifi: connect step 4 -> 0, status=1, err_time=1
I (292361) wifi: Start wifi disconnect
I (292365) wifi: connect step 0 -> 0, status=1, err_time=0
I (292370) wifi: connect status 1 -> 0, step=0, err_time=0
I (292375) wifi: filter: set rx policy=8
I (292379) wifi: filter: set rx policy=2
I (292382) wifi: clear blacklist
I (292385) wifi: clear rc list
I (292388) wifi: flush txq
I (292390) wifi: stop sw txq
I (292393) wifi: filter: set rx policy=0
I (292397) wifi: lmac stop hw txq
I (292400) wifi: Deinit lldesc rx mblock:4
Ex 729|FreeHeap 254068|Buf 568|Load: avg  61 max streak 31 max of 16 16
 | 49.6:loopTask | 38.5:IDLE0 | 8.7:wifi | 2.1:OpenMRN | 0.4:ex 0x3f403d28 | 0.3:eventTask | 0.1:ESP32-CAN TX
Ex 740|FreeHeap 254064|Buf 568|Load: avg  53 max streak 3 max of 16 12
 | 49.6:loopTask | 47.5:IDLE0 | 2.1:OpenMRN | 0.3:ex 0x3f403d28 | 0.1:ex 0x3f403bb4 | 0.1:ex 0x3f403d98
Ex 740|FreeHeap 254064|Buf 568|Load: avg  52 max streak 4 max of 16 11
 | 49.7:loopTask | 47.3:IDLE0 | 2.3:OpenMRN | 0.3:ex 0x3f403d28 | 0.1:ex 0x3f403d98 | 0.1:esp_timer
Ex 740|FreeHeap 254064|Buf 568|Load: avg  52 max streak 3 max of 16 10
 | 49.6:loopTask | 47.5:IDLE0 | 2.0:OpenMRN | 0.4:ex 0x3f403d28 | 0.1:tiT | 0.1:ex 0x3f403d98
Ex 740|FreeHeap 254064|Buf 568|Load: avg  52 max streak 5 max of 16 10
 | 49.8:loopTask | 47.2:IDLE0 | 2.6:OpenMRN | 0.3:ex 0x3f403d28
```

This suggests that something is wrong in the communication between the callbacks and the custom freertos thread of the wifi manager.

atanisoft:
Hmm, I haven't seen that before. I'm not sure the deinit/init cycle is meaningful. More likely if there was a previous connection we just need to call reconnect instead of deinit etc (especially since the check code is only in the init method).

I'll add this to the to-do list for cleanup. I've still not reached a stable point for the espduino32, two diff APs show the two failures you reported previously (timeout related)

balazsracz:
> Hmm, I haven't seen that before.

I think an easy repro is the following: Turn on ESP32, let it connect to wifi. Then turn off the AP. You should get this state.

> I'm not sure the deinit/init cycle is meaningful. More likely if there was a previous connection we just need to call reconnect instead of deinit etc (especially since the check code is only in the init method).

Based on my most recent tests I would be onboard to go with the simple handling (doing a call to connect from the disconnect handler) and drop the reinit code.

> 
> I'll add this to the to-do list for cleanup. I've still not reached a stable point for the espduino32, two diff APs show the two failures you reported previously (timeout related)

which failures are those? the AUTH_EXPIRE? which is the other?
I think these could be related to bad antenna design or even a manufacturing problem (a bad solder joint for example).

QQ: do you know if there is an API to change transmit power on the esp32?

atanisoft:
> I think an easy repro is the following: Turn on ESP32, let it connect to wifi. Then turn off the AP. You should get this state.

That would make sense since the task doesn't know how to reinit the stack. I think your simpler approach to just call connect again is perfect to fix this. During the init cycle we can improve upon it with the shutdown cycle if we find it needed but I haven't found it to be very helpful either way in avoiding some issues.

The only thing I'd like to add to the disconnected event handler is a log entry sharing the reason for disconnect which I'll raise as a new PR.

> which failures are those? the AUTH_EXPIRE? which is the other?

That was the primary one, the second was 4way handshake timeout. Both of these point to poor connection quality and/or poor manufacturing as you suggest. Mostly in antenna design, this is one area where the IPEX connector is helpful.

TX power API, there is one API for that: https://github.com/espressif/esp-idf/blob/master/components/esp_wifi/include/esp_wifi.h#L917

From what I've seen though, the default is max tx power output. I haven't seen many examples of it's use though so it may be worth exploring further still.

balazsracz:
> > I think an easy repro is the following: Turn on ESP32, let it connect to wifi. Then turn off the AP. You should get this state.
> 
> That would make sense since the task doesn't know how to reinit the stack. I think your simpler approach to just call connect again is perfect to fix this. During the init cycle we can improve upon it with the shutdown cycle if we find it needed but I haven't found it to be very helpful either way in avoiding some issues.
> 
> The only thing I'd like to add to the disconnected event handler is a log entry sharing the reason for disconnect which I'll raise as a new PR.
> 
> > which failures are those? the AUTH_EXPIRE? which is the other?
> 
> That was the primary one, the second was 4way handshake timeout. Both of these point to poor connection quality 

I have verified this (on the AUTH_EXPIRE) with a wifi sniffer yesterday. Sent you a chat on hangouts.

> and/or poor manufacturing as you suggest. Mostly in antenna design, this is one area where the IPEX connector is helpful.
> 
> TX power API, there is one API for that: https://github.com/espressif/esp-idf/blob/master/components/esp_wifi/include/esp_wifi.h#L917
> 
> From what I've seen though, the default is max tx power output. I haven't seen many examples of it's use though so it may be worth exploring further still.

atanisoft:
closing for rework after recent updates (mainly will pull in log updates)

PR #261, opened 2019-03-19
esp32 wifi connect hack

balazsracz:
this is the hack i tried to make the esp32 connect.


atanisoft:
I've merged in the reinit pieces you added here to https://github.com/bakerstu/openmrn/pull/260

balazsracz:
this is now obsolete

PR #265, opened 2019-04-03
Fixes missing HubDeviceSelect.hxx/cxx in the arduino compilation.

balazsracz:


atanisoft:
@balazsracz
Found the references:
https://github.com/bakerstu/openmrn/blob/master/src/utils/GridConnectHub.cxx#L44-L46
https://github.com/bakerstu/openmrn/blob/master/src/utils/GridConnectHub.cxx#L407-L411

We should fix these since this works for the ESP32, this is the other missing piece that got lost somewhere in the shuffle of the esp32 wifi mgr code.

balazsracz:
> @balazsracz
> Found the references:
> https://github.com/bakerstu/openmrn/blob/master/src/utils/GridConnectHub.cxx#L44-L46
> https://github.com/bakerstu/openmrn/blob/master/src/utils/GridConnectHub.cxx#L407-L411
> 
> We should fix these since this works for the ESP32, this is the other missing piece that got lost somewhere in the shuffle of the esp32 wifi mgr code.

I will make a separate PR to get those back, because it will need a feature macro.

PR #267, opened 2019-04-04
Reenable GcHub select feature for esp32.

balazsracz:
Updates ifdef guard around include of HubDeviceSelect to an openmrn_feature.h tracked symbol.
Adjusts this symbol to enable esp32 to have GcTcpHub use select feature enabled.

### DIFF
--- a/src/utils/GridConnectHub.cxx
+++ b/src/utils/GridConnectHub.cxx
@@ -404,7 +404,7 @@ struct GcHubPort : public Executable
     {
         LOG(VERBOSE, "gchub port %p", (Executable *)this);
         if (use_select) {
-#ifdef OPENMRN_FEATURE_EXECUTOR_SELECT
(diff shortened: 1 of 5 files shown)


src/utils/GridConnectHub.cxx
```diff
@@ -404,8 +404,8 @@ struct GcHubPort : public Executable
     {
         LOG(VERBOSE, "gchub port %p", (Executable *)this);
         if (use_select) {
-#ifdef ARDUINO
-            DIE("select is not supported on Arduino");
+#ifdef OPENMRN_FEATURE_EXECUTOR_SELECT
```
atanisoft: Shouldn't this be ifndef?

balazsracz: Good catch, thanks for looking carefully!

PR #270, opened 2019-04-08
Improves performance of RefreshLoop object

balazsracz:
Improves performance of RefreshLoop object
by shortcutting the yield to the Executor in case the polling
target has called the Notifiable inline. In this case there
is nothing to wait for, so we can iterate through the polling
members faster.

This avoids a problem where the CPU cost of executor yields can
exceed the amount of time between the polling loops, causing 100%
CPU usage just from the RefreshLoop. This problem could arise when
there are many polling members.

### DIFF
--- a/src/openlcb/RefreshLoop.hxx
+++ b/src/openlcb/RefreshLoop.hxx
@@ -64,7 +64,6 @@ class RefreshLoop : public StateFlowBase
 public:
     RefreshLoop(Node *node, const std::initializer_list<Polling *> &members)
         : StateFlowBase(node->iface())
-//        , node_(node)
         , timer_(this)
         , lastTimeout_(os_get_time_monotonic())
         , members_(members)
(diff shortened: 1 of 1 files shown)


atanisoft:
Based on the code in this PR it looks ok but I'm not 100% certain what is being fixed with it. Do you have a sample showing the failure?

balazsracz:
I repeatedly encountered a scenario during the loadtesting, when OpenMRN thread consumed 100% of the CPU, and the only actual flow that was running was the RefreshLoop. I believe the root cause of this problem was that there were many polling objects, and after K (=5) calls the Executor goes into and out of select(). Since the esp32 has a very expensive select implementation, we were paying a lot of costs just for the polling loop.
This is one of the pending fixes for that problem.
Another pending fix is that I'm adjusting the value of K for the esp32.

atanisoft:
Ok with that explanation it makes a lot more sense for this update.

Likely we can reproduce this in other envs with similar "broken" select implementations when we have enough pollers (esp32 examples have 8 today).

PR #271, opened 2019-04-08
Limits buffer memory between GC bridge and write flow.

balazsracz:
Adds LimitedPool, an implementation of a BufferPool that allows only a fixed number of buffers to be allocated.

After this is exhausted, further allocations get blocked until some earlier buffer gets freed. This is similar to FixedPool except the memory is allocated and returned to mainBufferPool, so the overhead on a per-instance basis is much lower.

Adds LimitedPool to the BufferPort component which is responsible for sending gridconnect packets to the Hub's write flow. A link-time configuration option defines whether we should limit the number of filled-in packets sent to the write flow.

Without this limitation if the write flow gets backed up, there is an unlimited amount of RAM we use up in the queue of Buffers going out of the BufferPort and waiting on the input of the WriteFlow. With this limit the batching size will automatically adjust to the throughput that the WriteFlow can sustain.

### DIFF
--- a/src/utils/BufferPort.hxx
+++ b/src/utils/BufferPort.hxx
@@ -36,6 +36,8 @@
 #ifndef _UTILS_BUFFERPORT_HXX_
 #define _UTILS_BUFFERPORT_HXX_
 
+#include "utils/LimitedPool.hxx"
+
 /// A wrapper class around a string-based Hub Port that buffersthe outgoing
 /// bytes for a specified delay timer before sending the data off. This helps
 /// accumulate more data per TCP packet and increase transmission efficiency.
@@ -84,6 +86,12 @@ public:
 private:
     Action entry() override
     {
+        if (!tgtBuf_) {
+            return allocate_and_call(downstream_, STATE(buf_alloc_done),
+                config_gridconnect_bridge_max_outgoing_packets() <= 1
+                    ? nullptr
+                    : &outputPool_);
+        }
         if (msg().size() < (bufSize_ - bufEnd_))
         {
             // Fits into the buffer.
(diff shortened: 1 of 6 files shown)


src/utils/BufferPort.hxx
```diff
@@ -36,6 +36,8 @@
 #ifndef _UTILS_BUFFERPORT_HXX_
 #define _UTILS_BUFFERPORT_HXX_
 
+#include "utils/LimitedPool.hxx"
+
 /// A wrapper class around a string-based Hub Port that buffersthe outgoing
```
atanisoft: typo near the end of line

balazsracz: fixed

src/utils/BufferPort.hxx
```diff
@@ -171,6 +182,10 @@ private:
         BufferPort *parent_; ///< what to notify upon timeout.
     } bufferTimer_{this}; ///< timer instance.
 
+    /// Pool implementation that limits the number of buffers allocateable to
```
atanisoft: allocatable

balazsracz: fixed

atanisoft:
is there an estimate on the capacity for the esp32? Should we add that to the examples?

balazsracz:
Capacity in terms of what?

--
Sent from my mobile device

atanisoft:
Just thinking if we should add an override to the defaults on the esp32 or not.

balazsracz:
At the end of my work I'll have a bunch of overrides which we should push for the esp32. I'm not yet ready with them as I'm still pursuing problems.

PR #277, opened 2019-04-12
Makes the ESP32 CAN driver wrapper print fewer warnings about overrun.

balazsracz:
Limits the number of times we print "overrun, frame dropped" message to 1/ten seconds.
Prints the number of packets dropped due to overruns in the stats line.
This fixes the problem that printing the overrun warnings uses so much CPU that it causes more and more overruns.

### DIFF
--- a/src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
+++ b/src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
@@ -63,8 +63,11 @@ public:
     /// transceiver RX.
     /// @param txPin is the ESP32 pin that is connected to the external
     /// transceiver TX.
-    Esp32HardwareCan(const char *name, gpio_num_t rxPin, gpio_num_t txPin, bool reportStats=true)
-        : Can(name), reportStats_(reportStats)
+    Esp32HardwareCan(const char *name, gpio_num_t rxPin, gpio_num_t txPin,
+        bool reportStats = true)
+        : Can(name)
+        , reportStats_(reportStats)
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
```diff
@@ -129,7 +132,9 @@ private:
 
     /// Enables/Disables the periodic reporting of CAN bus statistics to the
     /// default serial stream.
-    bool reportStats_;
+    bool reportStats_ : 1;
+    /// Set to true of the frame sropped warning is printed.
```
atanisoft: Sropped - stopped
Of - if

src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
```diff
@@ -129,7 +132,9 @@ private:
...
+    bool warningPrinted_ : 1;
```
atanisoft: Perhaps naming as overrunWarningPrinted to clarify it's usage further?

src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
balazsracz: done (dropped actually)

src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
balazsracz: good idea. done.

PR #278, opened 2019-04-12
Adds an implementation of Atomic for the esp32 using spinlocks.

balazsracz:
This is a lot cheaper than mutexes.


balazsracz:
> Looks good, we might want to consider docs on Atomic though (it's missing for the base version as well).

done

PR #282, opened 2019-04-12
arduino-esp32 1.0.2-rc2 (release candidate for 1.0.2) compatiblity fix

atanisoft:
With v1.0.2-rc2 picking up ESP-IDF v3.2 there are a few adjustments
required for the esp_wifi module, specifically the log enabling APIs are
unavailable in ESP-IDF v3.2 as they were added in ESP-IDF v3.3.
arduino-esp32 1.0.1 and 1.0.2-rc1 were using beta versions of ESP-IDF
v3.3. Since v1.0.2 is soon to be released it is necessary to add a
compatibility check for this functionality that is not available.


atanisoft:
> Maybe we should refactor the duplicated code to a helper function.

Makes sense, I've added a private method to the class to handle that so there is no duplication.

PR #287, opened 2019-04-14
Refactor SimpleStack object hierarchy to prepare for TCP stack

balazsracz:
Refactors SimpleCanStack class hierarchy in order to prepare for a SimpleTcpStack.

Specifically:
- creates SimpleStackBase with all objects that are defined at the
  GenMessage level interface.
- creates SimpleCanStackBase as a descendant that knows about the
  actual interface type and how to create that interface.
- moves all Can-specific functions (e.g. add_gridconnect_port) to the child class.

To maintain the required object construction order
(executor -> service -> interface -> generic handlers)
the interface object is not created by the child class,
but by a factory function passed into the base class
constructor.

Adds a SimpleStack unit test to check that the class compiles and links.

### DIFF
--- a/src/openlcb/SimpleStack.hxx
+++ b/src/openlcb/SimpleStack.hxx
@@ -38,13 +38,14 @@
 #include <fcntl.h>
 
 #include "executor/Executor.hxx"
+#include "nmranet_config.h"
 #include "openlcb/AliasAllocator.hxx"
 #include "openlcb/ConfigRepresentation.hxx"
 #include "openlcb/ConfigUpdateFlow.hxx"
 #include "openlcb/DatagramCan.hxx"
 #include "openlcb/DefaultNode.hxx"
-#include "openlcb/EventService.hxx"
 #include "openlcb/EventHandlerTemplates.hxx"
+#include "openlcb/EventService.hxx"
 #include "openlcb/IfCan.hxx"
 #include "openlcb/MemoryConfig.hxx"
 #include "openlcb/NodeInitializeFlow.hxx"
@@ -52,7 +53,6 @@
 #include "openlcb/SimpleNodeInfo.hxx"
 #include "openlcb/TractionTrain.hxx"
 #include "openlcb/TrainInterface.hxx"
-#include "nmranet_config.h"
 #include "utils/GcTcpHub.hxx"
 #include "utils/GridConnectHub.hxx"
 #include "utils/HubDevice.hxx"
@@ -61,7 +61,8 @@
 #include "utils/HubDeviceSelect.hxx"
 #endif
 
-namespace openmrn_arduino {
+namespace openmrn_arduino
+{
 class OpenMRN;
 }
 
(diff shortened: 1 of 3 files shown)


src/openlcb/SimpleStack.hxx
```diff
@@ -358,40 +258,216 @@ protected:
...
+    /// flow usually has two members: the interface object from the software
+    /// stack and the hardware connection via which to connect to the physical
+    /// bus (which may be a device driver or a gridconnect protocol converter).
+    CanHubFlow *can_hub()
+    {
+        return &static_cast<CanPhysicalIf *>(ifaceHolder_.get())->canHub0_;
+    }
+
+    /// Adds a CAN bus port with synchronous driver API.
+    void add_can_port_blocking(const char *device)
```
atanisoft: Should this be masked for Linux/Mach since it is not valid for esp32?

src/openlcb/SimpleStack.hxx
```diff
@@ -358,40 +258,216 @@ protected:
...
+    /// @param on_error Notifiable to wakeup on error
+    void add_can_port_select(int fd, Notifiable *on_error = nullptr)
+    {
+        auto *port = new HubDeviceSelect<CanHubFlow>(can_hub(), fd, on_error);
+        additionalComponents_.emplace_back(port);
+    }
+#endif
+
+    /// Adds a gridconnect port to the CAN bus.
+    void add_gridconnect_port(const char *path, Notifiable *on_exit = nullptr);
```
atanisoft: Possibly same on this method.

atanisoft: Also is there much difference between this and the method below?

src/openlcb/SimpleStack.hxx
balazsracz: This is only using POSIX APIs, so there should be no compilation problem on the esp32.
(There is no socketcan compatible fd-based driver in esp32 today, so it's not very useful but since it can compile it makes not a lot of sense to add conditional compilation around it.)

atanisoft: Sure, just kinda thinking out loud a bit on it.

src/openlcb/SimpleStack.hxx
balazsracz: on linux you want to use the TTY version because it uses the terminal kernel API to set the baud rate and most importantly disable echo.

balazsracz:
> Aside from a couple minor questions on methods being masked this looks good overall.
> 
> One side question, what is the plan for can + tcp stack? Do we want to refactor SimpleStackBase just a small bit further to support something like: void add_iface(If *) ?

At the moment I don't have an easy or immediate way to have can+tcp interface. There is a bridge program from Robert Heller that can bridge a CAN-network and a native TCP hub.
We would need a major refactoring to allow for CAN+TCP interface in one. Writing a standalone gateway might be easier.

atanisoft:
Ok, just thinking on the esp32 example where we have the can/tcp bridge today and not sure how this will work with this update. I suppose it will be something to think about further.

src/openlcb/SimpleStack.hxx
atanisoft: Ok that makes sense.

balazsracz:
> Ok, just thinking on the esp32 example where we have the can/tcp bridge today and not sure how this will work with this update. I suppose it will be something to think about further.

On the ESP32 we can have a CAN <> TCP-CAN gateway (as we do today). As a compile time switch we can make a native-TCP client as well. We don't have the code yet for native-TCP server.

At the moment since JMRI does not support connecting to native TCP, the most usability is in the TCP-CAN anyway.

PR #288, opened 2019-04-14
Add TCP version of SimpleStack

balazsracz:
This adds the TCP implementations into the SimpleStack class hierarchy:
- SimpleTcpStackBase that instantiates the TCP protocol interface
- SimpleTcpStack which has the DefaultNode.
In addition:
- Adds a create/destroy unit test for SimpleTcpStack
- Updates tcp_blink_client app to use SimpleTcpStack.

### DIFF
--- a/src/openlcb/SimpleStack.hxx
+++ b/src/openlcb/SimpleStack.hxx
@@ -446,8 +446,13 @@ private:
         {
             return &datagramService_;
         }
+        /// This flow is the connection between the stack and the device
+        /// drivers. It also acts as a hub to multiple different clients or CAN
+        /// ports.
         CanHubFlow canHub0_;
+        /// Implementation of OpenLCB interface.
         IfCan ifCan_;
+        /// Datagram service (and clients) matching the interface.
         CanDatagramService datagramService_;
     };
 
@@ -466,7 +471,7 @@ public:
     /// link.
     HubFlow *tcp_hub()
     {
-        return &static_cast<TcpPhysicalIf *>(ifaceHolder_.get())->tcpHub0_;
+        return &static_cast<TcpPhysicalIf *>(ifaceHolder_.get())->tcpHub_;
     }
 
 protected:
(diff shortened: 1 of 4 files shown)


src/openlcb/SimpleStack.hxx
```diff
@@ -456,6 +457,60 @@ private:
...
+        {
+            return &ifTcp_;
+        }
+        /// @return the Datagram service bound to the interface. Ownership is
+        /// not transferred.
+        DatagramService *datagram_service() override
+        {
+            return &datagramService_;
+        }
+        HubFlow tcpHub0_;
```
atanisoft: What is the significance of the 0 at the end? Also since this gets assigned a Service object, is this the right name/type?

src/openlcb/SimpleStack.hxx
```diff
@@ -456,6 +457,60 @@ private:
...
+        IfTcp ifTcp_;
+        TcpDatagramService datagramService_;
```
atanisoft: Should we get doxygen notes for these vars?

src/openlcb/SimpleStack.hxx
balazsracz: 0 is historical and I removed it now. Originally it was for CAN bus because some MCUs have two can-bus (can0 and can1).

The service object that it gets on the constructor is only a lightweight wrapper around the executor. It is only used to ensure that the flow runs on the main executor of the stack object.

src/openlcb/SimpleStack.hxx
balazsracz: Done.

src/openlcb/SimpleStack.hxx
atanisoft: I figured it was related to that, but it makes sense to drop the zero here since there isn't really a hardware device backing the tcp stack (network interface excluded).

The note on the service object makes sense and explains the type as well.

PR #289, opened 2019-04-14
Switches to the proper method for allocating links to the TCP interface.

balazsracz:
Hides the TCP hub object as it is an internal detail that is not safe for exposing to customers.


balazsracz:
The CAN hub does not have the same problem, because it is strongly typed (contains one struct can_frame in each message) so there is no segmentation problem.
The difficulty with TCP hub is that the HubFlow carries string type objects and that means we need to take care for segmentation and unintended interleave of messages. This is why hiding it is better.

atanisoft:
Ok that makes sense.

PR #291, opened 2019-05-18
CC32xxWiFi minor updates

balazsracz:
- Adds a helper function to allow unit tests to simulate various connection states under linux.x86 target (where most of the code is not compiled in)
- Adds support for detecting when the user supplied the wrong password
- Fixes bug that the connected/ipacquired state bits were not following reality when the WPS process was initiated.
- Make wlan_connect not touch the blinker and return immediately.  This API makes much more sense, as the connection status can be monitored via various status functions.


src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -1015,17 +1021,17 @@ void CC32xxWiFi::wlan_event_handler(WlanEvent *event)
...
-            if (ipAcquiredCallback_)
-            {
-                ipAcquiredCallback_(false);
-            }
-
-            if(SL_WLAN_DISCONNECT_USER_INITIATED == disconnect->ReasonCode)
+            if(SL_WLAN_DISCONNECT_SECURITY_FAILURE == disconnect->ReasonCode)
             {
+                securityFailure = 1;
+            } else {
```
bakerstu: brace style.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.hxx
```diff
@@ -219,6 +223,21 @@ public:
         return WlanState::OK;
     }
 
+    /** Used by unit tests to simulate wifi connection states.
```
bakerstu: Should we put a #if defined (GTEST) wrapper around this?

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
balazsracz: fixed

src/freertos_drivers/net_cc32xx/CC32xxWiFi.hxx
balazsracz: done

PR #296, opened 2019-05-31
Adding rest of src/dcc files to the Arduino library script.

atanisoft:
This PR adds the RailCom and other src/dcc files to the Arduino OpenMRNLite library.

Prepping for inclusion in the ESP32 Command Station code.

### DIFF
--- a/arduino/libify.sh
+++ b/arduino/libify.sh
@@ -138,7 +138,8 @@ copy_file src arduino/OpenMRNLite.{h,cpp} \
     include/freertos/{freertos_includes.h,endian.h} \
     include/freertos_select/ifaddrs.h
 
-copy_file src/dcc src/dcc/*.hxx src/dcc/*.h src/dcc/Packet.cxx
+copy_file src/dcc src/dcc/*.hxx src/dcc/*.h \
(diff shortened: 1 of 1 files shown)


arduino/libify.sh
```diff
@@ -162,7 +173,8 @@ copy_file src/sys include/sys/tree.hxx
 
 copy_file src/utils src/utils/*.{cxx,hxx,c,h}
 
-rm -f ${TARGET_LIB_DIR}/src/utils/ReflashBootloader.cxx
+rm -f ${TARGET_LIB_DIR}/src/utils/ReflashBootloader.cxx \
+    ${TARGET_LIB_DIR}/src/utils/if_tcp_test_helper.hxx
```
balazsracz: we have a big bunch of other _test_helpers.hxx, do you want to exclude those too?

atanisoft: I was thinking of doing that just haven't had a chance to track all of them down yet. I'm out of town this week but will be home next Tues July 16) and can go over the list.

atanisoft: @balazsracz added commits below to remove the helper bits that I could easily identify (test in filename and not referenced from cxx/hxx files outside of tests)

PR #302, opened 2019-07-29
Adds encryption capability to the CC32xx AES library.

balazsracz:
- Adds encryption capability to the CC32xx AES library.

- Refactors the comparison test runner to a separate file so that
  a dev-build project can include the test runner and run it on real
  hardware.

- Moves the additional test vectors into a header.


atanisoft:
esp32 also has hardware AES so an OS neutral abstraction might make sense.

balazsracz:
Writing a crypto library (e.g. one with a sensible API) is a quite big undertaking, and the use-cases are extremely limited in this field. If someone wants to contribute a secure bootloader to openmrn (in opensource), it would make sense to refactor this, but probably not before that.

PR #305, opened 2019-08-01
Esp32WiFiManager Updates (SoftAP support)

atanisoft:
1) Adding SoftAP support using the provided ssid/password when running in AP only mode. When operating in AP+STA the SoftAP will use the hostname for it's SSID.
2) Use the full system_event_t object in process_wifi_event.
3) Add support for callers to ask the wifi mgr for event callbacks.
4) Added defines for silencing most of the Esp32SocketParams::log_message output.
5) Changed to using ESP_ERROR_CHECK_WITHOUT_ABORT to mdns_lookup rather than checking/logging the error in this method.
6) Added define to silence mdns_lookup output for search result output.
7) Shorten task name to "Esp32WiFiMgr".
8) Move mdns system init to STA start event handler.
9) Performance/stability improvements to the LwIP stack based on values coming in arduino-esp32 1.0.3.
10) #ifdef -> #if for feature flags
11) Add check for WDT being enabled before using APIs
12) Expose Notifiable::notify_from_isr based on OPENMRN_FEATURE_MUTEX_FREERTOS (underlying impl uses this flag)
13) Increase of LOG(...) buffer size to 2kb for ESP32
14) Replace a couple lingering references to HAVE_BSDSOCKET with OPENMRN_FEATURE_BSD_SOCKET.

### DIFF
--- a/src/executor/Notifiable.hxx
+++ b/src/executor/Notifiable.hxx
@@ -46,7 +46,7 @@ class Notifiable : public Destructable
 public:
     /// Generic callback.
     virtual void notify() = 0;
-#ifdef __FreeRTOS__
+#if OPENMRN_FEATURE_MUTEX_FREERTOS
     virtual void notify_from_isr()
     {
         DIE("Unexpected call to notify_from_isr.");
@@ -71,7 +71,7 @@ public:
         sem_.post();
     }
 
-#ifdef __FreeRTOS__
+#if OPENMRN_FEATURE_MUTEX_FREERTOS
     /// Implementation of notification receive from a FreeRTOS interrupt
     /// context.
     void notify_from_isr() OVERRIDE
--- a/src/executor/Notifiable.hxx
+++ b/src/executor/Notifiable.hxx
@@ -46,12 +46,12 @@ class Notifiable : public Destructable
 public:
     /// Generic callback.
     virtual void notify() = 0;
-#if OPENMRN_FEATURE_MUTEX_FREERTOS
+#if OPENMRN_FEATURE_RTOS_FROM_ISR
     virtual void notify_from_isr()
     {
         DIE("Unexpected call to notify_from_isr.");
     }
-#endif
+#endif // OPENMRN_FEATURE_RTOS_FROM_ISR
 };
 
 /// A Notifiable for synchronously waiting for a notification.
@@ -71,7 +71,7 @@ public:
         sem_.post();
     }
 
-#if OPENMRN_FEATURE_MUTEX_FREERTOS
+#if OPENMRN_FEATURE_RTOS_FROM_ISR
     /// Implementation of notification receive from a FreeRTOS interrupt
     /// context.
     void notify_from_isr() OVERRIDE
@@ -79,7 +79,7 @@ public:
         int woken = 0;
         sem_.post_from_isr(&woken);
     }
-#endif
+#endif // OPENMRN_FEATURE_RTOS_FROM_ISR
 
     /// Blocks the current thread until the notification is delivered.
     void wait_for_notification()
--- a/src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
+++ b/src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
@@ -240,7 +240,7 @@ private:
             }
 
             /// ESP32 native CAN driver frame
-            can_message_t msg = {0};
+            can_message_t msg = {};
 
             msg.flags = CAN_MSG_FLAG_NONE;
             msg.identifier = can_frame->can_id;
@@ -306,7 +306,7 @@ private:
 #endif // CONFIG_TASK_WDT
 
             /// ESP32 native CAN driver frame
-            can_message_t msg = {0};
+            can_message_t msg = {};
             if (can_receive(&msg, pdMS_TO_TICKS(250)) != ESP_OK)
             {
                 // native CAN driver did not give us a frame.
--- a/src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
+++ b/src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
@@ -169,16 +169,20 @@ private:
         /// txBuf.
         Esp32HardwareCan *parent = reinterpret_cast<Esp32HardwareCan *>(can);
 
+#if CONFIG_TASK_WDT
         // Add this task to the WDT
         esp_task_wdt_add(parent->txTaskHandle_);
+#endif // CONFIG_TASK_WDT
 
         /// Tracks the last time that we displayed the CAN driver status.
         TickType_t next_status_display_tick_count = 0;
 
         while (true)
         {
+#if CONFIG_TASK_WDT
             // Feed the watchdog so it doesn't reset the ESP32
             esp_task_wdt_reset();
+#endif // CONFIG_TASK_WDT
 
             // periodic CAN driver monitoring and reporting, this takes care of
             // bus recovery when the CAN driver disables the bus due to error
@@ -289,13 +293,17 @@ private:
         /// Get handle to our parent Esp32HardwareCan object to access the rxBuf
         Esp32HardwareCan *parent = reinterpret_cast<Esp32HardwareCan *>(can);
 
+#if CONFIG_TASK_WDT
         // Add this task to the WDT
         esp_task_wdt_add(parent->rxTaskHandle_);
+#endif // CONFIG_TASK_WDT
 
         while (true)
         {
+#if CONFIG_TASK_WDT
             // Feed the watchdog so it doesn't reset the ESP32
             esp_task_wdt_reset();
+#endif // CONFIG_TASK_WDT
 
             /// ESP32 native CAN driver frame
             can_message_t msg = {0};
--- a/src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
+++ b/src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
@@ -240,7 +240,8 @@ private:
             }
 
             /// ESP32 native CAN driver frame
-            can_message_t msg = {};
+            can_message_t msg;
+            bzero(&msg, sizeof(can_message_t));
 
             msg.flags = CAN_MSG_FLAG_NONE;
             msg.identifier = can_frame->can_id;
@@ -306,7 +307,8 @@ private:
 #endif // CONFIG_TASK_WDT
 
             /// ESP32 native CAN driver frame
-            can_message_t msg = {};
+            can_message_t msg;
+            bzero(&msg, sizeof(can_message_t));
             if (can_receive(&msg, pdMS_TO_TICKS(250)) != ESP_OK)
             {
                 // native CAN driver did not give us a frame.
--- a/src/freertos_drivers/esp32/Esp32WiFiManager.cxx
+++ b/src/freertos_drivers/esp32/Esp32WiFiManager.cxx
@@ -39,6 +39,7 @@
 #include <esp_log.h>
 #include <esp_wifi.h>
 #include <esp_wifi_internal.h>
+#include <lwip/dns.h>
 #include <mdns.h>
 #include <rom/crc.h>
 #include <tcpip_adapter.h>
@@ -52,6 +53,18 @@ using openlcb::TcpDefs;
 using openlcb::TcpManualAddress;
 using std::string;
 using std::unique_ptr;
+
+#ifndef ESP32_WIFIMGR_SOCKETPARAMS_LOG_LEVEL
+/// Allows setting the log level for mDNS related log messages from 
+/// @ref DefaultSocketClientParams.
+#define ESP32_WIFIMGR_SOCKETPARAMS_LOG_LEVEL INFO
+#endif
+
+#ifndef ESP32_WIFIMGR_MDNS_LOOKUP_LOG_LEVEL
+/// Allows setting the log level for mDNS results in the @ref mdns_lookup
+/// method.
+#define ESP32_WIFIMGR_MDNS_LOOKUP_LOG_LEVEL INFO
+#endif
 
 // Start of global namespace block.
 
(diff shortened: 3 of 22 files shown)


src/executor/Notifiable.hxx
```diff
@@ -46,7 +46,7 @@ class Notifiable : public Destructable
 public:
     /// Generic callback.
     virtual void notify() = 0;
-#ifdef __FreeRTOS__
+#if OPENMRN_FEATURE_MUTEX_FREERTOS
```
balazsracz: I'm thinking whether this is the best feature macro to use here, vs. adding a new feature test macro.
Ultimately _from_isr() would need to exist for any RTOS, including an RTOS that has say pthreads APIs.

src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
```diff
@@ -236,7 +240,7 @@ private:
             }
 
             /// ESP32 native CAN driver frame
-            can_message_t msg = {0};
+            can_message_t msg = {};
```
balazsracz: what is the reason for this change? Will the new code also zero initialize?

src/os/OSSelectWakeup.cxx
```diff
@@ -10,7 +10,7 @@ void empty_signal_handler(int)
 #include <sys/stat.h>
 #include <fcntl.h>
 
-#include "../vfs/esp_vfs.h"
+#include <esp_vfs.h>
```
balazsracz: are you sure this also compiles under the arduino UI (as opposed to native IDF and whatever build system you use).

src/utils/logging.cxx
```diff
@@ -35,6 +35,8 @@
 
 #if defined(__linux__) || defined(__MACH__)
 char logbuffer[4096];
+#elif defined(ESP32)
+char logbuffer[2048];
```
balazsracz: 256 was not enough for your log lines?

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -372,93 +407,238 @@ void Esp32WiFiManager::factory_reset(int fd)
...
-            // connect process.
-            if (xEventGroupGetBits(wifiStatusEventGroup_) & WIFI_CONNECTED_BIT)
+        // Set the default mDNS instance name to the generated hostname.
+        ESP_ERROR_CHECK(mdns_instance_name_set(hostname_.c_str()));
+
+        if (stationStaticIP_)
+        {
+            // Stop the DHCP service before connecting, this allows us to
+            // specify a static IP address for the WiFi connection
+            LOG(INFO, "[DHCP] Stoping DHCP Client (if running).");
```
balazsracz: pp

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -372,93 +407,238 @@ void Esp32WiFiManager::factory_reset(int fd)
...
+        // Start the SSID connection process.
+        esp_wifi_connect();
+    }
+    else if (event->event_id == SYSTEM_EVENT_STA_CONNECTED)
+    {
+        LOG(INFO, "[WiFi] Connected to SSID: %s", ssid_);
+        // Set the flag that indictes we are connected to the SSID.
+        xEventGroupSetBits(wifiStatusEventGroup_, WIFI_CONNECTED_BIT);
+    }
+    else if (event->event_id == SYSTEM_EVENT_STA_GOT_IP)
```
balazsracz: is this invoked also when we are in static IP mode or in AP mode?

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -372,93 +407,238 @@ void Esp32WiFiManager::factory_reset(int fd)
...
+                                          softAPStaticIP_));
+
+            // Convert the Soft AP Static IP to a uint32 for manipulation
+            uint32_t apIP = ntohl(ip4_addr_get_u32(&softAPStaticIP_->ip));
+
+            // Default configuration is for DHCP addresses to follow
+            // immediately after the static ip address of the Soft AP.
+            ip4_addr_t first_ip, last_ip;
+            ip4_addr_set_u32(&first_ip, htonl(apIP + 1));
+            ip4_addr_set_u32(&last_ip, htonl(apIP + softAPMaxStations_));
```
balazsracz: reserve a few dozen IPs instead. If someone connects a wifi extender you get to deal with a lot more MAC addresses than 4.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -507,109 +684,156 @@ void Esp32WiFiManager::start_wifi_system()
...
+
+    // If we are using the STATION interface, this will block until the ESP32
+    // starts the connection process, note it may not have an IP address
+    // immediately thus the need to check the connection result a few times
+    // before giving up with a FATAL error.
+    if (wifiMode_ == WIFI_MODE_APSTA || wifiMode_ == WIFI_MODE_STA)
+    {
+        uint8_t attempt = 0;
+        EventBits_t bits = 0;
+        uint32_t bitMask = WIFI_CONNECTED_BIT;
```
balazsracz: bit_mask

src/os/OSSelectWakeup.cxx
atanisoft: I tested it on the 1.0.2 arduino-esp32 release and didn't see any failures. I also checked the default include path for 1.0.1 and newer versions and the vfs include directory is present

src/utils/logging.cxx
atanisoft: For the CS it wasn't. Perhaps this could be smaller 1k instead of 2k though.

src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
atanisoft: Gcc complained on incomplete initializer. This is supposed to zero it as well but I have no problems putting in a memset instead of you prefer

src/executor/Notifiable.hxx
atanisoft: I wasn't sure either and considered adding a new define but all usages of this appeared to be on OSSem.notify_from_isr which used this define.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: Esp32 only supports a max of 4 connected stations. I can increase it though.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: Pp?

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -289,15 +324,15 @@ ConfigUpdateListener::UpdateAction Esp32WiFiManager::apply_configuration(
...
     if (lseek(fd, cfg_.offset(), SEEK_SET) != cfg_.offset())
     {
-        LOG(WARNING, "lseek failed to reset fd offset, REBOOT_NEEDED");
+        LOG_ERROR("lseek failed to reset fd offset, REBOOT_NEEDED");
         return ConfigUpdateListener::UpdateAction::REBOOT_NEEDED;
     }
 
     // If we are unable to read the full configuration from persistent storage
     // give up and request a reboot.
     if (read(fd, crcbuf.get(), cfg_.size()) != cfg_.size())
```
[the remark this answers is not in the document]
atanisoft: I'll look at the repeated_read and switch to that.

src/freertos_drivers/esp32/Esp32WiFiManager.hxx
```diff
@@ -85,11 +87,42 @@ public:
...
+    /// @param hostname_prefix is the hostname prefix to use for this node.
+    /// The @ref NodeID will be appended to this value. The maximum length for
+    /// final hostname is 32 bytes.
+    /// @param wifi_mode is the WiFi operating mode. When set to WIFI_MODE_STA
+    /// the Esp32WiFiManager will attempt to connect to the provided WiFi SSID.
+    /// When the wifi_mode is WIFI_MODE_AP the Esp32WiFiManager will create an
+    /// AP with the provided SSID and PASSWORD. When the wifi_mode is
+    /// WIFI_MODE_APSTA the Esp32WiFiManager will connect to the provided WiFi
+    /// AP and create an AP with the SSID of "<hostname>" and the provided
+    /// password. Note, the password for the AP will not be used if
```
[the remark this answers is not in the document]
atanisoft: I can add a pw for AP with default to the station pw when null and mode is not open

src/freertos_drivers/esp32/Esp32WiFiManager.hxx
```diff
@@ -85,11 +87,42 @@ public:
...
     /// node uptime.
     Esp32WiFiManager(const char *ssid, const char *password,
-        openlcb::SimpleCanStack *stack, const WiFiConfiguration &cfg);
+        openlcb::SimpleCanStack *stack, const WiFiConfiguration &cfg,
+        const char *hostname_prefix = "esp32_",
+        wifi_mode_t wifi_mode = WIFI_MODE_STA,
+        tcpip_adapter_ip_info_t *station_static_ip = nullptr,
+        ip_addr_t primary_dns_server = ip_addr_any,
+        uint8_t soft_ap_channel = 1,
+        uint8_t soft_ap_max_stations = 4,
```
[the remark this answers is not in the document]
atanisoft: Likely not, at least I can't think of anything specific. On the CS this is intended mainly as means for initial config but it can also be used for non-home (ie: train show) environments.

src/freertos_drivers/esp32/Esp32WiFiManager.hxx
```diff
@@ -189,8 +236,9 @@ private:
     /// periodic health checks of the connected hubs or clients.
     os_thread_t wifiTaskHandle_;
 
-    /// Dynamically generated hostname for this node, esp32_{node-id}.
-    std::string hostname_{"esp32_"};
+    /// Dynamically generated hostname for this node, esp32_{node-id}. This is
+    /// also used for the SoftAP SSID name (if enabled).
+    std::string hostname_{""};
```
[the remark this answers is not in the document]
atanisoft: Will drop, I shifted it to the constructor and forgot to clean this up

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: Yes, it is raised almost immediately after connect in that case. I can add it to the bit_mask when static IP is used so we don't block except for connect

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
balazsracz: yes, you have a typo here.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
balazsracz: no that's fine. Less code is better. Just wanted to make sure it actually works :)

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
balazsracz: The number of stations does not necessarily equal to the number of MAC addresses visible. If you add a wifi extender, you will see only one station (the wifi extender) but a bunch of MAC addresses that need separate IP addresses.

src/freertos_drivers/esp32/Esp32WiFiManager.hxx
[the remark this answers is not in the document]
balazsracz: sounds good

src/freertos_drivers/esp32/Esp32WiFiManager.hxx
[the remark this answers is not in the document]
balazsracz: i would recommend to remove this configuration option as it is unnecessary complexity in your code.

src/executor/Notifiable.hxx
balazsracz: let's add a new feature macro OPENMRN_FEATURE_NOTIFY_FROM_ISR

src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
atanisoft: switching to memset instead.

src/utils/logging.cxx
atanisoft: dropped to 1k

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: got it, setting up a constant with a default value of 48 and capping at max of 4 stations (max for esp32).

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: fixed

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: sorry missed that, fixed

src/freertos_drivers/esp32/Esp32WiFiManager.hxx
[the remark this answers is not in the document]
atanisoft: removed

src/executor/Notifiable.hxx
atanisoft: I'll update the refs that are in this PR to this, I spotted a few others that might make sense but will hold them back for a diff PR since I'm not certain of their usages.

Issue #306, opened 2019-08-02
esp32: config file corruption (SD storage)


atanisoft:
On ESP32 CS I'm shifting to using SD for storage rather than SPIFFS and encountered the following failure after restarting the CS:
```
[Config] /sdcard/LCC/cdi.xml (6563 bytes)
[Config] /sdcard/LCC/config (0 bytes)
....
Crashed in file lib/OpenMRNLite/src/utils/FdUtils.hxx line 56: Unexpected EOF reading the config file.
assertion "0" failed: file "lib/OpenMRNLite/src/utils/FdUtils.hxx", line 56, function: static void FdUtils::repeated_read(int, void*, size_t)
abort() was called at PC 0x4016f753 on core 0
```

When using SPIFFS as the storage the configuration file is correctly flushed to SPIFFS immediately after write. With SD storage though the file does not automatically flush to the storage. A call to fsync(fd) is likely needed immediately after factory reset completes and after any config updates.

Likely a zero size check should be added [here](https://github.com/bakerstu/openmrn/blob/master/src/openlcb/SimpleStack.cxx#L257-L260) to force a reset when the file is explicitly zero bytes.

PR #307, opened 2019-08-03
handle zero byte internal config file

atanisoft:
fix for #306, when the file exists but is zero bytes we can't read from it to check the version so it should treated as a reset instead.

### DIFF
--- a/src/openlcb/SimpleStack.cxx
+++ b/src/openlcb/SimpleStack.cxx
@@ -258,6 +258,13 @@ int SimpleStackBase::create_config_file_if_needed(const InternalConfigData &cfg,
     {
         extend = true;
     }
+    // Handle the case where the file exists but is zero bytes. This was
+    // observed on the esp32 with SD storage which does not automatically
+    // flush to disk on write.
+    if (statbuf.st_size == 0)
+    {
+        reset = true;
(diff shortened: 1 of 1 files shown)


src/openlcb/SimpleStack.cxx
```diff
@@ -258,6 +258,13 @@ int SimpleStackBase::create_config_file_if_needed(const InternalConfigData &cfg,
     {
         extend = true;
     }
+    // Handle the case where the file exists but is zero bytes. This was
+    // observed on the esp32 with SD storage which does not automatically
+    // flush to disk on write.
+    if (statbuf.st_size == 0)
```
balazsracz: try to compare st_size against cfg.version().end_offset() instead.

atanisoft:
```
[LCC] Corrupt configuration file detected, /sdcard/LCC/config is too small: 0 bytes, expected: 489 bytes
/sdcard/LCC/config is too short (0 vs 130), forcing reset.
```
The first line is from the CS code to detect corruptions, the second line is a new LOG (VERBOSE) inside this new condition check. It appears to work correctly, I'll update the PR for this adjustment.

PR #308, opened 2019-08-03
Adding utility state flow for automatic calls to fsync

atanisoft:
Simple state flow which will periodically call fsync on the provided file handle, default interval is 1sec. This is primarily useful on the ESP32 when using SD for storage due to FFat utilizing a 512b (sector size) cache in memory for the open file.

### DIFF
--- a/src/utils/AutoSyncFileFlow.hxx
+++ b/src/utils/AutoSyncFileFlow.hxx
@@ -1,0 +1,83 @@
+/** \copyright
+ * Copyright (c) 2019, Mike Dunston
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  - Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ *  - Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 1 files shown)


TrainzLuvr:
I admit I did not read the spec on fsync and how it exactly works - IIRC calling fsync flushes the buffer and writes any files to the disk.

I'm having my doubts about this being a good idea. Every second there will be a write (or more) to the SD card. In a 4 hour session that's 14,400 writes.

IMHO will reduce SD card's lifespan significantly (most are already cheap pieces of crap unless you spend $ on a good one), so I wonder is this really necessary to do every second, or is there another way, perhaps to check whether a write should be made first?

bakerstu:
Typically, fsync will only write stale (cached) data, if implemented properly.  I don't know the intended application in this case, but I doubt that it would result in an SD card write on every call.  Rather than timer based sync, you could also utilize an event based sync whereby the application signals an event when it is done writing and want to sync.

balazsracz:
It's the responsibility of the system integrator to choose the right flush strategy. If you have an underlying system that writes to the card even when the file is not dirty, do not use this flow.

Mike checked on the ffat implementation he is using in his CS, and if it's not dirty flush is a noop.

atanisoft:
The esp32 FFat lib used for SD storage by default uses an in memory cache (512b, default sector size for SD but is configurable) and when a file is fully within this cache it won't be persisted to disk until the file closes. OpenMRN keeps the config file open for the duration of app runtime and the typical size of the config is smaller than the cache size.

FFat f_sync has a dirty check which will prevent unnecessary wear on the SD when the data in the cache has not been altered. f_write and f_read have similar dirty checks and will fsync automatically if the cached window shifts.

Adding an event based listener would be doable but could lead to race condition on the fsync call if the apply_configuration callback is still modifying the file. This could be solved by starting a timer to do bg fsync on expire with the timer triggered on config events.

TrainzLuvr:
It's all nice and dandy when devs put it into integrators corner to worry about things like this.

Consider that this software is being made for general public who will not know whether their system integration writes to the card or not, even when file is not dirty.

Matter a fact they will never know - most want a "fire and forget" integration. I believe we should take that extra step to create it like so, if we care for a wide adoption of this API.

Just my 2c.

atanisoft:
> Consider that this software is being made for general public who will not know whether their system integration writes to the card or not, even when file is not dirty.

Correct, but in the general case they will also be using SPIFFS instead of an SD. This only came up in my code due to shifting to SD storage. Further complicating this is the fact that the cache size and caching entirely can be reconfigured on a per-project basis! The default however for Arduino-esp32 is to have caching ON.

> Matter a fact they will never know - most want a "fire and forget" integration. I believe we should take that extra step to create it like so, if we care for a wide adoption of this API.

Quite true, there are a lot of end users who do not delve into the platforms they are working with to have any understanding of it and they will file issues as soon as anything odd comes up (even compilation failures!) I don't see this hxx file being the end of discussion on this topic, likely a doc note will need to be added as this is specific to the esp32 SD FS implementation. It may come up in other environments and such a utility flow could be reused.

balazsracz:
This software is not for end users. It is for manufacturers. Manufacturers
need to know when their flash performs writes otherwise they cannot plan
lifetime of their device.

atanisoft:
> This software is not for end users.

Sorry in my post this was intended not as the end user who uses a device which uses this library but instead the creator of such device. I should have clarified that point.

However the doc note point is still valid and I've added a note to the Arduino readme as that is one primary concern for where this code might be needed if the conditions are met.

PR #309, opened 2019-08-05
adding loop_executor() option

atanisoft:
Adding the loop_executor() method which can be used to donate the current running thread to the executor. WDT check is in place for ESP32 and when WDT is active it will warn rather than cause fatal crash.

### DIFF
--- a/arduino/OpenMRNLite.h
+++ b/arduino/OpenMRNLite.h
@@ -42,9 +42,10 @@
 #include "freertos_drivers/arduino/Can.hxx"
 #include "freertos_drivers/arduino/WifiDefs.hxx"
 #include "openlcb/SimpleStack.hxx"
-#include "utils/GridConnectHub.hxx"
-#include "utils/Uninitialized.hxx"
 #include "utils/FileUtils.hxx"
+#include "utils/GridConnectHub.hxx"
+#include "utils/logging.h"
+#include "utils/Uninitialized.hxx"
 
 #if defined(ESP32)
 
@@ -343,7 +344,7 @@ public:
     {
         for (auto *e : loopMembers_)
         {
-#if defined(ESP32)
+#if defined(ESP32) && CONFIG_TASK_WDT
             // Feed the watchdog so it doesn't reset the ESP32
             esp_task_wdt_reset();
 #endif // ESP32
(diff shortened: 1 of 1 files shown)


arduino/OpenMRNLite.h
```diff
@@ -343,36 +344,69 @@ class OpenMRN : private Executable
...
+        // check if watchdog is enabled and print a warning if it is
+        if (esp_task_wdt_status(idleTask) == ESP_OK)
+        {
+            LOG(WARNING, "WDT detected as enabled on core %d!", current_core);
+        }
+#endif // ESP32 && CONFIG_TASK_WDT
+        haveExecutorThread_ = true;
+
+        // donate this thread to the executor
+        stack_->executor()->thread_body();
```
balazsracz: pleae do stack_->loop_executor() instead of calling this directly.

arduino/OpenMRNLite.h
```diff
@@ -343,36 +344,69 @@ class OpenMRN : private Executable
...
+        // uses 100% cpu and it is pinned to CPU 0.
         disableCore0WDT();
+#endif // CONFIG_TASK_WDT_CHECK_IDLE_TASK_CPU0
+        xTaskCreatePinnedToCore(&thread_entry         // entry point
+                              , "OpenMRN"             // task name
+                              , OPENMRN_STACK_SIZE    // stack size
+                              , this                  // entry point arg
+                              , OPENMRN_TASK_PRIORITY // priority
+                              , nullptr               // task handle
+                              , PRO_CPU_NUM);         // cpu core
```
balazsracz: we weren't pinning before this PR. Is there a reason to pin now?

arduino/OpenMRNLite.h
atanisoft: that would call start_stack() again, is that the intent here? In the thread_entry method it also calls like this.

atanisoft: also should thread_entry call this method?

atanisoft: I tested locally and calling start_stack again has no impact so will switch accordingly. Open q on thread_entry calling this method instead of duplication of code.

balazsracz: apologies this was a bad suggestion from me. calling start_stack(...) more than once is not a good idea. I forgot that we already have a begin() call which does that. Since we do not want to retire support for begin() + loop() we should have as you originally write the code.

PR #310, opened 2019-08-06
adding unregister_train to TrainService.

atanisoft:
1. Code cleanup/reformat for consistency.
2. Adding docs to hxx file where missing.
3. Adding unregister_train(TrainNode *) to allow dynamic node add/remove capability.

### DIFF
--- a/src/openlcb/TractionTrain.cxx
+++ b/src/openlcb/TractionTrain.cxx
@@ -52,13 +52,15 @@ TrainNode::TrainNode(TrainService *service, TrainImpl *train)
 
 TrainNode::~TrainNode()
 {
-    while (!consistSlaves_.empty()) {
+    while (!consistSlaves_.empty())
+    {
         delete consistSlaves_.pop_front();
     }
 }
 
 TrainNodeForProxy::TrainNodeForProxy(TrainService *service, TrainImpl *train)
-    : TrainNode(service, train) {
+    : TrainNode(service, train)
+{
     service_->register_train(this);
 }
 
@@ -282,7 +284,9 @@ struct TrainService::Impl
                     p[1] = TractionDefs::CTRLRESP_ASSIGN_CONTROLLER;
                     NodeHandle supplied_controller = {0, 0};
                     if (size() < 9)
+                    {
                         return reject_permanent();
+                    }
                     supplied_controller.id = data_to_node_id(payload() + 3);
                     if (size() >= 11 && (payload()[2] & 0x01))
                     {
@@ -328,7 +332,9 @@ struct TrainService::Impl
                 {
                     NodeHandle supplied_controller = {0, 0};
                     if (size() < 9)
+                    {
                         return reject_permanent();
+                    }
                     supplied_controller.id = data_to_node_id(payload() + 3);
                     if (size() >= 11 && (payload()[2] & 0x01))
                     {
@@ -366,8 +372,10 @@ struct TrainService::Impl
                 case TractionDefs::CNSTREQ_ATTACH_NODE:
                 {
                     if (size() < 9)
+                    {
                         return reject_permanent(
                             Defs::ERROR_INVALID_ARGS_MESSAGE_TOO_SHORT);
+                    }
                     NodeID target = data_to_node_id(payload() + 3);
                     uint8_t flags = payload()[2];
                     if (target == train_node()->node_id())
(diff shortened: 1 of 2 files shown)


src/openlcb/TractionTrain.cxx
```diff
@@ -429,24 +446,32 @@ struct TrainService::Impl
             }
             uint8_t cmd = payload()[0];
             bool flip_speed = false;
-            if (cmd == TractionDefs::REQ_SET_SPEED) {
-                if (flags & TractionDefs::CNSTFLAGS_REVERSE) {
+            if (cmd == TractionDefs::REQ_SET_SPEED)
+            {
```
balazsracz: I appreciate cleanups in the code, but please do not mix a feature implementation with random cleanups. It makes it take unnecessarily long to review the change.
Two changes, one whitespace-only and one meaningful takes much shorter to review total.

src/openlcb/TractionTrain.cxx
```diff
@@ -628,4 +655,13 @@ void TrainService::register_train(TrainNode *node)
     HASSERT(nodes_.find(node) != nodes_.end());
 }
 
+void TrainService::unregister_train(TrainNode *node)
+{
+    iface_->delete_local_node(node);
+    AtomicHolder h(this);
```
balazsracz: If you need this you are likely violating the requirement of calling delete_local_node.
```
    /** Removes a local node from this interface. This function must be called
     * from the interface's executor. [...] */
    virtual void delete_local_node(Node *node) = 0;
```

As it stands now, you must repeat this requirement on the contract of unregister_train_node. We should probably make an assertion that the caller holds themselves to it (`service()->executor()->assert_current()`).

src/openlcb/TractionTrain.cxx
```diff
@@ -628,4 +655,13 @@ void TrainService::register_train(TrainNode *node)
...
+    nodes_.erase(node);
+    LOG(VERBOSE, "Unregistered node %p for traction.", node);
+    HASSERT(nodes_.find(node) == nodes_.end());
```
balazsracz: this is not necessary

src/openlcb/TractionTrain.cxx
atanisoft: Sorry let me pull this PR and clean it up in two PRs instead.

src/openlcb/TractionTrain.cxx
atanisoft: Probably right, I'm not 100% certain if this method is actually going to be needed or who exactly would call it. I suspect it should be called from the destructor of the classes which call register_train() though.

src/openlcb/TractionTrain.cxx
atanisoft: will remove from cleaned up PR

src/openlcb/TractionTrain.cxx
balazsracz: actually that does not matter anymore because the effort was spent -- just for the future please.

atanisoft: I see a few cleanups I missed as well in here. I'll resubmit this PR once the comments are addressed.

src/openlcb/TractionTrain.cxx
atanisoft: @balazsracz A follow up on this one, I added the following to both register_train and unregister_train:
```
iface_->executor()->assert_current();
```
I've also confirmed this is currently triggered with my current usage:
```
0x4008c285: invoke_abort at C:\users\mdunston\.platformio\packages\framework-espidf\components\esp32/panic.c:687
0x4008c4f5: abort at C:\users\mdunston\.platformio\packages\framework-espidf\components\esp32/panic.c:687
0x4017c77b: __assert_func at /Users/ivan/e/newlib_xtensa-2.2.0-bin/newlib_xtensa-2.2.0/xtensa-esp32-elf/newlib/libc/stdlib/../../../.././newlib/libc/stdlib/assert.c:63 (discriminator 8)
0x400f785a: ExecutorBase::assert_current() at c:\trains\ESP32CommandStation/src\stateflows/InfoScreen.cpp:190
0x4010c966: openlcb::TrainService::register_train(openlcb::TrainNode*) at c:\trains\ESP32CommandStation/lib/OpenMRNLite/src/openlcb/TractionTrain.hxx:226
0x4010c9e1: openlcb::TrainNodeForProxy::TrainNodeForProxy(openlcb::TrainService*, openlcb::TrainImpl*) at c:\trains\ESP32CommandStation/lib/OpenMRNLite/src/openlcb/TractionTrain.hxx:226
0x400e5921: Locomotive::Locomotive(unsigned short, openlcb::TrainService*) at c:\users\mdunston\.platformio\packages\toolchain-xtensa32\xtensa-esp32-elf\include\c++\5.2.0\bits/shared_ptr_base.h:127 (discriminator 1)
```
This came in from JMRI sending a DCC++ message for a loco. I'll be reworking that interface (and similar interfaces) to execute in a StateFlow similar to the WiThrottle/HubDevice flows (go to sleep until data arrives, parse into chunks, send it off to be processed, etc)

PR #313, opened 2019-08-14
Enable add_from_isr and wakeup_from_isr via OPENMRN_FEATURE_RTOS_FROM_ISR feature macro.

atanisoft:
The ESP32 ::select() wakeup from an ISR required the addition of OSSelectWakeup::esp_wakeup_from_isr() which uses ISR versions of those in OSSelectWakeup::esp_wakeup(). Note the LOG statement has been dropped as it is not safe from inside an ISR.

I wasn't sure if pthread_kill is safe to call from inside the ISR context so left it commented in wakeup_from_isr(). @balazsracz can you confirm on if this is safe or not?

### DIFF
--- a/src/utils/Atomic.hxx
+++ b/src/utils/Atomic.hxx
@@ -95,12 +95,34 @@ public:
     /// Locks the specific critical section.
     void lock()
     {
-        portENTER_CRITICAL(&mux);
+        // This should really use portENTER_CRITICAL_SAFE() but that is not
+        // available prior to ESP-IDF 3.3 which is not available in the
+        // arduino-esp32 environment generally. The below code is the
+        // implementation of that macro.
+        if (xPortInIsrContext())
+        {
+            portENTER_CRITICAL_ISR(&mux);
+        }
+        else
+        {
+            portENTER_CRITICAL(&mux);
+        }
     }
     /// Unlocks the specific critical section.
     void unlock()
     {
-        portEXIT_CRITICAL(&mux);
+        // This should really use portEXIT_CRITICAL_SAFE() but that is not
+        // available prior to ESP-IDF 3.3 which is not available in the
+        // arduino-esp32 environment generally. The below code is the
+        // implementation of that macro.
+        if (xPortInIsrContext())
+        {
+            portEXIT_CRITICAL_ISR(&mux);
+        }
+        else
+        {
+            portEXIT_CRITICAL(&mux);
+        }
     }
 
 private:
--- a/src/executor/Executor.hxx
+++ b/src/executor/Executor.hxx
@@ -87,7 +87,7 @@ public:
      * the execution is completed. @param fn is the closure to run. */
     void sync_run(std::function<void()> fn);
 
-#ifdef __FreeRTOS__
+#if OPENMRN_FEATURE_RTOS_FROM_ISR
     /** Send a message to this Executor's queue. Callable from interrupt
      * context.
      * @param action Executable instance to insert into the input queue
@@ -95,7 +95,7 @@ public:
      */
     virtual void add_from_isr(Executable *action,
                               unsigned priority = UINT_MAX) = 0;
-#endif
+#endif // OPENMRN_FEATURE_RTOS_FROM_ISR
 
     /** Adds a file descriptor to be watched to the select loop.
      * @param job Selectable structure that describes the descriptor to watch.
@@ -314,7 +314,7 @@ public:
 #endif
     }
 
-#ifdef __FreeRTOS__
+#if OPENMRN_FEATURE_RTOS_FROM_ISR
     /** Send a message to this Executor's queue. Callable from interrupt
      * context.
      * @param msg Executable instance to insert into the input queue
@@ -326,7 +326,7 @@ public:
             msg, priority >= NUM_PRIO ? NUM_PRIO - 1 : priority);
         selectHelper_.wakeup_from_isr();
     }
-#endif
+#endif // OPENMRN_FEATURE_RTOS_FROM_ISR
 
     /** If the executor was created with NO_THREAD, then this function needs to
      * be called to run the executor loop. It will exit when the execut gets
--- a/src/executor/Executor.hxx
+++ b/src/executor/Executor.hxx
@@ -332,8 +332,8 @@ public:
 #else
         queue_.insert_locked(
             msg, priority >= NUM_PRIO ? NUM_PRIO - 1 : priority);
-        selectHelper_.wakeup_from_isr();
 #endif // ESP32
+        selectHelper_.wakeup_from_isr();
     }
 #endif // OPENMRN_FEATURE_RTOS_FROM_ISR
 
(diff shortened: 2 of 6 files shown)


balazsracz:
did the esp32 had __FreeRTOS__ before?

I'm wondering if we indeed are in a good position to have ISR wakeups for the esp32. The problem is in all other cases ISR == critical section. Not on the esp32, you also need a spin loop there.

atanisoft:
> did the esp32 had `__FreeRTOS__` before?

No it did not. That is part of the reason for the various FEATURE macros.



> The problem is in all other cases ISR == critical section. Not on the esp32, you also need a spin loop there.

Right, in the case of ESP32 ISR though it is either a critical section or actual ISR. What sort of spin loop are you thinking is needed here?

Example usage can be see in [ENCODE_PACKET](https://github.com/atanisoft/ESP32CommandStation/blob/development/src/DCC/RMTTrackDevice.cpp#L293-L296) and in it's [usage](https://github.com/atanisoft/ESP32CommandStation/blob/development/src/DCC/RMTTrackDevice.cpp#L900-L903) (called from [here](https://github.com/atanisoft/ESP32CommandStation/blob/development/src/DCC/RMTTrackDevice.cpp#L258) to [here](https://github.com/atanisoft/ESP32CommandStation/blob/development/src/DCC/RMTTrackDevice.cpp#L650)).

balazsracz:
Let me give an example: stateflow:: notify_from_isr will call executor:: add_from_isr, which will call Queue::add_from_isr.
utils/Queue is a thread-safe structure and add needs locking. Since the locking is very simple and this is a very frequent operation, Queue uses an Atomic to protect itself. Atomic internally starts a critical section.
When the critical section is in effect, all interrupts that are allowed to call *_from_isr routines are masked.
Conversely, when we are in a *_from_isr call, we can be sure that no section protected by an atomic is running.

However, in a multi-core environment you can have an Atomic running in one core and a _from_isr running on the other core. So we could corrupt our queues and pretty much anything else that we are protecting with an Atomic.

We cannot add atomic holders in the _from_isr routines, because that would crash a real FreeRTOS.

atanisoft:
Ahh yes, I see what you are meaning. Perhaps an esp32 OSMux/Holder usage in the _from_isr methods? This would be similar to some of the spinlock impls in esp-idf.

atanisoft:
Actually OSMux/Holder will not be ISR safe without more modifications. However in ESP-IDF they use basically the Atomic object (portMUX_TYPE) with portENTER_CRITICAL_SAFE/portEXIT_CRITICAL_SAFE which is ISR/non-ISR aware. Let me test with an updated ESP32 Atomic impl using the _SAFE variant and add an AtomicHolder to the _from_isr *WHEN* on the ESP32.

atanisoft:
portENTER_CRITICAL_SAFE/portEXIT_CRITICAL_SAFE is not present in IDF 3.2.2 (latest stable) I'll use the impl from those macros instead.

balazsracz:
As this seems like a difficult architectural consideration, let's focus on getting you unblocked for whatever you were developing for your CS first.

I propose that we identify a "safe" set of use-cases where the _from_isr can be rolled out to the esp32, and find a structure for your commandstation code to use one of those concepts.

It seems clear that any use-case where our implementation is just a direct proxy of some FreeRTOS concept is within this safe range. So all the semaphores etc that you've enabled so far are safe.

A simple solution is to post to a semaphore and wake up a dedicated task that has then to notify the flow. We know that notifications coming from tasks are safe.

You are not planning to use the select implementation we have in openmrn for freertos, right? You are integrating the esp vfs? Do you have a select implementation for your device fd?
Maybe that would be a better alternative than trying to get this notify_from_isr to work.

atanisoft:
> You are not planning to use the select implementation we have in openmrn for freertos, right?

Yes, the esp_select wrapper you implemented is intended for use in the CS.

> You are integrating the esp vfs? Do you have a select implementation for your device fd?
Yes, the impl uses notify_from_isr on Notifiable today which routes back down to Executable.

> Maybe that would be a better alternative than trying to get this notify_from_isr to work.

I've got an alternative impl that appears to be working so far and would satisfy the critical section needs:
```
    void add_from_isr(Executable *msg, unsigned priority = UINT_MAX) override
    {
#ifdef ESP32
        // On the ESP32 we need to call insert instead of insert_locked to
        // ensure that all code paths lock the queue for consistency since
        // this code path is not guaranteed to be protected by a critical
        // section.
        queue_.insert(
            msg, priority >= NUM_PRIO ? NUM_PRIO - 1 : priority);
#else
        queue_.insert_locked(
            msg, priority >= NUM_PRIO ? NUM_PRIO - 1 : priority);
#endif // ESP32
        selectHelper_.wakeup_from_isr();
    }
```

And in Atomic for the ESP32 there will be the following blocks:
```
        if (xPortInIsrContext())
        {
            portENTER_CRITICAL_ISR(&mux);
        }
        else
        {
            portENTER_CRITICAL(&mux);
        }
```
lock/unlock would be similar.

atanisoft:
> > You are not planning to use the select implementation we have in openmrn for freertos, right?
> 
> Yes, the esp_select wrapper you implemented is intended for use in the CS.

I should clarify, that impl is intended for use in TCP connections only. If it is used by StateFlowSelectHelper though it could indirectly be invoked.

The "wakeup" flow from my side is [here](https://github.com/atanisoft/ESP32CommandStation/blob/development/src/DCC/RMTTrackDevice.cpp#L291-L296) with the Notifable object passed via ioctl [here](https://github.com/atanisoft/ESP32CommandStation/blob/development/src/DCC/RMTTrackDevice.cpp#L592-L596) from this [call](https://github.com/atanisoft/ESP32CommandStation/blob/development/src/DCC/LocalTrackIf.cpp#L68-L71).

atanisoft:
I've been testing this latest update for Executor on my CS and it appears to work but I'd like to get more testing on it to ensure stability.

src/utils/Atomic.hxx
```diff
@@ -95,12 +95,34 @@ public:
...
     void unlock()
     {
-        portEXIT_CRITICAL(&mux);
+        // This should really use portEXIT_CRITICAL_SAFE() but that is not
+        // available prior to ESP-IDF 3.3 which is not available in the
+        // arduino-esp32 environment generally. The below code is the
+        // implementation of that macro.
+        if (xPortInIsrContext())
+        {
+            portEXIT_CRITICAL_ISR(&mux);
```
balazsracz: if you check the implementation of this, you will see that it is exactly the same as the other branch.

atanisoft: For 3.2 yes, but in IDF 3.3/4.0 it is [changing](https://github.com/espressif/esp-idf/blob/6517c5033aab27afe491e433d6beeba93dfb1aa7/components/freertos/include/freertos/portmacro.h#L265-L294). This is a bit of future proofing .

atanisoft: If/When portENTER_CRITICAL_SAFE/portEXIT_CRITICAL_SAFE is available it would be better to move to that though.

PR #315, opened 2019-08-21
Adding shutdown() method to AutoSyncFileFlow

atanisoft:
Adding ability to shutdown the AutoSyncFileFlow so the underlying file/filesystem can be modified without the fsync call(s) being invoked.

The intent of this method is to discontinue automatic fsync calls prior to the unmounting of the VFS on the esp32 and prevent the ERRNOCHECK() call from calling abort() when fsync returns ERRNO=9 (bad file descriptor) since the underlying VFS has disappeared.

### DIFF
--- a/src/utils/AutoSyncFileFlow.hxx
+++ b/src/utils/AutoSyncFileFlow.hxx
@@ -62,6 +62,13 @@ public:
       start_flow(STATE(sleep_and_call_sync));
     }
 
+    /// Discontinues the automatic calls to fsync.
+    void shutdown()
+    {
+        timer_.cancel();
+        reset_flow(STATE(exit));
+    }
+
 private:
     const int fd_;
     const uint64_t interval_;
--- a/src/utils/AutoSyncFileFlow.hxx
+++ b/src/utils/AutoSyncFileFlow.hxx
@@ -62,11 +62,18 @@ public:
       start_flow(STATE(sleep_and_call_sync));
     }
 
-    /// Discontinues the automatic calls to fsync.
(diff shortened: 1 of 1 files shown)


atanisoft:
Usage pattern: [reboot()](https://github.com/atanisoft/ESP32CommandStation/blob/development/src/OpenMRNEsp32Overrides.cpp#L22) which calls the destructor for [ConfigurationManager](https://github.com/atanisoft/ESP32CommandStation/blob/development/src/ConfigurationManager.cpp#L259-L293) which unmounts the VFS after shutting down the executor.

src/utils/AutoSyncFileFlow.hxx
```diff
@@ -62,6 +62,13 @@ public:
       start_flow(STATE(sleep_and_call_sync));
     }
 
+    /// Discontinues the automatic calls to fsync.
+    void shutdown()
+    {
+        timer_.cancel();
```
balazsracz: This is not safe to call in production code (it's clearly marked in the documentation of that call).
I recommend the following pattern:
- add a boolean variable to shutdown
- at the beginning of the sync() state check the shutdown variable, if set, do `return exit()` instead of sync.

This guarantees that after calling shutdown there will be no further fsync calls without getting into race conditions of canceling timers.

atanisoft: I've switched to timer_.ensure_triggered() since this could be called from external. I've also added a Notifiable which will allow the caller to be know when the timer expires and fsync is stopped.

PR #316, opened 2019-08-23
Update CONSTANT_TRUE and CONSTANT_FALSE values

atanisoft:
@balazsracz @bakerstu if you can please review the updated constant values for other platforms applicability.

### DIFF
--- a/src/utils/constants.hxx
+++ b/src/utils/constants.hxx
@@ -115,25 +115,25 @@
 
 #endif // native C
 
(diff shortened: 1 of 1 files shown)


atanisoft:
complete test case showing the problem: https://github.com/atanisoft/esp32-broken-equals

balazsracz:
The files changed here look fine, but I need to go and review all occurrences where we are using boolean constant values to check whether it would be broken somewhere.

PR #317, opened 2019-08-28
Adding OPENMRN_FEATURE_REBOOT macro for reboot().

atanisoft:
This was being not called from ConfigUpdateFlow except when ```__FreeRTOS__``` was defined even though the weak function was defined everywhere except on Mac OS.


src/openlcb/MemoryConfig.hxx
```diff
@@ -46,7 +47,7 @@ class Notifiable;
 extern "C" {
 extern void enter_bootloader();
 /** @todo need to find an alternative for reboot() for MacOS */
-#if !defined (__MACH__)
```
balazsracz: there must be one more instance in this file

PR #318, opened 2019-08-31
Adds an option to TivaDCC to not perform the poweron sequence.

balazsracz:


src/freertos_drivers/ti/TivaDCC.hxx
```diff
@@ -274,7 +274,14 @@ public:
 
     /// Turns on DCC output.
     void enable_output() {
```
bakerstu: You could choose to fix the brace to a hanging brace.  Not technically part of the pull request.  Up to you.

balazsracz: i'll rather reformat the whole file in one go.

PR #320, opened 2019-09-01
Updating constants.hxx to fix issue on esp32

atanisoft:
Using the datatype as "void" and _sym_XXX as a function causes the xtensa GCC compiler to generate code which is unable to successfully evaluate config_XXX() == CONSTANT_TRUE as boolean true even when config_XXX() returns CONSTANT_TRUE when printed. By switching the datatype to int and treating it as a variable instead, the compiler generates the expected equality test.

The previous PR for this used constant values of 4 and 8 which also worked but simply changing the declaration of the symbol is a smaller change.

### DIFF
--- a/src/utils/constants.hxx
+++ b/src/utils/constants.hxx
@@ -94,7 +94,7 @@
 #else  // native C
 
 #define DECLARE_CONST(name)                                                    \
-    EXTERNC extern void _sym_##name(void);                                     \
+    EXTERNC extern int _sym_##name;                                            \
     EXTERNCEND typedef unsigned char                                           \
     _do_not_add_declare_and_default_const_to_the_same_file_for_##name;         \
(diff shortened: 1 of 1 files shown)


atanisoft:
@RobertPHeller you may need this PR as well on your local copy, the ESP32 has an issue with the true/false constants.

balazsracz:
I think this broke the constants feature for ARM targets (including the vast majority of MCUs). I don't quite know why, but I am getting problems in binaries cmpiled off recent OpenMRN master that point to the constants and overrides not working as they should be. Rolling back openmrn version fixes those problems.

atanisoft:
Hmm, I can't think of any reason it might break there. Perhaps we should add an explicit check for ESP32 and define the macros accordingly?

PR #321, opened 2019-09-15
Adding a simple service locator

JohnSL:


JohnSL:
I added another test that shows one difference between a service locator and a singleton. Namely, you can register an instance using an interface type. That means, depending on conditions, you could register different implementations. The consumer requests the instance based on the type of the interface, and therefore doesn't need to know the actual type.

This pattern also helps with testing. So you can register a mock instance easily and the consumer has no idea it's using a mock or a test double.

balazsracz:
Singletons also have test injection support and an interface can also be
inheriting of a Singleton, which provides the polymorphism feature you
mentioned.

--
Sent from my mobile device

src/utils/ServiceLocator.cxxtest
```diff
@@ -0,0 +1,70 @@
...
+    OtherImplementation1 first;
+    OtherImplementation2 second;
+
+    OtherInterface::register_service(&first);
+    OtherInterface *service = OtherInterface::get_service();
+    EXPECT_EQ("One", service->get_name());
+
+    OtherInterface::register_service(&second);
+    service = OtherInterface::get_service();
+    EXPECT_EQ("Two", service->get_name());
```
balazsracz: Please expand this test with the following statements:

1. Show after line 69 how to retrieve the implementation "first".
2. Then destruct second.
3. Then show that "first" can still be retrieved.

JohnSL:
As I've said many times, this is not the same as a singleton, and it can do things that you can't do with a singleton. I've rewritten the implementation to make that absolutely clear. The idea as I've implemented is to allow easier access to classes that are already effectively singletons, but are not available elsewhere. Plus, in the case where I've actually used it in the throttle (but not pushed a branch), I'm using it for a field that is within another class. This other class is really hard to create within the context of a test as it has so many dependencies. Without a service locator, it would take modifying a whole lot of code so I can pass this pointer along a long change of classes. And then you still have the issue above where the instance could be freed while the consumer still has a pointer to it.
But if you like, I can refactor this to use shared_ptr.

JohnSL:
OK, I've refactored the code yet again, this time using smart pointers instead of dumb pointers.

- Use-after-free isn't an issue anymore because of smart pointers
- It is absolutely not hostile to test environments. I added a `clear` method that tests can use to ensure they start with an empty registration. But in any event, tests should register a new instance for anything being used by the code under test (good test practice)
- Double registration? What do you mean by this? You can't register two instances with the same interface type. A new registration will overwrite a previous registration. Please see the tests
- Thread safety. Are singletons thread safe? If so, how?
- Arbitrary name. Are you referring to using aliases for registration? This is not something we need now, and is a more advanced, less often use, feature in _some_ implementations of service locators. All service locators allow for registration based just on type

PR #322, opened 2019-09-21
Add support in ServiceLocator for registering with an alias

JohnSL:


src/utils/ServiceLocator.cxxtest
```diff
@@ -82,3 +83,27 @@ TEST_F(ServiceLocatorTests, register_via_interface)
...
+    // Default still returns first
+    service = ServiceLocator<OtherInterface>::get_service();
+    ASSERT_NE(nullptr, service);
+    EXPECT_EQ("One", service->get_name());
+
+    // Alias returns second
+    service = ServiceLocator<OtherInterface>::get_service("other");
+    ASSERT_NE(nullptr, service);
+    EXPECT_EQ("Two", service->get_name());
+}
```
bakerstu: It is probably redundant given the get_name(), but should we have a test that the first service pointer != the second service pointer?

PR #323, opened 2019-09-29
Handle wlan_connect failures.

bakerstu:


src/freertos_drivers/common/WifiDefs.hxx
```diff
@@ -45,6 +45,12 @@ enum class CountryCode : uint8_t
     UNKNOWN, ///< unknown country code
 };
 
+enum WlanConnectResult
```
balazsracz: Since this is at the toplevel namespace, please either
- make this an `enum class` (like the above enums)
- prefix all enum values with WLAN_CONNECT_RESULT_

because right now we have a symbol called 'OK' poisoning the toplevel namespace which will cause problems.

src/freertos_drivers/common/WifiDefs.hxx
```diff
@@ -45,6 +45,12 @@ enum class CountryCode : uint8_t
...
+{
+    OK = 0, ///< success
```
balazsracz: consider `CONNECT_OK`. I'm worried that some badly written C headers (like in mingw or some embedded middlewares) might have a `#define` for OK.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -698,8 +699,15 @@ void CC32xxWiFi::wlan_connect(const char *ssid, const char* security_key,
...
     int result = sl_WlanConnect((signed char*)ssid, strlen(ssid), 0,
                                 &sec_params, 0);
-    HASSERT(result >= 0);
 
+    switch (result)
+    {
+        default:
+            SlCheckError(result);
+            return OK;
+        case SL_ERROR_WLAN_PASSWORD_ERROR:
```
balazsracz: is it possible that we can run into the same error with the wlan_profile_add?

PR #327, opened 2019-10-20
Add SocketCAN support to the hub application on the Raspberry Pi

kiwi64ajs:
Hi Guys,
Here's the changes I made to add SocketCAN support to the hub application, to build on the Raspberry Pi.
I've also included some build notes in the README.md file with links to the CAN Interface and Linux kernel driver build instructions

### DIFF
--- a/applications/hub/main.cxx
+++ b/applications/hub/main.cxx
@@ -226,7 +226,6 @@ int appl_main(int argc, char *argv[])
         bind(s, (struct sockaddr *)&addr, sizeof(addr));
 
         new HubDeviceSelect<CanHubFlow>(&can_hub0, s);
-//        connections.emplace_back(port);
     }
 
     while (1)
--- a/applications/hub/main.cxx
+++ b/applications/hub/main.cxx
@@ -37,6 +37,17 @@
 #include <stdio.h>
 #include <unistd.h>
 
+#if defined(__linux__) || defined(__MACH__)
+#include <net/if.h>
+#include <termios.h> /* tc* functions */
+#endif
+
+#if defined(__linux__)
+#include "utils/HubDeviceSelect.hxx"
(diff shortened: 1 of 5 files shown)


applications/hub/main.cxx
```diff
@@ -178,6 +195,40 @@ int appl_main(int argc, char *argv[])
...
+
+        ::ioctl(s, SIOCGIFINDEX, &ifr);
+
+        addr.can_family = AF_CAN;
+        addr.can_ifindex = ifr.ifr_ifindex;
+
+        bind(s, (struct sockaddr *)&addr, sizeof(addr));
+
+        new HubDeviceSelect<CanHubFlow>(&can_hub0, s);
+//        connections.emplace_back(port);
```
bakerstu: Remove dead code.

applications/hub/targets/linux.rpi1/AvaHiMDNS.cxx
```diff
@@ -0,0 +1,140 @@
+/** \copyright
```
bakerstu: This file looks like an exact copy of the one under targets/linux.x86.  Please make this a symbolic link instead.

atanisoft: Instead, it would be better to move this code to use os/MDNS.hxx/cxx I think since this code looks almost verbatim for both the linux.x86 and now linux.rpi1. I was contemplating making this update since I'm using this same hub app on my rPi 3B+ but the linux.x86 app compiled and works on there without alteration (I ran make from top level dir)

kiwi64ajs: Yes, I had to do a copy/paste of the code to get a clean compile. Making this a shared file would be good

PR #333, opened 2019-11-03
SHA256 helper for the CC32xx hardware and for linux using OpenSSL.

balazsracz:
Adds a helper library for computing SHA256 on the CC32xx hardware.
Adds linux based OpenSSL helper function for SHA256.
Adds test vectors of SHA256.
Updates test suite runners to also run the SHA256 tests as well (both on linux and C32xx).

### DIFF
--- a/bin/sha-test-vectors.awk
+++ b/bin/sha-test-vectors.awk
@@ -1,0 +1,20 @@
+#!/usr/bin/gawk -f
+
+# This file is used to preprocess the SHA test vectors file.
+# Usage: $0 < SHA256LongMsg.rsp
+
+BEGIN {
+  idx = 0;
+}
+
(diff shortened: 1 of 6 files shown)


atanisoft:
Should we consider a platform neutral interface with platform impls of it?

balazsracz:
> Should we consider a platform neutral interface with platform impls of it?

I think at this point we only have extremely specialized usecases. None of the OpenLCB protocols actually use these, conversely there are no example applications in openMRN either.
I'm adding it only for the purpose of some proprietary bootloaders to use it in their implementation.

So it might not be worth our time to write a general crypto library API.

atanisoft:
Fair enough, it can be something to consider in the future if it enters the general protocol needs

PR #334, opened 2019-11-03
Fix constants override feature for ARM processors.

balazsracz:
Fixes bug in #320 
GCC apparently optimized away some comparisons because it believed that an int typed variable must be aligned.

### DIFF
--- a/src/utils/constants.hxx
+++ b/src/utils/constants.hxx
@@ -94,9 +94,9 @@
 #else  // native C
 
 #define DECLARE_CONST(name)                                                    \
-    EXTERNC extern int _sym_##name;                                            \
+    EXTERNC extern char _sym_##name;                                           \
     EXTERNCEND typedef unsigned char                                           \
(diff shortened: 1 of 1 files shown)


src/utils/constants.hxx
```diff
@@ -94,9 +94,9 @@
 #else  // native C
 
 #define DECLARE_CONST(name)                                                    \
-    EXTERNC extern int _sym_##name;                                            \
+    EXTERNC extern char _sym_##name;                                           \
```
atanisoft: Are there any concerns with limiting to 8bit values vs 32bit from int? I'm not certain it will make a difference since we use ptrdiff_t when retrieving the values.

balazsracz: The value is the address of this symbol. That address is 32-bit in a 32-bit processor.

atanisoft: Ok that is what I was thinking but wasn't 100% sure on

PR #335, opened 2019-11-06
Bakerstu stm32f0 bare metal build

bakerstu:


bakerstu:
Please take a close look at the FreeRTOSConfig.h changes. I'm want these scrutinized, I made these when trying to get the application running for the first time.

include/freertos/FreeRTOSConfig.h
```diff
@@ -137,17 +137,38 @@ standard names - or at least those used in the unmodified vector table. */
...
+#define configCPU_CLOCK_HZ             ( cpu_clock_hz )
+#define configMINIMAL_STACK_SIZE       ( ( unsigned short ) 256 )
+#define configTOTAL_HEAP_SIZE          ( ( size_t ) ( 7000 ) )
+#define configTIMER_TASK_STACK_DEPTH   256
+
+
+#define configKERNEL_INTERRUPT_PRIORITY         255
+/* !!!! configMAX_SYSCALL_INTERRUPT_PRIORITY must not be set to zero !!!!
+See http://www.FreeRTOS.org/RTOS-Cortex-M3-M4.html. */
+#define configMAX_SYSCALL_INTERRUPT_PRIORITY     0xa0 /* equivalent to 191, or priority 5. */
```
balazsracz: is there such a thing for Cortex-M0?
I thought interrupt priorities are only there on M3/M4.

targets/bare.armv6m/freertos/sources
```diff
@@ -0,0 +1,14 @@
+VPATH := $(FREERTOSPATH)/Source: \
+         $(FREERTOSPATH)/Source/portable/GCC/ARM_CM3: \
```
balazsracz: this is odd; why do we have a freertos directory in the bare.armv6m target?

include/freertos/FreeRTOSConfig.h
bakerstu: The Cortex-M0 does have interrupt priorities, however, interrupts can only be globally disabled, not by priority. I removed these two defines, they are unnecessary.

targets/bare.armv6m/freertos/sources
bakerstu: I agree that it is odd. We have it for the bare.armv7m target as well. That is where I copied it from.

PR #336, opened 2019-11-24
Adds support for inverted banks in CC3220SF SPIFFS.

balazsracz:
Adds support for Flash Mirrored Mode to the 3220SF SPIFFS driver. In mirrored mode the upper and lower banks are reversed in address space of the MCU; however, when talking to the flash controller, the original addresses need to be used.
This PR adds code to automatically decide which of the flash mode we are using, assuming that by the time of the static initializers are run the flash controller is reconfigured.

Adds some debug options and workarounds:
- can enable INFO logging of flash operations
- adds custom locking around flash operations, with a few options on what could be used as a lock (e.g. global interrupt disable).
- Adds selectable ROM or Flash based addresses for the actual flash write/erase API calls.

### DIFF
--- a/src/freertos_drivers/spiffs/cc32x0sf/CC32x0SFSPIFFS.cxx
+++ b/src/freertos_drivers/spiffs/cc32x0sf/CC32x0SFSPIFFS.cxx
@@ -61,13 +61,14 @@ constexpr unsigned minpri = 0x40;
 #define DI() 
 #define EI() 
 
+static unsigned addr_mirror = (HWREG(FLASH_CONF) & FCMME) ? 0x80000 : 0;
 
 #if 1
-#define FPG ROM_FlashProgram
-#define FER ROM_FlashErase
+#define FPG(data, addr, size) ROM_FlashProgram(data, (addr) ^ addr_mirror, size)
+#define FER(addr) ROM_FlashErase((addr) ^ addr_mirror)
 #else
-#define FPG FlashProgram
-#define FER FlashErase
+#define FPG(data, addr, size) FlashProgram(data, (addr) ^ addr_mirror, size)
+#define FER(addr) FlashErase((addr) ^ addr_mirror)
 #endif
 
 //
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/spiffs/cc32x0sf/CC32x0SFSPIFFS.cxx
```diff
@@ -31,11 +31,64 @@
...
+// #define EI() CPUbasepriSet(ppri);
+
+// No write locking.
+#define DI()
+#define EI()
+
+// This ifdef decides whether we use the ROM or the flash based implementations
+// for Flash write and erase. It also supports correcting for the reversed bank
+// addresses.
+#if 1
```
bakerstu: I presume you have hard coded the use of the ROM API to be enabled. Is the #else dead code then?

src/freertos_drivers/spiffs/cc32x0sf/CC32x0SFSPIFFS.cxx
```diff
@@ -31,11 +31,64 @@
...
+/// This bit in the FLASH_CONF register means that the banks are reversed by
+/// their address mapping.
+#define FCMME 0x40000000
+/// This value defines up to one bit that needs to be XOR-ed to the flash
+/// address before calling the flash APIs to cover for reversed flash banks.
+static const unsigned addr_mirror = (HWREG(FLASH_CONF) & FCMME) ? 0x80000 : 0;
+
+// Different options for what to set for flash write locking.
+
+// Global disable interrupts.
```
bakerstu: How is the option chosen at an application specific level? Would it be better to define these as weak functions that the application can override?

src/freertos_drivers/spiffs/cc32x0sf/CC32x0SFSPIFFS.cxx
balazsracz: We will not use the other code in production, but flipping these ifdefs around is helpful to debug and reproduce crashes around the flash.

src/freertos_drivers/spiffs/cc32x0sf/CC32x0SFSPIFFS.cxx
balazsracz: It is not expected that the application will have to choose, but these are alternatives for debugging only.

Issue #212, opened 2019-01-30
esp32: task priorities


atanisoft:
Currently the default task level is set to ESP_TASK_TCPIP_PRIO (https://github.com/espressif/esp-idf/blob/master/components/esp32/include/esp_task.h#L53) and this is leveraged for HW CAN RX and TX tasks. We need to explore if we can lower the priorities of these tasks and if so, what makes sense on the ESP32. We need to balance performance and usability.

balazsracz:
We are no longer reliant on separate tasks for TX and RX of CAN packets. The CAN / TWAI driver supports both select-based and async processing APIs.

PR #339, opened 2019-12-02
Lower default priority for OpenMRN tasks on ESP32

atanisoft:
Lower the default priority of OpenMRN tasks on the ESP32 by one, this is to ensure that none of the OpenMRN tasks block the TCP/IP task from timely execution. Further refinements/documentation of task priorities will be done in a subsequent PR.

### DIFF
--- a/arduino/OpenMRNLite.h
+++ b/arduino/OpenMRNLite.h
@@ -57,11 +57,13 @@ namespace openmrn_arduino {
 /// Default stack size to use for all OpenMRN tasks on the ESP32 platform.
 constexpr uint32_t OPENMRN_STACK_SIZE = 4096L;
 
-/// Default thread priority for any OpenMRN owned tasks on the ESP32
-/// platform. ESP32 hardware CAN RX and TX tasks run at lower priority
-/// (-1 and -2 respectively) of this default priority to ensure timely
-/// consumption of CAN frames from the hardware driver.
-constexpr UBaseType_t OPENMRN_TASK_PRIORITY = ESP_TASK_TCPIP_PRIO;
+/// Default thread priority for any OpenMRN owned tasks on the ESP32 platform.
+/// ESP32 hardware CAN RX and TX tasks run at lower priority (-1 and -2 
+/// respectively) of this default priority to ensure timely consumption of CAN
+/// frames from the hardware driver.
+/// Note: This defaults to one priority level lower than the TCP/IP task uses
+/// on the ESP32.
+constexpr UBaseType_t OPENMRN_TASK_PRIORITY = ESP_TASK_TCPIP_PRIO - 1;
 
 } // namespace openmrn_arduino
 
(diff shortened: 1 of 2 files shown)


arduino/OpenMRNLite.h
```diff
@@ -57,11 +57,13 @@ namespace openmrn_arduino {
...
-/// Default thread priority for any OpenMRN owned tasks on the ESP32
-/// platform. ESP32 hardware CAN RX and TX tasks run at lower priority
-/// (-1 and -2 respectively) of this default priority to ensure timely
-/// consumption of CAN frames from the hardware driver.
-constexpr UBaseType_t OPENMRN_TASK_PRIORITY = ESP_TASK_TCPIP_PRIO;
+/// Default thread priority for any OpenMRN owned tasks on the ESP32 platform.
+/// ESP32 hardware CAN RX and TX tasks run at lower priority (-1 and -2 
+/// respectively) of this default priority to ensure timely consumption of CAN
+/// frames from the hardware driver.
+/// Note: This defaults to one priority level lower than the TCP/IP task uses
```
balazsracz: don't say "defaults" because it is not adjustable.

arduino/OpenMRNLite.h
```diff
@@ -57,11 +57,13 @@ namespace openmrn_arduino {
 /// Default stack size to use for all OpenMRN tasks on the ESP32 platform.
 constexpr uint32_t OPENMRN_STACK_SIZE = 4096L;
 
-/// Default thread priority for any OpenMRN owned tasks on the ESP32
-/// platform. ESP32 hardware CAN RX and TX tasks run at lower priority
-/// (-1 and -2 respectively) of this default priority to ensure timely
-/// consumption of CAN frames from the hardware driver.
-constexpr UBaseType_t OPENMRN_TASK_PRIORITY = ESP_TASK_TCPIP_PRIO;
+/// Default thread priority for any OpenMRN owned tasks on the ESP32 platform.
+/// ESP32 hardware CAN RX and TX tasks run at lower priority (-1 and -2
```
balazsracz: this is not true anymore.

atanisoft: That is true, the CAN tasks will need to be adjusted as part of this PR. Will add that.

arduino/OpenMRNLite.h
atanisoft: Will update the text to make it clear that it sets the priority rather than defaults it.

atanisoft:
> thanks! I'm looking forward to a diagram of communication/dataflow and priorities :)

Yes, an update to the docs will be needed to clearly outline this, I opened https://github.com/bakerstu/openmrn/issues/212 to track this but haven't had a chance to address it properly. I'll start on the documentation for the ESP32 as one of my next tasks (after syncing up the Esp32WiFiManager code with fixes for Station+SoftAP)

PR #344, opened 2019-12-26
Adds forward allocator to optimize many small memory allocations

balazsracz:
Adds a unidirectional allocator, optimized for allocating many small objects with no overhead, then freeing them all together.

nit: fixed leaking all bufferpool memory upon teardown.
tested: ran unit test under valgrind. No leaks found.

### DIFF
--- a/src/utils/ForwardAllocator.hxx
+++ b/src/utils/ForwardAllocator.hxx
@@ -64,7 +64,7 @@ public:
 
     /// Usable only from unittests. Re-creates the pool to be empty.
     static void TEST_recreate_pool();
-    
+
 private:
     /// A primitive type that has sufficient alignment to what we support.
     typedef uint64_t primitive_t;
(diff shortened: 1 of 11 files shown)


src/utils/ForwardAllocator.hxx
```diff
@@ -0,0 +1,104 @@
...
+
+    /// Allocates a block of memory. Thread-safe.
+    /// @param size the number of bytes to allocate.
+    /// @param align alignment of allocated object in bytes (1 to 8).
+    void *allocate(size_t size, size_t align);
+
+    ForwardAllocator();
+    ~ForwardAllocator();
+
+    /// Usable only from unittests. Re-creates the pool to be empty.
```
bakerstu: Should we wrap in a #if defined (GTEST)?

balazsracz: done.

PR #346, opened 2019-12-29
Fix race conditions found by thread sanitizer in an example unittest

balazsracz:
This PR fixes a number of multi-threading issues, some of which are test-only, while other are serious bugs impacting any application that runs in more than one thread/executor:

- Adds the ability to run the 'cov' target using ThreadSanitizer mode and disabling the coverage reporting. (Coverage reporting creates a lot of false positives for thread sanitizer.)
- Fixes a reported data race in the shutdown() method of the Executor by using atomic integer. This is typically invoked in tests only.
- Fixes a reported data race in the ActivetTimers::notify() method by using atomic integer and cmpxchg.
- Fixes a major bug in the reference counting of buffers by using atomic integers and cmpxchg.
- Fixes a global destructor ordering bug that probably caused the segfaults upon program exit under linux. The issue was that the LinkedObject utility class uses a statically allocated mutex, which, upon static destruction, was deleted. Depending on the destruction order, there were still stateflows that were trying to access this mutex.
- Fixes incorrect ordering of lock and test in Q::insert. This was a race condition but probably harmless.
- Fixes a test destructor bug that accessed the timer object in the wrong thread.
- Fixes a bunch of race conditions relating to the ordering of entering expectations in a unit test.

### DIFF
--- a/src/utils/LinkedObject.hxx
+++ b/src/utils/LinkedObject.hxx
@@ -64,7 +64,7 @@ public:
     }
 
 private:
-    std::atomic_uint_fast8_t isInitialized_;
+    std::atomic_uint_least8_t isInitialized_;
     uninitialized<Atomic> atomic_;
 };
 
--- a/src/utils/LinkedObject.hxx
+++ b/src/utils/LinkedObject.hxx
@@ -36,7 +36,37 @@
 #ifndef _UTILS_LINKEDOBJECT_HXX_
 #define _UTILS_LINKEDOBJECT_HXX_
 
+#include <atomic>
+
 #include "utils/Atomic.hxx"
+#include "utils/Uninitialized.hxx"
+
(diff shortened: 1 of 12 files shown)


src/utils/LinkedObject.hxx
```diff
@@ -36,15 +36,45 @@
...
 
+#include <atomic>
+
 #include "utils/Atomic.hxx"
+#include "utils/Uninitialized.hxx"
+
+/// This object encapsulates an Atomic, which never gets destroyed. It is
+/// intended to be used only as static object, because we depend on zero
+/// initialization at construction time.
+class CreateOnlyAtomic
```
bakerstu: Should this be a template class for the atomic type?

balazsracz: actually this is an `utils/Atomic.hxx` (of which there is only one), not an std::atomic (which is a template).

PR #347, opened 2020-01-04
Update Doxyfile

JohnFFlanagan:
DOXYGEN version 1.8.16 produces no output with the FILE_PATTERNS default.  This is a know DOXYGEN issue and is fixed in the leading edge code.  This code is not available as a general release.

1.8.13 is the default Ubuntu release as well as others.
1.8.16 is the default MAC OSX release.  Which is where I initially installed openmrn.

This is my first PR.  Please let me know of anything I should have done differently.

John


JohnFFlanagan:
I noticed that DOXYGEN version 1.8.17 was released on Dec 27.  I downloaded and tested that.  It works with both the FILE_PATTERN commented out and no commented out.  So only 1.8.16 fails.

PR #350, opened 2020-01-31
Handle delay in initialization of FreeRTOSTCP task

smcharg:
Fixes an issue where calls to FreeRTOS_socket would fail should the FreeRTOSTCP task not have completed initialization.

Adds another location for finding the FreeRTOSTCP source to path.mk

### DIFF
--- a/etc/path.mk
+++ b/etc/path.mk
@@ -220,6 +220,7 @@ SEARCHPATH := \
   /opt/FreeRTOSPlus/TCP \
   /opt/FreeRTOSPlus/default/TCP \
   /opt/FreeRTOS/plus-tcp \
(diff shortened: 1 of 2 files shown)


balazsracz:
I found that the priority of the net_task is 4 and the priority of the FreeRTOS_IP task is also 4. I'm wondering if we have a task priority inversion problem here. Should our network task have the top priority?
Reducing the priority of net_task to configMAX_PRIORITIES - 2 seems to fix this problem, because then the net_task will not even start until the FreeRTOS_IP task has run to its first blocking state, at which point it will declare that it is initialized == true.
WDYT?

balazsracz:
@smcharg for the last question.

PR #351, opened 2020-02-02
Openmrnlite missing newline

RobertPHeller:
This just adds the missing but needed NUL (EOS) in OpenMRNLite.h.  The spec for CDI strings requires this byte.


balazsracz:
This looks fine, but please remove your signature from the line you added. Git will keep track of that.

PR #357, opened 2020-02-26
Bakerstu 74hc595goi driver

bakerstu:


src/freertos_drivers/common/SN74HC595GPO.hxx
```diff
@@ -0,0 +1,242 @@
...
+        for ( ; /* forever */ ; )
+        {
+            sem_.wait();
+            if (ioPending_)
+            {
+                ioPending_ = false;
+                for (unsigned i = 0; i < N; ++i)
+                {
+                    uint8_t data = gpoData_[0];
+                    ::write(spiFd_, &data, 1);
```
balazsracz: why aren't we writing all bytes in one go?
Specifically, this will not work for more than 1 chip, because the latch gets toggled inbetween, the mutex gets released and other SPI transaction can happen, sending garbage to the remaining shift registers.

src/freertos_drivers/common/SN74HC595GPO.hxx
```diff
@@ -0,0 +1,242 @@
...
+
+/// General Purpose Output (GPO) instance on the 74HC595.
+/// @tparam N number of chips on the bus, starting at 1
+template <uint8_t N> class SN74HC595GPO : public Gpio
+{
+public:
+    /// Constructor.
+    /// @param instance parrent 74HC595 instance that "owns" the interface.
+    /// @param chip_index index on the bus for the chip, starting at 0.
+    /// @param bit bit index (0 through 1) of the output.
```
balazsracz: 0 through 7

bakerstu: Fixed.

src/freertos_drivers/common/SN74HC595GPO.hxx
bakerstu: Fixed.

src/freertos_drivers/common/SPI.cxx
```diff
@@ -40,6 +40,49 @@
...
+            csDeassert();
+            bus_unlock();
+            lock_.unlock();
+            return result;
+        }
+        if (msgs->cs_change)
+        {
+            if (UNLIKELY(msgs->delay_usec))
+            {
+                usleep(msgs->delay_usec);
```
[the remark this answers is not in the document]
bakerstu: No, this is the correct order, per the Linux reference that this contract was modeled after.

PR #368, opened 2020-03-17
gc_format fixes

balazsracz:
- Adds support for parsing an entire gridconnect packet including leading : and trailing ;.
- Fixes constant override for generate newlines.


src/utils/gc_format.cxx
```diff
@@ -242,7 +247,7 @@ char* gc_format_generate(const struct can_frame* can_frame, char* buf, int doubl
         output(buf, nibble_to_ascii(can_frame->data[offset] & 0xf));
     }
     output(buf, ';');
-    if (config_gc_generate_newlines()) {
+    if (config_gc_generate_newlines() == CONSTANT_TRUE) {
```
atanisoft: this line should have the { moved down one

PR #370, opened 2020-03-21
Bakerstu stm32f0 hse value

bakerstu:


balazsracz:
could you check if there are any other HwInit.cxx that need to be updated?

bakerstu:
I'm pretty sure I got the only HwiInit.cxx. It appears to be linked in the necessary spots.

PR #373, opened 2020-03-21
STM32 Arduino fixes.

balazsracz:
- Upgrade Arduino API to be build compatible with ESP32 1.0.4 and STM32 1.8.0.
- Fixes breakages of STM32.
- Updates Serial bridge to be compatible with current HardwareSerial API for both STM32 and ESP32.
- blacklists some modules that do not compile under single-threaded environment.
- adds STM32Can to be exported via libify.
- Adds a guard around CpuLoad to be only compiled under freertos (esp-arduino or openmrn native), and Stm32Can to only be compiled for openmrn-native or arduino-stm32.

### DIFF
--- a/arduino/examples/ESP32SerialBridge/ESP32SerialBridge.ino
+++ b/arduino/examples/ESP32SerialBridge/ESP32SerialBridge.ino
@@ -161,7 +161,7 @@ void setup() {
 #endif // PRINT_PACKETS
 
     // Add Serial1 as a bridge
-    openmrn.add_gridconnect_port(new Esp32HardwareSerialAdapter(Serial1));
+    openmrn.add_gridconnect_port(&Serial1);
 
     // Add the hardware CAN device as a bridge
     openmrn.add_can_port(
--- a/arduino/libify.sh
+++ b/arduino/libify.sh
@@ -169,6 +169,9 @@ copy_file src/freertos_drivers/arduino \
 copy_file src/freertos_drivers/esp32 \
           src/freertos_drivers/esp32/*
 
+copy_file src/freertos_drivers/stm32 \
+          src/freertos_drivers/st/Stm32Can.*
+
 copy_file src/os src/os/*.h src/os/*.c src/os/*.hxx \
           src/os/{OSImpl,MDNS,OSSelectWakeup}.cxx
 
--- a/arduino/libify.sh
+++ b/arduino/libify.sh
@@ -156,8 +156,6 @@ copy_file src/openlcb src/openlcb/*.hxx src/openlcb/*.cxx
 rm -f ${TARGET_LIB_DIR}/src/openlcb/CompileCdiMain.cxx \
     ${TARGET_LIB_DIR}/src/openlcb/Stream.cxx \
     ${TARGET_LIB_DIR}/src/openlcb/Stream.hxx \
-    ${TARGET_LIB_DIR}/src/openlcb/IfTcp.cxx \
-    ${TARGET_LIB_DIR}/src/utils/HubDeviceSelect.cxx \
 
 copy_file src/freertos_drivers/arduino \
           src/freertos_drivers/common/DeviceBuffer.{hxx,cxx} \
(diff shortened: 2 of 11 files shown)


arduino/examples/ESP32SerialBridge/ESP32SerialBridge.ino
```diff
@@ -161,7 +161,7 @@ void setup() {
 #endif // PRINT_PACKETS
 
     // Add Serial1 as a bridge
-    openmrn.add_gridconnect_port(new Esp32HardwareSerialAdapter(Serial1));
+    openmrn.add_gridconnect_port(&Serial1);
```
atanisoft: since this example no longer has hard dependencies on ESP32 should it be renamed to simply ArduinoSerialBridge?

arduino/libify.sh
```diff
@@ -155,7 +155,9 @@ copy_file src/openlcb src/openlcb/*.hxx src/openlcb/*.cxx
 
 rm -f ${TARGET_LIB_DIR}/src/openlcb/CompileCdiMain.cxx \
     ${TARGET_LIB_DIR}/src/openlcb/Stream.cxx \
-    ${TARGET_LIB_DIR}/src/openlcb/Stream.hxx
+    ${TARGET_LIB_DIR}/src/openlcb/Stream.hxx \
+    ${TARGET_LIB_DIR}/src/openlcb/IfTcp.cxx \
+    ${TARGET_LIB_DIR}/src/utils/HubDeviceSelect.cxx \
```
atanisoft: perhaps instead of excluding the cxx we can add guards to prevent compilation on the STM32?

src/os/os.c
```diff
@@ -918,7 +918,7 @@ int ignore_fn(void)
     return 0;
 }
 
-#if !defined(ESP32)
+#if !defined(ARDUINO)
```
atanisoft: Can we go with:
```#if !defined(ARUDINO) && !defined(ESP32)```
Since I no longer have arduino defined in my CS build this may break when I next sync.

File: arduino/examples/ESP32SerialBridge/ESP32SerialBridge.ino (hunk 1 above)
Comment (balazsracz) [reply arduino/examples/ESP32SerialBridge/ESP32SerialBridge.ino]: why would this be not specific to the ESP32?
- it has the ESP32 CAN
- it is using the fd-based configuration system that ESP32 has
- it is using ESP32 GPIO numbers and assignments.

Comment (atanisoft) [reply arduino/examples/ESP32SerialBridge/ESP32SerialBridge.ino]: Ahh, forgot that it had the can GPIO pins and the SPIFFS VFS paths. Scratch that suggestion on moving to generic. Is there a need for a similar example for STM32?

File: arduino/libify.sh (hunk 2 above)
Comment (balazsracz) [reply arduino/libify.sh]: done

Comment (atanisoft) [reply arduino/libify.sh]: Thanks!

File: src/os/os.c (hunk 3 above)
Comment (balazsracz) [reply src/os/os.c]: done

File: arduino/examples/ESP32SerialBridge/ESP32SerialBridge.ino (hunk 1 above)
Comment (balazsracz) [reply arduino/examples/ESP32SerialBridge/ESP32SerialBridge.ino]: yes i'm going to make some stm32 examples.

File: src/os/os.c (hunk 3 above)
Comment (atanisoft) [reply src/os/os.c]: Thanks

File: arduino/libify.sh [hunk 4]
```diff
@@ -155,7 +155,7 @@ copy_file src/openlcb src/openlcb/*.hxx src/openlcb/*.cxx
 
 rm -f ${TARGET_LIB_DIR}/src/openlcb/CompileCdiMain.cxx \
     ${TARGET_LIB_DIR}/src/openlcb/Stream.cxx \
-    ${TARGET_LIB_DIR}/src/openlcb/Stream.hxx
+    ${TARGET_LIB_DIR}/src/openlcb/Stream.hxx \
```
atanisoft: remove trailing \

balazsracz: done

PR #374, opened 2020-03-22
Arduino-stm32 example sketches for CAN-Serial

balazsracz:
Adds two example sketches that do CAN-Serial conversion for STM32 under arduino:
- one basic implementation that is just a hub
- one advanced implementation that is an OpenLCB node (albeit without writeable CDI).

Tested with Nucleo F303 and F091 on the openlcb DevKit.


arduino/examples/Stm32CanSerial/Stm32CanSerial.ino
```diff
@@ -0,0 +1,83 @@
...
+ * Example Arduino sketch for the Stm32 showing how to make a Serial-CAN
+ * adapter with minimal size and dependencies.
+ *
+ * @author Balazs Racz
+ * @date 22 March 2020
+ */
+
+#include <Arduino.h>
+#include <OpenMRNLite.h>
+#include "freertos_drivers/stm32/Stm32Can.hxx"
```
atanisoft: Should this include be automatically pulled in as part of OpenMRNLite.h like on the ESP32?

arduino/examples/Stm32CanSerial/Stm32CanSerial.ino
```diff
@@ -0,0 +1,83 @@
...
+
+/// Specifies which Stm32 Pin should be used for CAN_RX signal. The default
+/// matches the pinout of the OpenLCB Dev Kit daughterboard.
+#define CAN_RX_PIN PB_8
+
+Stm32Can Can("/dev/can0");
+
+OVERRIDE_CONST_TRUE(gc_generate_newlines);
+
+/// These objects perform the crossbar siwtching of CAN frames.
```
atanisoft: typo in comment: siwtching -> switching

src/os/OS.hxx
```diff
@@ -741,6 +741,7 @@ extern unsigned critical_nesting;
 extern uint32_t SystemCoreClock;
 }
 #define cm3_cpu_clock_hz SystemCoreClock
+#define cpu_clock_hz SystemCoreClock
```
atanisoft: should these defines and the above extern block be below the !defined(ESP32) below? I don't see any usages of these in the ESP32 code paths. There likely isn't harm in having them here.

arduino/examples/Stm32CanSerial/Stm32CanSerial.ino
balazsracz: Good idea, done.

arduino/examples/Stm32CanSerial/Stm32CanSerial.ino
balazsracz: done

src/os/OS.hxx
balazsracz: done

Issue #256, opened 2019-03-15
add const for listener_stack_size in socket_listener.cxx


atanisoft:
Currently this is defined in socket_listener.cxx and needs to be increased for the ESP32 to work, moving it to DECLARE_CONST / OVERRIDE_CONST will be a cleaner long term implementation.

PR #377, opened 2020-03-23
[WIP] ESP32: ESP-IDF v4.x support

atanisoft:
This PR enables compilation of OpenMRN/OpenMRNLite in an ESP-IDF v4.x env. This will also include ESP32-S2 support (new single core modules to replace ESP8266).

Also fixes https://github.com/bakerstu/openmrn/issues/256.


atanisoft:
IDF v4 uses cmake to build, I have a CMakeFiles.txt that I can include that will build the required files that show up in OpenMRNLite with bare IDF and IDF+Arduino. I wasn't sure how we would want to package that though.

atanisoft:
@balazsracz I've tagged it as WIP in the title for now, I don't expect any additional work on this outside of the HW Can driver and ESP-IDF example(s) which I don't think would fit under the Arduino directory. Where should I create the ESP-IDF example app, perhaps under applications?

src/freertos_drivers/common/DeviceBuffer.cxx
```diff
@@ -34,7 +34,8 @@
 
 #include "DeviceBuffer.hxx"
 
-#ifndef ARDUINO
+#include "openmrn_features.h"
+#if OPENMRN_FEATURE_DEVTAB
```
balazsracz: ifdef ?

atanisoft: It could be done either way, I modeled this one after other feature flag usages since they are bool in nature. If you prefer ifdef I'll change it over.

My only concern on this change is for the STM32 Arduino port since it should also not include this functionality.

atanisoft: @balazsracz looking at the usage of FEATURE_* in the codebase there seems to be a mix of #if and #ifdef. Which would you prefer, I can update this one or send in a PR with mass cleanup for consistency of others.

PR #380, opened 2020-04-11
Fixes a major priority computation bug in stm32 uart driver.

balazsracz:
KERNEL_INTERRUPT_PRIORITY is 255, so K_I_P + 0x20 is not a valid interrupt priority.

### DIFF
--- a/src/freertos_drivers/st/Stm32Uart.cxx
+++ b/src/freertos_drivers/st/Stm32Uart.cxx
@@ -151,8 +151,8 @@ Stm32Uart::Stm32Uart(const char *name, USART_TypeDef *base, IRQn_Type interrupt)
 #if defined(GCC_ARMCM0)    
     HAL_NVIC_SetPriority(interrupt, 3, 0);
 #elif defined(GCC_ARMCM3)    
-    // Below kernel interrupt priority.
-    NVIC_SetPriority(interrupt, configKERNEL_INTERRUPT_PRIORITY + 0x20);
+    // Below kernel-compatible interrupt priority.
+    NVIC_SetPriority(interrupt, configMAX_SYSCALL_INTERRUPT_PRIORITY + 0x20);
 #else
 #error not defined how to set interrupt priority
 #endif    


szabodabo:
I've been looking into some of the UART code because some of my test devices consistently drop incoming data over serial due to overrun errors. Including this commit made my test F103 device crash on startup with a stack like
```
#0  diewith() at HwInit.cxx:143
#1  vPortValidateInterruptPriority () at /opt/FreeRTOS/Source/portable/GCC/ARM_CM3/port.c:677
...
#6  DeviceBufferBase::signal_condition_from_isr ()
    at DeviceBuffer.hxx:77
#7  Stm32Uart::interrupt_handler (<uart0>) at freertos_drivers/st/Stm32Uart.cxx:298
#8  <signal handler called>
```
and debugging the assertion that's failing says that the UART interrupt priority is 0, which is higher (lower in number) than is allowed because of re-entry issues.

At least in the F1 CMSIS drivers, `NVIC_SetPriority(uint32_t priority)` does this:
```
NVIC->IP[((uint32_t)(int32_t)IRQn)] = (uint8_t)((priority << (8U - __NVIC_PRIO_BITS)) & (uint32_t)0xFFUL);
```
which means it's expecting the 0-15 priority level, as the call site shows above for `GCC_ARMCM0`.

So then, originally, `NVIC_SetPriority(255 + 0x20)` was really calculating an 8-bit (as specified by FreeRTOS, and at least on an F103 test device I have) priority of 0x11110000, which just so happened to be the same as calling `NVIC_SetPriority(15)`.

But then changing this to `configMAX_SYSCALL_INTERRUPT_PRIORITY + 0x20` meant that the input parameter became 0xa0 + 0x20 --> 0xc0 --> 0x11000000, which becomes 0.

I patched this locally to
```
#define TO_NVIC_PRIORITY_LEVEL(uint8t_priority) uint8t_priority >> 4
NVIC_SetPriority(interrupt, TO_NVIC_PRIORITY_LEVEL(configMAX_SYSCALL_INTERRUPT_PRIORITY));
```
and it stopped crashing. It's hacky, though, and I suspect you want a cleaner solution. I don't know what to suggest because I don't know the extent of how FreeRTOS's interrupts are configured in openmrn.

Unfortunately, though, raising the effective priority of the USART interrupt didn't help my overrun issue.

balazsracz:
It looks like we are using NVIC_SetPriority incorrectly. This is probably true all across the entire codebase.
Honestly that function has a really stupid interface, because the values you can pass to the function depend on the specific hardware implementation. This makes writing portable code really hard. The processor IP is a lot more sensible,  which has left-aligned values in 8 bit wide registers. FreeRTOS uses this standard in the configMAX_SYSCALL_INTERRUPT_PRIORITY configuration setting and the TI middleware uses this format as well.

The right fix is that we need to discover every single usage of NVIC_SetPriority and fix them. We would probably be wise to just avoid using this function at all.

PR #384, opened 2020-04-16
Nucleo IO PortD Snap and Port C Tortoise Updates

rilull:
In the Nucleo IO app, we uncovered a code issue with the Tortoise drivers on port C where the existing code referenced them during the Port D setup, causing line 1 to 4 on Port C to not function as needed to drive tortoise machines. I promised on the list I would submit a fix and am finally getting around to it.

During additional testing, I found that having PortD setup to be used as a snap switch style output (pulsed) caused an LED that was connected to flash only very briefly. In other words, for a LED that is in a wayside signal, we need an output to come on and stay on until such a time that we would turn it off. I utilized a pre-processor define in the same style as the IO expander board to enable or disable the snap switch setup as it appears that there isn't an easy/quick way to set it from one type to another once the node is running.

I made changes to main.cxx and config.hxx. Main.cxx has the bulk of the changes with my inline comments. Config.hxx only got the define, matching what was done for the expansion board.


applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/main.cxx
```diff
@@ -482,14 +499,16 @@ openlcb::RefreshLoop loopab(stack.node(),
         producer_b3.polling(), producer_b4.polling(), //
         producer_b5.polling(), producer_b6.polling(), //
         producer_b7.polling(), producer_b8.polling(), //
-        &turnout_pulse_consumer_1,                    //
+#if PORTD_SNAP > 0
```
balazsracz, line 502: `#ifdef`

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/main.cxx
```diff
@@ -337,6 +337,21 @@ constexpr const Gpio *const kPortDEGpio[] = {
 openlcb::MultiConfiguredConsumer portde_consumers(stack.node(), kPortDEGpio,
```
balazsracz, line 337: This would need to be updated to not instantiate the port D consumers if there are portD snap switches operating.
That needs two changes:
- the kPortDEGpio needs to be shorter (fewer entries) based on ifdefs
- the number of repetitions defined in the config.hxx needs to change based on the ifdef.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/main.cxx
```diff
@@ -337,6 +337,21 @@ constexpr const Gpio *const kPortDEGpio[] = {
...
     ARRAYSIZE(kPortDEGpio), cfg.seg().portde_consumers());
 
+// Rick Lull - April 15 2020
+// Snap switches and LED lights conflict on same port. When GPIO pin has
+// snap configuration in place, LED will quickly flash on consumer event recv and
+// not stay on as desired/needed for signal driver.
+// When PORTD_SNAP is set to 1 (enabled), we will set portD to be used for snap
```
balazsracz, line 344: Please move this documentation to the config.hxx where the ifdef is for PORTD_SNAP

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/main.cxx
rilull: There are 12 snap switch entries, which suggest that we would use PortD.0-7 and PortE.0-3 for snaps, when enabled, while PortE.4-7 is always standard constant on outputs. I've tweaked this to hopefully give us the consumers for "standard" use and not give us those consumers when we are using snaps.
Please review that code and let me know your thoughts.

rilull: In an offline conversation, we have settled on 8 Snap outputs, using PortD. PortE will be standard, non-pulsed, outputs. The fixes for this have been pushed to my master fork and included in this pull request.

rilull:
F303 fully tested. F091 undergoing testing currently.

rilull:
Full testing on F091 and F303 has been completed.

PR #387, opened 2020-04-26
Update API for controlling DCC outputs

balazsracz:
Builds an explicit API for controlling DCC outputs of a command station or booster:
- allows more than one output to be present in one application (e.g. main vs program track output, or dual-output boosters)
- makes the railcom cutout configuration have an explicit API
- makes each output have multiple veto bits, each of which can cause the output to be disabled (e.g. short, user via the network, thermal overload, etc). This allows different parts of the application business logic to not need to be aware of each other.
- updates the TivaDCC driver to be able to support 3 independent outputs.
- Removes the POWER_TURNON feature from TivaDCC. It does not work well for certain DCC decoders.


src/freertos_drivers/common/DccOutput.hxx
```diff
@@ -0,0 +1,364 @@
...
+        SHORTED = 8,
+        /// The system is in thermal shutdown.
+        THERMAL = 16,
+        /// This output should be off due to the conflict between program track
+        /// and normal operation mode.
+        PGM_TRACK_LOCKOUT = 32,
+    };
+
+    /// Disables the output, marking in a bitmask why.
+    virtual void disable_output_for_reason(DisableReason bit) = 0;
```
bakerstu, line 80: Add a line break (space)

atanisoft:
would it make sense to surface the basic API at a higher level than freertos_drivers/common? Perhaps under dcc or similar? I've implemented a similar structure in ESP32CS recently (splitting OPS and PROG) and can easily shift over to this API.

src/freertos_drivers/common/DccOutput.hxx
```diff
@@ -0,0 +1,364 @@
...
+        else
+        {
+            clear_disable_output_for_reason(bit);
+        }
+    }
+
+    /// @return Bitmask of all currently set disable reasons.
+    virtual uint8_t get_disable_output_reasons() = 0;
+
+    enum class RailcomCutout
```
atanisoft, line 104: can you add docs to this enum?

src/freertos_drivers/common/DccOutput.hxx
```diff
@@ -0,0 +1,364 @@
...
+    /// Private constructor. These objects cannot be initialized and must only
+    /// have static members.
+    DccOutputHw();
+};
+
+template <int N>
+std::atomic_uint8_t DccOutputHw<N>::outputDisableReasons_ {
+    (uint8_t)DccOutput::DisableReason::INITIALIZATION_PENDING};
+template <int N>
+std::atomic_uint8_t DccOutputHw<N>::isRailcomCutoutEnabled_ {2};
```
atanisoft, line 234: any reason not to use the enum const here?

balazsracz:
> This seems reasonable. Do we have the ability to both have a RailCom cutout on one output while no RailCom cutout and standard bit stream on another?

Yes, there are independent settings of railcom cutout for the different outputs. The DCC driver applies them independently as well. (Of course an application may choose to tie them together for various reasons.)

src/freertos_drivers/common/DccOutput.hxx
balazsracz: done

src/freertos_drivers/common/DccOutput.hxx
balazsracz: Done

balazsracz:
> would it make sense to surface the basic API at a higher level than freertos_drivers/common? Perhaps under dcc or similar? I've implemented a similar structure in ESP32CS recently (splitting OPS and PROG) and can easily shift over to this API.

That's a good idea. Moved it to dcc/DccOutput.hxx.

PR #389, opened 2020-05-14
Factory reset example for ESP32

atanisoft:
Add example of using a free GPIO pin to force a factory reset of persistent configuration data during node startup.

There is a 10sec countdown with warning printed before the delete of the cdi.xml and openlcb_config files.

### DIFF
--- a/arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
+++ b/arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
@@ -288,19 +288,27 @@ void setup()
 #if defined(FACTORY_RESET_GPIO_PIN)
     // Check the factory reset pin which should normally read HIGH (set), if it
     // reads LOW (clr) delete the cdi.xml and openlcb_config
-    if (FACTORY_RESET_pin::instance()->is_clr())
+    if (!FACTORY_RESET_Pin::get())
     {
         printf("!!!! WARNING WARNING WARNING WARNING WARNING !!!!\n");
         printf("The factory reset GPIO pin %d has been triggered.\n",
                FACTORY_RESET_GPIO_PIN);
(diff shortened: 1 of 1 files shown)


arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -269,16 +282,35 @@ void setup()
         }
     }
 
+    // initialize all declared GPIO pins
+    GpioInit::hw_init();
+
+#if defined(FACTORY_RESET_GPIO_PIN)
+    // Check the factory reset pin which should normally read HIGH (set), if it
+    // reads LOW (clr) delete the cdi.xml and openlcb_config
+    if (FACTORY_RESET_pin::instance()->is_clr())
```
balazsracz, line 291: !FACTORY_RESET_Pin::get()

arduino/examples/ESP32IOBoard/ESP32IOBoard.ino
```diff
@@ -269,16 +282,35 @@ void setup()
...
+    {
+        printf("!!!! WARNING WARNING WARNING WARNING WARNING !!!!\n");
+        printf("The factory reset GPIO pin %d has been triggered.\n",
+               FACTORY_RESET_GPIO_PIN);
+        for (uint8_t sec = 10; sec > 0; sec--)
```
balazsracz, line 296: There does not seem to be any way to exit from this loop. Shouldn't we check that the pin is still held down?

atanisoft: Are you suggesting that we should verify on each loop that they still have the pin held LOW and if it is not still LOW at the end we do not reset?

balazsracz:
correct

PR #397, opened 2020-07-08
Switches the default hard fault handler to keep interrupts disabled.

balazsracz:
Switches the default hard fault handler to keep interrupts disabled.
It is not safe to reenable interrupts during hard fault handler.

In exchange we wait with an infinite loop by inline calling the blinker
routine.

---

Fixes missing blinking in the default interrupt handler.

### DIFF
--- a/boards/armv7m/default_handlers.h
+++ b/boards/armv7m/default_handlers.h
@@ -248,6 +248,18 @@ __attribute__((optimize("-O0"),unused)) void hard_fault_handler_step_2(unsigned 
     // hard_fault_handler_step_3.
 }
 
+/// This function will be called in an infinite loop from the hard fualt handler.
+///
+/// Define it in HwInit.cxx to enable blinking during hard faults. The function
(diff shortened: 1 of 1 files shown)


boards/armv7m/default_handlers.h
```diff
@@ -261,20 +273,25 @@ void hard_fault_handler_step_3(void) {
...
-        " ldr r3, [r0, 12]   \n"
-        " ldr r2, [r0, 8]    \n"
-        " ldr r1, [r0, 4]    \n"
-        " ldr r0, [r0, 0]    \n"
-        ".infloop:           \n"
-        " b.n .infloop        \n"
-        :: "i" ( 0x20 ) : "r0", "r1", "r2", "r3"
-        );
-    while (1);
+    while (1) {
```
bakerstu, line 276: Brace style.

balazsracz: done

PR #400, opened 2020-07-10
makes toolchain lookup compatible with `apt-get install clang-10` on linux

balazsracz:
Adds a linux.llvm port of the hub app.

### DIFF
--- a/etc/path.mk
+++ b/etc/path.mk
@@ -559,7 +559,10 @@ endif #EMLLVMPATH
 ##################### CLANGPPP ######################
 ifndef CLANGPPPATH
 SEARCHPATH := \
-  /usr/bin
+  /usr/bin \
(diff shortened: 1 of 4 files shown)


etc/path.mk
```diff
@@ -559,7 +559,10 @@ endif #EMLLVMPATH
 ##################### CLANGPPP ######################
 ifndef CLANGPPPATH
 SEARCHPATH := \
-  /usr/bin
+  /usr/bin \
+  /usr/lib/llvm-10/bin \
+  /usr/lib/llvm-9/bin \
+  /usr/lib/llvm-8/bin \
```
atanisoft, line 565: can we also add /usr/lib/llvm/{version}/bin as that is the path used on Gentoo

PR #405, opened 2020-07-15
Fixes FIFO overflow error in C32xxSPI driver

balazsracz:
causing infinite loop in the case a more than 32-byte long
transfer is initiated with 2 or 4 byte words.

### DIFF
--- a/src/freertos_drivers/ti/CC32xxSPI.hxx
+++ b/src/freertos_drivers/ti/CC32xxSPI.hxx
@@ -206,7 +206,7 @@ private:
         do
         {
             /* fill TX FIFO but make sure we don't fill it to overflow */
-            if (tx_len && (rx_len - tx_len) < (8 * sizeof(T)))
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/ti/CC32xxSPI.hxx
```diff
@@ -206,7 +206,7 @@ private:
         do
         {
             /* fill TX FIFO but make sure we don't fill it to overflow */
-            if (tx_len && (rx_len - tx_len) < (8 * sizeof(T)))
+            if (tx_len && ((rx_len - tx_len) < (32 / sizeof(T))))
```
bakerstu, line 209: The added parentheses should not be required, but also does no harm.

balazsracz: there is actually a warning that some compiler versions produce, -Wsuggest-parentheses. So adding these proactively when convenient is a good idea.

PR #407, opened 2020-07-19
Feature macro work to remove ARDUINO ifdefs.

balazsracz:
Adds feature macro OPENMRN_HAVE_POSIX_FD.
Replaces uses of ifdef ARDUINO with feature macros.

### DIFF
--- a/include/openmrn_features.h
+++ b/include/openmrn_features.h
@@ -48,6 +48,10 @@
 #define OPENMRN_FEATURE_REENT 1
 #endif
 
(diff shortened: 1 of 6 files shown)


include/openmrn_features.h
```diff
@@ -48,6 +48,10 @@
 #define OPENMRN_FEATURE_REENT 1
 #endif
 
+#if defined(__linux__) || defined(__MACH__) || defined(__WINNT__) || defined(ESP32) || defined(OPENMRN_FEATURE_DEVTAB)
+#define OPENMRN_HAVE_POSIX_FD 1
```
atanisoft, line 52: for consistency please add a /// comment for what this define enables.

balazsracz: good point, done.

PR #408, opened 2020-07-19
Adds a single-file implementation of the newlib file APIs for Arduino.

balazsracz:
The implementation forwards the calls to Arduino-STM32's
eeprom library. This is the bare minimum to support OpenMRN's
use of config file as the storage mechanism.

### DIFF
--- a/src/freertos_drivers/arduino/ArduinoFs.hxx
+++ b/src/freertos_drivers/arduino/ArduinoFs.hxx
@@ -60,7 +60,7 @@ public:
     /// Number of possible open file desriptors.
     static constexpr unsigned MAX_FD = 8;
     /// Offset markng that a file descriptor is not in use.
-    static constexpr unsigned UNUSED_FILE = (off_t)-1;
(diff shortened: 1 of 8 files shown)


src/freertos_drivers/arduino/ArduinoFs.hxx
```diff
@@ -0,0 +1,304 @@
...
+
+/// Class that holds information and code for the single-file filesystem for
+/// emulating eeprom.
+class FsStatic
+{
+public:
+    /// Number of possible open file desriptors.
+    static constexpr unsigned MAX_FD = 8;
+    /// Offset markng that a file descriptor is not in use.
+    static constexpr unsigned UNUSED_FILE = (off_t)-1;
```
atanisoft, line 63: would the datatype be better as off_t since it is only ever used in that type?

balazsracz: done

PR #409, opened 2020-07-28
Bakerstu macos cleanup

bakerstu:


etc/mach.x86_64.mk
```diff
@@ -1,6 +1,6 @@
 ifndef TOOLPATH
 TOOLPATH := $(shell \
-sh -c "if [ -d /usr/include/mach ]; then echo /usr/bin; \
+sh -c "if [ -d /Applications/Xcode.app/Contents/Developer ]; then echo /usr/bin; \
```
balazsracz, line 3: why isn't the toolpath somewhere in that directory that you check for?

bakerstu: It's complicated. I'll make a more detailed reply, but I don't have time at the moment.

balazsracz: I think if the binaries are in usr/bin then toolpath needs to be usr/bin so that's correct.
To make this code less confusing and not depend on where the xcode app was installed we could check like 
If [ -x /usr/bin/clang++ ]

bakerstu: Fixed.

PR #413, opened 2020-08-06
Add linux.aarch64 target.

RobertPHeller:
Add 64-bit ARM Linux as a OpenMRN target.   Initially supporting my Banana Pi M64 (Allwinner A64 SBC with a Raspberry Pi B+ form factor, including the Raspberry Pi 40-pin GPIO header.  I'm guessing this will also support a 'Pi 4 or really any 64-bit ARM Linux SBC running a 64-bit (aarch64) Linux O/S.

This is *lightly* tested -- I built and tested the hub application.  Not sure how to run more detailed tests.

### DIFF
--- a/etc/linux.aarch64.mk
+++ b/etc/linux.aarch64.mk
@@ -1,0 +1,45 @@
+# Get the toolchain paths for openmrn
+include $(OPENMRNPATH)/etc/path.mk
+
+
+ifndef TOOLPATH
+#TOOLPATHCOMMAND := $(shell \
+#sh -c "which aarch64-linux-gnu-gcc" \
+#)
+TOOLPATH := $(AARCH64LINUXGCCPATH)
+endif
+
+$(info armv7alinux toolpath '$(TOOLPATH)')
+
+# Get the $(CFLAGSENV), $(CXXFLAGSENV), $(LDFLAGSENV)
+include $(OPENMRNPATH)/etc/env.mk
+
+CC = $(TOOLPATH)/aarch64-linux-gnu-gcc
+CXX = $(TOOLPATH)/aarch64-linux-gnu-g++
+AR = $(TOOLPATH)/aarch64-linux-gnu-ar
+LD = $(TOOLPATH)/aarch64-linux-gnu-g++
+OBJDUMP = $(TOOLPATH)/aarch64-linux-gnu-objdump
+
+AROPTS=D
+
+HOST_TARGET := 1
+
+STARTGROUP := -Wl,--start-group
+ENDGROUP := -Wl,--end-group
+
+ARCHOPTIMIZATION = -g3 -O0 -march=armv8-a
+
+CSHAREDFLAGS = -c $(ARCHOPTIMIZATION) -Wall -Werror -Wno-unknown-pragmas \
+               -MD -MP -fno-stack-protector -D_GNU_SOURCE
+
+CFLAGS = $(CSHAREDFLAGS) -std=gnu99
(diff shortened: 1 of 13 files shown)


atanisoft:
@RobertPHeller do you know if this will work on an older Pi 2B+? If so, I'll pull the changes and test there for other parts as I've run other tests on there previously.

etc/linux.aarch64.mk
```diff
@@ -0,0 +1,45 @@
...
+
+CC = $(TOOLPATH)/aarch64-linux-gnu-gcc
+CXX = $(TOOLPATH)/aarch64-linux-gnu-g++
+AR = $(TOOLPATH)/aarch64-linux-gnu-ar
+LD = $(TOOLPATH)/aarch64-linux-gnu-g++
+OBJDUMP = $(TOOLPATH)/aarch64-linux-gnu-objdump
+
+AROPTS=D
+
+HOST_TARGET := 1
```
balazsracz, line 25: Is your intention to use this target *on* the device to compile stuff locally, or on your desktop machine to perform cross-compilation to the target, then copy the binaries?
A host target usually has /usr/bin/g++ as CXX. The paths above seem like a cross-compiling target.

Generally linux.x86 happens to work as an architecture-unspecific host target. It works for local compilation on all linux distributions and all architectures, including raspberry pis, beaglebones and others, also 32-bit and 64-bit. Yes, it is misnamed that way and should not really have x86 in the name of the target.

RobertPHeller: Actually, like the linux.armv7a target, it is actually both:  Right now, I compile (using the local "native" compiler) on one Raspberry Pi binaries for other 32-bit ARM Linux machines (other Raspberry Pis or Beagleboards).  Or I could run the program locally on the build system as a workbench test. I see the same thing here.

While in the generic case using the "linux.x86" target for all arch. Linux builds works, the linux.armv7a and linux.aarch64 have a few arch. specific optimizations (-march=).  The paths are not actually cross-compiling targets.  I just copied what was already there for linux.armv7a.  I guess it could be a cross-compile as well as native.

RobertPHeller:
@atanisoft  The PI 2B+ is a 32-bit processor.  It is not going to run 64-bit code.  Yes the linux.armv7a target does work with its native compiler -- I do that all the time targeting Beagles and Pi 3's.  It just that I have a Banana Pi M64, which is running Armbian Focal with an aarch64 kernel and userland so I added the build infrastructure for aarch64...  I've made no attempt to build an aarch64 cross-compiler on the (old) Pi 2B+ -- I expect that would take like forever to build --  it did take like forever to build the cross compiler for the STM & Tiva boards on the (old) Pi 2B+ (well about 3 days and used most of the 32gig microSD card and all of its 1gig of RAM).

atanisoft:
> The PI 2B+ is a 32-bit processor. It is not going to run 64-bit code.

Ahh yes, it has been a while since I looked at the specs for it. Is there an upgrade version you would recommend, ie the Pi 4?

RobertPHeller:
Pi 3 and Pi 4 are both 64 bit.  I don't know if Raspbian is available in aarch64.  The Banana Pi M64 is 64-bit and runs Armbian Focal. Has the same 40-pin header.  There are also cubiboards that are 64-bit and have the 40-pin header.  Also there is an Orange Pi board -- guessing it too has the 40-pin header and I believe it is 64-bit.  Common 64-bit ARM Linux chip is the Allwinner A64.

PR #417, opened 2020-08-09
Make dead code into less dead code.

bakerstu:


src/utils/FileUtils.cxx
```diff
@@ -106,10 +106,15 @@ void write_string_to_file(const string &filename, const string &data)
...
         offset += nr;
-        if (offset >= data.size()) break;
+        if (offset >= data.size())
+        {
+            break;
+        }
     }
-    if (nr < 0) {
-        fprintf(stderr, "error writing: %s\n", strerror(errno));
+    if (offset  != data.size())
```
balazsracz, line 114: there is an extra space here left of the !=

PR #418, opened 2020-08-12
Adds a partial implementation of VirtualMemorySpace

balazsracz:
The VirtualMemorySpace is an implementation of a MemorySpace interface which does not have a linear storage space behind. Instead, each variable is registered with a callback to write and read. This allows various advanced features, such as remapping address space, creating holes, really huge repetitions, etc.
This is the first cut of an implementation that is not feature complete, only one data type is supported, and has todos left.

### DIFF
--- a/src/openlcb/VirtualMemorySpace.hxx
+++ b/src/openlcb/VirtualMemorySpace.hxx
@@ -79,7 +79,7 @@ public:
     size_t write(address_t destination, const uint8_t *data, size_t len,
         errorcode_t *error, Notifiable *again) override
     {
-        if (destination > maxAddress_ || destination + len <= minAddress_)
+        if ((destination > maxAddress_) || ((destination + len) <= minAddress_))
         {
             *error = MemoryConfigDefs::ERROR_OUT_OF_BOUNDS;
             return 0;
@@ -129,7 +129,7 @@ public:
     size_t read(address_t source, uint8_t *dst, size_t len, errorcode_t *error,
         Notifiable *again) override
     {
-        if (source > maxAddress_ || source + len <= minAddress_)
+        if ((source > maxAddress_) || ((source + len) <= minAddress_))
         {
             *error = MemoryConfigDefs::ERROR_OUT_OF_BOUNDS;
             return 0;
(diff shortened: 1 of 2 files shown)


dpharris:
Cool!

atanisoft:
I agree, this looks very interesting.

src/openlcb/VirtualMemorySpace.hxx
```diff
@@ -0,0 +1,329 @@
...
+    /// @param error if set to non-null, then the operation has failed. If the
+    /// operation needs to be continued, then sets error to
+    /// MemorySpace::ERROR_AGAIN, and calls the Notifiable
+    /// @param again when a re-try makes sense. The caller should call write
+    /// once more, with the offset adjusted with the previously returned
+    /// bytes.
+    size_t write(address_t destination, const uint8_t *data, size_t len,
+        errorcode_t *error, Notifiable *again) override
+    {
+        if (destination > maxAddress_ || destination + len <= minAddress_)
```
bakerstu, line 82: Should parentheses be around (destination + len)? I'm unsure of the C rules and I have always used parentheses in this case.

src/openlcb/VirtualMemorySpace.hxx
```diff
@@ -0,0 +1,329 @@
...
+    /** @returns the number of bytes successfully read (before hitting end of
+     * space). If *error is set to non-null, then the operation has failed. If
+     * the operation needs to be continued, then sets error to ERROR_AGAIN, and
+     * calls the Notifiable @param again when a re-try makes sense. The caller
+     * should call read once more, with the offset adjusted with the previously
+     * returned bytes. */
+    size_t read(address_t source, uint8_t *dst, size_t len, errorcode_t *error,
+        Notifiable *again) override
+    {
+        if (source > maxAddress_ || source + len <= minAddress_)
```
bakerstu, line 132: same...

src/openlcb/VirtualMemorySpace.hxx
balazsracz: done

src/openlcb/VirtualMemorySpace.hxx
balazsracz: done

PR #422, opened 2020-09-01
Adds SocketCan support to the openmrn hub.

balazsracz:
This enables running the hub on a BeagleBone or a Raspberry Pi with a CAN HAT.
All other options are still usable, including repeating traffic between a SocketCAN port and a TCP client,
or a SocketCAN port and a USB port via GridConnect.


atanisoft:
> Does this make Alex's pull request for the same thing obsolete?

I looked over what @kiwi64ajs proposed in https://github.com/bakerstu/openmrn/pull/327 and I think that there may be a couple pieces that might still be relevant after this PR has been merged. Specifically the addition of the AvaHiMDNS.cxx file for the rpi target (perhaps a symlink?) and the readme describing how to configure/run the hub.

applications/hub/main.cxx
```diff
@@ -64,11 +66,16 @@ bool timestamped = false;
...
 
 void usage(const char *e)
 {
     fprintf(stderr, "Usage: %s [-p port] [-d device_path] [-u upstream_host] "
-                    "[-q upstream_port] [-m] [-n mdns_name] [-t] [-l]\n\n",
+                    "[-q upstream_port] [-m] [-n mdns_name] "
+#if defined(__linux__)
+            "[-s socketcan_interface] "
+#endif            
+            "[-t] [-l]\n\n",
```
atanisoft, line 78: should the text be aligned uniformly?

balazsracz:
ahh bummer. I forgot about that and did not notice it yesterday.
I will work on merging these two PRs. I see there is some unrelated changes popping up in Alex's branch that also need to be cleaned up.

TrainzLuvr:
I pest0red Balazs about SocketCAN support and he got this PR in overnight.

Meanwhile, I also found Alex's branch couple of days ago and compiled the Hub from his OpenMRN fork. But, I don't think it's working for me over the CAN bus (possibly due to my CAN shield - used an Arduino one, then swapped the transceivers to get 3V3 logic levels).

Hub does work over WiFi/Ethernet though as I can see my UWT-100 and JMRI sending packets to it. Did not connect the ESP32CS yet as I need a longer CAN cable for my test setup.

I am not sure what benefit AvahimDNS support adds, because the connecting devices would need to run mDNS queries via UDP, so this to me seems limited to TCP/IP networks only, unless it is something WiThrottle needs?

atanisoft:
> I am not sure what benefit AvahimDNS support adds, because the connecting devices would need to run mDNS queries via UDP, so this to me seems limited to TCP/IP networks only, unless it is something WiThrottle needs?

mDNS is used by the UWT-100 and other devices to find the hub without knowing it's IP address.

TrainzLuvr:
Ah right, WiFi...still TCP/IP. I keep thinking LCC because I see LCC below the WiFi bars and somehow I associate that with CAN bus. :)

applications/hub/main.cxx
balazsracz: done

balazsracz:
i merged Alex's changes into this PR.

PR #423, opened 2020-09-02
Moving socket listener stack size and backlog count to constants.

atanisoft:
As we discussed over email, the default number of sockets available for use is 10 with a max of 16. With the default backlog size of five it leaves only five additional sockets before errors will be reported by LwIP. ESP-IDF examples limit the backlog to one and thus that is the default with this PR.

This PR also includes optional naming of the SocketListener thread (useful when more than one listener is in use)

Fixes https://github.com/bakerstu/openmrn/issues/256

### DIFF
--- a/src/utils/constants.cxx
+++ b/src/utils/constants.cxx
@@ -152,3 +152,15 @@ DEFAULT_CONST(gridconnect_bridge_max_incoming_packets, 1);
 DEFAULT_CONST(gridconnect_bridge_max_outgoing_packets, 1);
 
 DEFAULT_CONST_FALSE(gridconnect_tcp_use_select);
+
+#ifdef ESP32
+/// Use a stack size of 3kb for SocketListener tasks.
+DEFAULT_CONST(socket_listener_stack_size, 3072);
+/// Allow one socket to be pending for accept() in SocketListener.
+DEFAULT_CONST(socket_listener_backlog, 1);
+#else
+/// Use a stack size of 1000 for SocketListener tasks.
+DEFAULT_CONST(socket_listener_stack_size, 1000);
+/// Allow up to five sockets to be pending for accept() in SocketListener.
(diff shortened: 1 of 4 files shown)


src/utils/constants.cxx
```diff
@@ -152,3 +152,15 @@ DEFAULT_CONST(gridconnect_bridge_max_incoming_packets, 1);
 DEFAULT_CONST(gridconnect_bridge_max_outgoing_packets, 1);
 
 DEFAULT_CONST_FALSE(gridconnect_tcp_use_select);
+
+#ifdef ESP32
+/// Use a stack size of 3kb for SocketListener tasks.
+DEFAULT_CONST(socket_listener_stack_size, 3072);
```
balazsracz, line 158: this used to be 2k, did it need to be bumped to 3k?

src/utils/socket_listener.cxx
```diff
@@ -51,34 +51,27 @@
 #include <signal.h>
 #include <unistd.h>
 
-#include "utils/socket_listener.hxx"
-
-#include "utils/macros.h"
+#include "nmranet_config.h"
 #include "utils/logging.h"
+#include "utils/macros.h"
+#include "utils/socket_listener.hxx"
```
balazsracz, line 57: please do keep the `$self.hxx` at the very top of the include list. Having it in the middle/bottom often hides missing includes.
In fact it should be above the `#include <...>` as well

balazsracz (approved):

src/utils/constants.cxx
atanisoft: Yes, it was unfortunately needed after further testing as part of ESP32 CS.

src/utils/socket_listener.cxx
atanisoft: Rearranged.

PR #425, opened 2020-09-04
Refactors all usages of strncpy.

### LINKED ISSUE #424: GCC8/9 produces warning on strncpy
Got this error:

```
arm-linux-gnueabihf-g++ -c -g -O3 -march=armv6 -marm -Wall -Werror -Wno-unknown-pragmas -MD -MP -fno-stack-protector -D_GNU_SOURCE -std=c++0x -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -D__USE_LIBSTDCPP__ -I./ -I/home/user/Development/openmrn/src/ -I/home/user/Development/openmrn/include -MD -MF SimpleNodeInfo.d /home/user/Development/openmrn/src/openlcb/SimpleNodeInfo.cxx -o SimpleNodeInfo.o
/home/user/Development/openmrn/src/openlcb/SimpleNodeInfo.cxx: In function ‘void openlcb::init_snip_user_file(int, const char*, const char*)’:
/home/user/Development/openmrn/src/openlcb/SimpleNodeInfo.cxx:70:12: error: ‘char* strncpy(char*, const char*, size_t)’ specified bound 63 equals destination size [-Werror=stringop-truncation]
     strncpy(data.user_name, user_name, sizeof(data.user_name));
     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/user/Development/openmrn/src/openlcb/SimpleNodeInfo.cxx:71:12: error: ‘char* strncpy(char*, const char*, size_t)’ specified bound 64 equals destination size [-Werror=stringop-truncation]
     strncpy(data.user_description, user_description,
     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             sizeof(data.user_description));
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

balazsracz:
Adds a new function str_populate that invokes strncpy in length-safe manner,
ensuring that a terminating null is present even if the string is truncated.
Using C++ templates this also allows the caller not to need to specify the
length of the char[] array, which is safer.

Replaces all calls to strncpy that were writing into a fixed size C array
with the new function.

Fixes #424

### DIFF
--- a/src/utils/format_utils.hxx
+++ b/src/utils/format_utils.hxx
@@ -36,6 +36,7 @@
 #define _UTILS_FORMAT_UTILS_HXX_
 
 #include <string>
+#include <string.h>
 
 /** Renders an integer to string, left-justified. @param buffer must be an at
  * @param buffer must be an at least 10 character long array.
@@ -161,6 +162,19 @@ string ipv4_to_string(uint8_t ip[4]);
 inline string ipv4_to_string(uint32_t ip)
 {
     return ipv4_to_string((uint8_t*)&ip);
+}
+
+/// Populates a character array with a C string. Copies the C string,
+/// appropriately truncating if it is too long and filling the remaining space
+/// with zeroes. Ensures that at least one null terminator character is
+/// present.
+/// @param dst a character array of fixed length, declared as char sdata[N]
+/// @param src a C string to fill it with.
+template <unsigned int N>
+inline void str_populate(char (&dst)[N], const char *src)
+{
+    strncpy(dst, src, N - 1);
+    dst[N - 1] = 0;
 }
 
 #endif // _UTILS_FORMAT_UTILS_HXX_
--- a/src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
+++ b/src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
@@ -765,7 +765,7 @@ void CC32xxWiFi::wlan_setup_ap(const char *ssid, const char *security_key,
                (uint8_t*)ssid);
     if (wlanRole == WlanRole::AP)
     {
-        strncpy(this->ssid, ssid, sizeof(this->ssid));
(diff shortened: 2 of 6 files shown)


src/utils/format_utils.hxx
```diff
@@ -163,4 +164,17 @@ inline string ipv4_to_string(uint32_t ip)
     return ipv4_to_string((uint8_t*)&ip);
 }
 
+/// Populates a character array with a C string. Copies the C string,
+/// appropriately truncating if it is too long and filling the remaining space
+/// with zeroes. Ensures that at least one null terminator character is
```
atanisoft, line 169: The comment of "filling the remaining space with zeros" does not seem to be implemented here. Perhaps adding:
```
bzero(dst, N);
```
would satisfy that?

balazsracz: Actually it does fill with null: https://pubs.opengroup.org/onlinepubs/009695399/functions/strncpy.html

>If the array pointed to by s2 is a string that is shorter than n bytes, null bytes shall be appended to the copy in the array  pointed to by s1, until n bytes in all are written.

TrainzLuvr:
I feel that the name of this function is somewhat misleading.

Why not just call it _safe_strncpy_ instead, so it suggests what's really behind it.

src/utils/format_utils.hxx
atanisoft: Ok, so based on that it should be zeroing the array automatically. We can consider this a non-issue in that case.

balazsracz:
> I feel that the name of this function is somewhat misleading.
> 
> Why not just call it _safe_strncpy_ instead, so it suggests what's really behind it.

because a function called safe_strncpy should have three arguments.

bakerstu (approved):

PR #434, opened 2020-09-18
Makes the total number of open file descriptors configurable.

balazsracz:
The configuration works with weak symbols that the application owner has
to override in their own .cxx file if needed to be changed.

It was not possible to make use of the usual OVERRIDE_CONST facility,
because it is important to have the files[] array allocated in .bss
and an array definition does not compile with the C++ compiler when
using an extern size symbol. Having files[] in .bss allows us not to
worry about initialization time race conditions.


bakerstu (approved):
Simple, but effective.

PR #435, opened 2020-09-19
Adds an approximate LRU counting algorithm.

balazsracz:
The GlobalLruCounter and a set of LruCounter<> objects cooperate in order
to create an approximate LRU order over a set of objects. The particular
optimization criterion is that the memory storage per object should be very
low, target is one byte.

### DIFF
--- a/src/utils/LruCounter.hxx
+++ b/src/utils/LruCounter.hxx
@@ -142,7 +142,8 @@ public:
             ++counter_;
             return;
         }
-        if (counter_ == std::numeric_limits<T>::max()) {
+        if (counter_ == std::numeric_limits<T>::max())
+        {
             // Counter is saturated.
             return;
         }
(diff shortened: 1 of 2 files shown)


src/utils/LruCounter.hxx
```diff
@@ -0,0 +1,158 @@
...
+private:
+    template <class T> friend class LruCounter;
+    /// Setting defining the exponent.
+    unsigned bitsPerBit_;
+    /// Rolling counter of global ticks. This is used by the local counters to
+    /// synchronize their increments.
+    unsigned tick_ {0};
+};
+
+template <class T> class LruCounter
```
bakerstu, line 120: Could you add a header comment to this template class?

bakerstu (approved):
This is interesting. I presume this may be used to manage a locomotive database to avoid a "slot == full" scenario by purging old entries on demand.

src/utils/LruCounter.hxx
balazsracz: done

balazsracz:
> This is interesting. I presume this may be used to manage a locomotive database to avoid a "slot == full" scenario by purging old entries on demand.

yup

Issue #440, opened 2020-10-07
Crash upon closing CC32xx socket


balazsracz:
We have a crash if using CC32xxWiFi sockets via HubDevice<> and then call ::close(fd) in the main thread.
The speciality of this case is that there are two threads that are pending (blocked) on performing read/write operations on the fd, while a third thread calls close on it.

Backtraces:
```
#0  0x20019036 in vPortExitCritical () at /home/bracz/FreeRTOS/FreeRTOSv10.0.0/FreeRTOS/Source/include/../../Source/portable/GCC/ARM_CM3/portmacro.h:230
#1  0x20019d68 in xQueueGenericSend (xQueue=0x20027600, pvItemToQueue=0x2002b5e0, pvItemToQueue@entry=0x0, xTicksToWait=<optimized out>, xTicksToWait@entry=0, xCopyPosition=64, 
    xCopyPosition@entry=0) at /FreeRTOS/FreeRTOSv10.0.0/FreeRTOS/Source/queue.c:841
#2  0x20016320 in sem_post (semaphore=semaphore@entry=0x20023d30 <g_SlInternalSpawnCB>)
    at /CC3220SDK/simplelink_cc32xx_sdk_3_40_00_05/source/ti/posix/freertos/semaphore.c:151
#3  0x200059d8 in _SlInternalSpawn (pEntry=pEntry@entry=0x20004fe5 <_SlDrvMsgReadSpawnCtx>, pValue=pValue@entry=0x0, flags=flags@entry=2)
    at /CC3220SDK/simplelink_cc32xx_sdk_3_40_00_05/source/ti/drivers/net/wifi/source/spawn.c:124
#4  0x20003f30 in _SlDrvMsgReadCmdCtx (IsLockRequired=<optimized out>, cmdOpcode=37890)
    at /CC3220SDK/simplelink_cc32xx_sdk_3_40_00_05/source/ti/drivers/net/wifi/source/driver.c:1819
#5  _SlDrvCmdOp (pCmdCtrl=pCmdCtrl@entry=0x2001ae54 <_SlSockCloseCmdCtrl>, pTxRxDescBuff=pTxRxDescBuff@entry=0x20025d30, pCmdExt=pCmdExt@entry=0x0)
    at /CC3220SDK/simplelink_cc32xx_sdk_3_40_00_05/source/ti/drivers/net/wifi/source/driver.c:866
#6  0x200167a6 in sl_Close (sd=<optimized out>)
[...]

PR #441, opened 2020-10-07
Adds an intermediate stage in closing an fd.

balazsracz:
Adds a new bit to mark an fd being in shutdown. This is a state
during close that marks the fd as being bad, with all kernel calls
(from other threads) returning EBADF if they refer to this fd.
At the same time the fd is not marked as free for reuse by another
open.

This fixes the race condition in
https://github.com/bakerstu/openmrn/issues/440

### DIFF
--- a/src/freertos_drivers/common/Device.cxx
+++ b/src/freertos_drivers/common/Device.cxx
@@ -148,6 +148,7 @@ int Device::close(struct _reent *reent, int fd)
     int result = f->dev->close(f);
     if (result < 0)
     {
+        files[fd].inshdn = false;
         errno = -result;
         return -1;
     }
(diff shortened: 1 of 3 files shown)


src/freertos_drivers/common/Device.cxx
```diff
@@ -144,6 +144,7 @@ int Device::close(struct _reent *reent, int fd)
         // stdin, stdout, and stderr never get closed
         return 0;
     }
+    files[fd].inshdn = true;
```
bakerstu, line 147: Question: What if f->dev->close(f) returns an error? Should we reset the inshdn back to false? Otherwise, I don't think we can ever free the fd because we cannot make another close() call later.

bakerstu (approved):

src/freertos_drivers/common/Device.cxx
balazsracz: Good catch, fixed.

PR #445, opened 2020-10-13
Always report from railcom driver

balazsracz:
Ensures that the railcom driver returns data from the feedback channel even if
there was no cutout generated.

### DIFF
--- a/src/freertos_drivers/common/RailcomDriver.hxx
+++ b/src/freertos_drivers/common/RailcomDriver.hxx
@@ -58,6 +58,9 @@ public:
   /** Instructs the driver that the railcom cutout is over now. The driver
    *  will use this information to disable the UART receiver. */
   virtual void end_cutout() = 0;
+  /** Called instead of start/mid/end-cutout at the end of the current packet
+   * if there was no cutout requested. */
+  virtual void no_cutout() = 0;
   /** Specifies the feedback key to write into the received railcom data
    *  packets. This feedback key is used by the application layer to correlate
    *  the stream of DCC packets to the stream of Railcom packets. This method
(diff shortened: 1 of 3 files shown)


src/freertos_drivers/ti/TivaRailcom.hxx
```diff
@@ -440,6 +440,24 @@ private:
         Debug::RailcomCh2Data::set(false);
         Debug::RailcomDriverCutout::set(false);
     }
+
+    void no_cutout() OVERRIDE
```
bakerstu, line 444: Please add a Doxygen header or "cond IGNORE" block.

src/freertos_drivers/common/RailcomDriver.hxx
```diff
@@ -74,6 +77,7 @@ class NoRailcomDriver : public RailcomDriver {
   void start_cutout() OVERRIDE {}
   void middle_cutout() OVERRIDE {}
   void end_cutout() OVERRIDE {}
+  void no_cutout() OVERRIDE {}
```
bakerstu, line 80: Please add Doxygen Header. Could use a "cond IGNORE" block tag if you prefer no suppress Doxygen warnings to keep these definitions inline and brief since the comment could be considered trivial, already covered in the base class.

bakerstu (approved):

src/freertos_drivers/common/RailcomDriver.hxx
balazsracz: I found this: https://www.doxygen.nl/manual/config.html#cfg_inherit_docs

Based on that if I do add any comment here, I will defeat the purpose of inheriting docs. I generated the docs and this is what I got for NoRailcomDriver:

image

so it seems to be working as expected.

src/freertos_drivers/ti/TivaRailcom.hxx
balazsracz: see below

PR #447, opened 2020-10-17
Keeps track of the number of connected clients

balazsracz:
in the GridConnect TCP hub.

### DIFF
--- a/src/openlcb/SimpleStack.hxx
+++ b/src/openlcb/SimpleStack.hxx
@@ -383,8 +383,15 @@ public:
     {
         /// @TODO (balazs.racz) make this more efficient by rendering to string
         /// only once for all connections.
-        /// @TODO (balazs.racz) do not leak this.
-        new GcTcpHub(can_hub(), port);
+        gcHubServer_.reset(new GcTcpHub(can_hub(), port));
+    }
+
+    /// Retrieve the instance of the GridConnect Hub server, hwich was started
+    /// with start_tcp_hub_server().
+    /// @return the TCP hub server, or nullptr if no server was ever started.
+    GcTcpHub *get_tcp_hub_server()
+    {
+        return gcHubServer_.get();
     }
 
     /// Connects to a CAN hub using TCP with the gridconnect protocol.
@@ -473,6 +480,9 @@ private:
     /// the CAN interface to function. Will be called exactly once by the
     /// constructor of the base class.
     std::unique_ptr<PhysicalIf> create_if(const openlcb::NodeID node_id);
+
+    /// Holds the ownership of the TCP hub server (if one was created).
+    std::unique_ptr<GcTcpHub> gcHubServer_;
 };
 
 class SimpleTcpStackBase : public SimpleStackBase
--- a/src/openlcb/SimpleStack.hxx
+++ b/src/openlcb/SimpleStack.hxx
@@ -386,7 +386,7 @@ public:
         gcHubServer_.reset(new GcTcpHub(can_hub(), port));
     }
 
-    /// Retrieve the instance of the GridConnect Hub server, hwich was started
+    /// Retrieve the instance of the GridConnect Hub server, which was started
     /// with start_tcp_hub_server().
     /// @return the TCP hub server, or nullptr if no server was ever started.
     GcTcpHub *get_tcp_hub_server()
--- a/src/utils/GcTcpHub.cxx
+++ b/src/utils/GcTcpHub.cxx
@@ -60,8 +60,8 @@ void GcTcpHub::notify()
 
 GcTcpHub::GcTcpHub(CanHubFlow *can_hub, int port)
     : canHub_(can_hub)
-    , tcpListener_(port, std::bind(&GcTcpHub::on_new_connection, this,
-                                   std::placeholders::_1))
+    , tcpListener_(port,
+          std::bind(&GcTcpHub::on_new_connection, this, std::placeholders::_1))
 {
 }
 
--- a/src/utils/GcTcpHub.cxx
+++ b/src/utils/GcTcpHub.cxx
@@ -38,16 +38,29 @@
 #include "nmranet_config.h"
 #include "utils/GridConnectHub.hxx"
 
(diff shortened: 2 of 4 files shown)


atanisoft:
Perhaps this can be extended slightly further to actively reject new clients when the current client count is over a certain limit?

src/openlcb/SimpleStack.hxx
```diff
@@ -383,8 +383,15 @@ public:
     {
         /// @TODO (balazs.racz) make this more efficient by rendering to string
         /// only once for all connections.
-        /// @TODO (balazs.racz) do not leak this.
-        new GcTcpHub(can_hub(), port);
+        gcHubServer_.reset(new GcTcpHub(can_hub(), port));
+    }
+
+    /// Retrieve the instance of the GridConnect Hub server, hwich was started
```
atanisoft, line 389: hwich -> which

src/utils/GcTcpHub.hxx
```diff
@@ -60,16 +60,28 @@ public:
         return tcpListener_.is_started();
     }
 
+    /// @return currently connected client count.
+    unsigned get_num_clients()
+    {
+        return numClients_;
```
atanisoft, line 66: do we need an AtomicHolder here?

src/openlcb/SimpleStack.hxx
balazsracz: done

src/utils/GcTcpHub.hxx
balazsracz: I don't think it's super meaningful. It would be necessary if the data type of numClients_ were thread-compatible but not thread-safe for the read accessor that we are using.

balazsracz:
> Perhaps this can be extended slightly further to actively reject new clients when the current client count is over a certain limit?

I don't need that feature. Do you?

src/utils/GcTcpHub.hxx
atanisoft: Perhaps it isn't necessary but there is an AtomicHolder usage for increment/decrement of this same var.

balazsracz: That's needed because it is a write access. A primitive type is thread-compatible but not thread-safe for writes.

atanisoft:
> I don't need that feature. Do you?

Possibly, the esp32 heap quickly reduces (and may crash) when both CAN and TCP/IP GC Hub are enabled and more than a couple clients are connected. I've considered dropping Hub support on the esp32 entirely for this reason and have already disabled it on the esp32s2 (slightly lower heap to start with).

src/utils/GcTcpHub.hxx
atanisoft: Ahh, yes that is correct. We can ignore this here then.

balazsracz:
> > I don't need that feature. Do you?
> 
> Possibly, the esp32 heap quickly reduces (and may crash) when both CAN and TCP/IP GC Hub are enabled and more than a couple clients are connected. I've considered dropping Hub support on the esp32 entirely for this reason and have already disabled it on the esp32s2 (slightly lower heap to start with).
> 
> If the GC connections were sharing the payload it may not be as much of an issue though.

Did you debug why this crashes? Or under which condition?
- e.g. does it need a heavy load of gridconnect packets (e.g. a firmware upgrade or a CDI download?)
- does it need the remote end to be unresponsive (i.e. having a lot of unacked TCP packets in the window).

Did you try setting SO_SNDBUF to a low value (e.g. 2 kbytes)?

atanisoft:
> Did you debug why this crashes?

Not fully yet, I'll be delving into it before taking any action to disable.

> Did you try setting SO_SNDBUF to a low value (e.g. 2 kbytes)?

I'll need to do some testing with that, the default is 5744 bytes (4x MSS).

src/utils/GcTcpHub.cxx
```diff
@@ -38,17 +38,30 @@
...
 
-void GcTcpHub::OnNewConnection(int fd)
+void GcTcpHub::on_new_connection(int fd)
 {
     const bool use_select =
         (config_gridconnect_tcp_use_select() == CONSTANT_TRUE);
-    create_gc_port_for_can_hub(canHub_, fd, nullptr, use_select);
+    {
+        AtomicHolder h(this);
+        numClients_++;
```
bakerstu, line 47: Not sure if modern compilers are smart enough to substitute a pre-increment if they can, but a pre-increment should be less instruction cycles.

src/utils/GcTcpHub.cxx
```diff
@@ -38,17 +38,30 @@
...
+    }
+    create_gc_port_for_can_hub(canHub_, fd, this, use_select);
+}
+
+void GcTcpHub::notify()
+{
+    AtomicHolder h(this);
+    if (numClients_)
+    {
+        numClients_--;
```
bakerstu, line 57: same.

bakerstu (approved):

src/utils/GcTcpHub.cxx
balazsracz: The recommendation to use pre-increment instead of post-increment comes from Scott Meyers'  Effective C++ book, and specifically applies to class types that define the pre- and post-increment operator. The most typical place where we encounter such classes are STL iterators. The recommendation was true until about GCC 2.95, beyond which however GCC figured out how to optimize away the copies even from iterators if the implementation available for inlining. It also means that this recommendation has never been true for primitive types.

balazsracz:
> > Did you debug why this crashes?
> 
> Not fully yet, I'll be delving into it before taking any action to disable.
> 
> > Did you try setting SO_SNDBUF to a low value (e.g. 2 kbytes)?
> 
> I'll need to do some testing with that, the default is 5744 bytes (4x MSS).

I will defer the client count limiting, because it really has to be a feature of the SocketListener instead. When we have limiting we must make sure not to call the accept() after the connection limit has been reached. So the accept thread needs to be paused and resumed, which needs API changes there.
Feel free to file a feature request.

PR #453, opened 2020-10-19
Revision script fixes

### LINKED ISSUE #233: Revision.py prints fake errors
When revision.py runs, even upon a successful run it will print error messages to the standard error.
This is confusing for the users; we should not print error messages on successful runs.

balazsracz:
- suppresses the error messages that are currently printed on the output at every run.
- adds a command to the hxx output that looks like this:
  #define REVISION_GIT_HASH "b60e1f7:-d-u"
  This is suitable for being included into the SNIP version string.

Fixes #233.

### DIFF
--- a/bin/revision.py
+++ b/bin/revision.py
@@ -93,6 +93,8 @@ if options.gcc != None :
     outputcxx += '    "' + gcc + '",\n'
     outputhxx += '"' + gcc + '\\n"\n'
 
+main_git_hash = None
+    
 for x in inputs :
     print x
     # go into the root of the repo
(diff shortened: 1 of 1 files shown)


bin/revision.py
```diff
@@ -165,14 +172,18 @@
...
                            "-I", """Sun, """, "-I", """Mon, """,
                            "-I", """Tue, """, "-I", """Wed, """,
                            "-I", """Thu, """, "-I", """Fri, """,
                            "-I", """Sat, """,
                            options.output + 'Try.cxxout',
-                           options.output + '.cxxout'], stdout=f_null)
+                           options.output + '.cxxout'],
+                          stdout=f_null, stderr=f_null)
+# disable this because we are not actually writing a cxx file above.
+diffcxx = 0
```
bakerstu, line 186: Does this result in dead code that should be removed from the script?

bakerstu (approved):

bin/revision.py
balazsracz: I think you might be better in advising on this question as the author of this script. The script seems to be written for generating an hxx and a cxx file, but it only generates one with the concatenation of the code for both. Why is this the case? The comments in the code did not give a hint. So this can be cleaned up in two ways and I would hope you can decide which way to go:
1. Make it actually generate a cxx file.
2. Clean up the code to that was there to generate the cxx file.

balazsracz: I'll merge this now as is and we can discuss which way to move forward.

PR #454, opened 2020-10-20
Add node brownout and ident event constants

atanisoft:
Currently I'm defining the brownout event in my esp32-s2 io code but would rather consume it directly from openlcb::Defs if possible. There are other well-known events but these two seemed to be more universally applicable to nodes.


src/openlcb/Defs.hxx
```diff
@@ -246,6 +246,16 @@ struct Defs
     /// station may react to this by restoring locomotive speed settings.
     static constexpr uint64_t CLEAR_EMERGENCY_STOP_EVENT = 0x010000000000FFFCULL;
 
+    /// "Power supply brownout detected below minimum required by node"
```
balazsracz, line 249: Isn't there a second of this with a similar flavor but different formulation? It might be worth adding both of them.

atanisoft: Yes, there is a "below standard" one as well. I wasn't sure if it should be added or not but can add it.

src/openlcb/Defs.hxx
```diff
@@ -246,6 +246,16 @@ struct Defs
...
+    /// This event can be generated when a node detects that it has
+    /// insufficient power for normal operations.
+    static constexpr uint64_t NODE_POWER_BROWNOUT_EVENT = 0x010000000000FFF1ULL;
+
+    /// "Ident button combination pressed"
+    /// This event can be generated by a node when it is instructed to generate
+    /// an identification event.
+    static constexpr uint64_t NODE_IDENT_BUTTON_EVENT = 0x010000000000FE01ULL;
```
balazsracz, line 257: This is FE00 in the standard.

atanisoft: Oops, will fix in a min. I read it off-by-one line.

balazsracz (approved):

PR #455, opened 2020-10-25
Adds a flag to indicate consist messages

balazsracz:
Adds a flag bit to the command byte in the traction request messages.
This bit will be set to one when a set speed / set fn message is forwarded to a listener.
The listener (e.g. a consisted train node) can deduce from this whether the message is
applied from a consist or directly from a throttle.


balazsracz:
@JimKueneman FYI.

bakerstu (approved):

PR #456, opened 2020-10-25
Adds traction node listener policy callback

balazsracz:
Adds a policy function to the traction node which allows the node implementation to decide whether a function change request should be applied or not.
Adds additional unit tests for consisting with function forwarding.


src/openlcb/TractionTrain.cxx
```diff
@@ -213,7 +213,20 @@ struct TrainService::Impl
                     uint16_t value = payload()[4];
                     value <<= 8;
                     value |= payload()[5];
-                    train_node()->train()->set_fn(address, value);
+                    bn_.reset(this);
+                    bool should_apply =
+                        train_node()->function_policy(nmsg()->src, payload()[0],
+                            address, value, bn_.new_child());
+                    if (!bn_.abort_if_almost_done())
```
bakerstu, line 220: What is this "abort_if_almost_done()" about. Could you ad some more detailed commentary?

bakerstu (approved):

src/openlcb/TractionTrain.hxx
```diff
@@ -146,7 +162,15 @@ private:
 class TrainNodeWithConsist : public TrainNode {
 public:
     ~TrainNodeWithConsist();
-    
+
+    bool function_policy(NodeHandle src, uint8_t command_byte, uint32_t fnum,
+        uint16_t value, Notifiable *done) override
+    {
+        AutoNotify an(done);
+        // The default function policy applies everything.
```
bakerstu, line 170: This comment should be the doxygen header comment for the function.

src/openlcb/TractionTrain.cxx
balazsracz: done

src/openlcb/TractionTrain.hxx
balazsracz: done

PR #464, opened 2020-11-16
Protect unimplemented POSIX calls in BroadcastTime::clear_timezone() to allow OpenMRNLite to compile on Arduino flavored platforms.

RobertPHeller:
setenv() and tset() not available in the Arduino ESP32 platform.  This fix takes those calls out. I don't know what (if anything) could replace them, but making BroadcastTime::clear_timezone() seems like a harmless short term solution.

### DIFF
--- a/src/openlcb/BroadcastTime.cxx
+++ b/src/openlcb/BroadcastTime.cxx
@@ -46,8 +46,10 @@ namespace openlcb
 //
 void BroadcastTime::clear_timezone()
 {
+#ifndef ARDUINO
         setenv("TZ", "GMT0", 1);
         tzset();
+#endif
 }
 
 } // namespace openlcb
--- a/src/openlcb/BroadcastTime.cxx
+++ b/src/openlcb/BroadcastTime.cxx
@@ -46,7 +46,7 @@ namespace openlcb
 //
 void BroadcastTime::clear_timezone()
 {
-#ifndef ARDUINO
+#ifndef ESP32
         setenv("TZ", "GMT0", 1);
         tzset();
 #endif
--- a/src/freertos_drivers/common/WifiDefs.hxx
+++ b/src/freertos_drivers/common/WifiDefs.hxx
@@ -28,20 +28,12 @@ enum class WlanState : uint8_t
     UPDATE_DISPLAY = 20,
 };
 
-// This is needed when this file is included into a ESP32 OpenMRNLite 
-// application.  The esp32 Arduino.h file defines DEFAULT as a macro
-// constant and that breaks the compilation of of the WlanRole enum
-// class below.  RPH November 2020
-#ifdef DEFAULT
-#undef DEFAULT
-#endif
-
 /** Operating Role.
  */
 enum class WlanRole : uint8_t
 {
     UNKNOWN = 0,       /**< Default mode (from stored configuration) */
-    DEFAULT = UNKNOWN, /**< Default mode (from stored configuration) */
+    DEFAULT_ROLE = UNKNOWN, /**< Default mode (from stored configuration) */
     STA,               /**< Wi-Fi station mode */
     AP                 /**< Wi-Fi access point mode */
 };
--- a/src/freertos_drivers/common/WifiDefs.hxx
+++ b/src/freertos_drivers/common/WifiDefs.hxx
@@ -28,6 +28,14 @@ enum class WlanState : uint8_t
     UPDATE_DISPLAY = 20,
 };
 
+// This is needed when this file is included into a ESP32 OpenMRNLite 
+// application.  The esp32 Arduino.h file defines DEFAULT as a macro
+// constant and that breaks the compilation of of the WlanRole enum
+// class below.  RPH November 2020
+#ifdef DEFAULT
+#undef DEFAULT
+#endif
+
 /** Operating Role.
  */
 enum class WlanRole : uint8_t


balazsracz:
tbh I have no idea why this code performs global timezone related operations. It seems not very friendly for a library to unconditionally change timezone settings of any application that it is linked into.
@bakerstu could you describe why these calls are ever needed and whether we could get rid of them entirely?

src/openlcb/BroadcastTime.cxx
```diff
@@ -46,8 +46,10 @@ namespace openlcb
 //
 void BroadcastTime::clear_timezone()
 {
+#ifndef ARDUINO
```
atanisoft, line 49: Please use `ESP32` instead of  `ARDUINO` since these two APIs are available in a later version of arduino-esp32 (by nature of pulling in ESP-IDF v4.0+).

atanisoft:
These APIs will be available only with arduino-esp32 v2.0.0 or later which is built on top of ESP-IDF v4.2 (the APIs are available with v4.0 or later). However, they are problematic since the TZ variable applies to all threads and not just the one thread that is setting this.

src/openlcb/BroadcastTime.cxx
RobertPHeller: Done.  I sort of guessed as to which preprocessor symbol to use.

atanisoft: `ARDUINO` is a good guess, but having it as `ESP32` is better since this library will also work in ESP-IDF env which does not define `ARDUINO`.

bakerstu:
@balazsracz First off, this setting is process local, it does not effect other processes in the system. I have confirmed this by running the example application on Linux.

My recollection is that this sets the ground work for transparent support of localtime(), which uses the timezone setting to automatically apply timezone offsets to the result, including correcting for date, day of week, month, year, etc. This is a planned feature of our implementation, though it looks like I never implemented it. It would allow two fast clocks to be synchronized using a common understanding of GMT, but transparently display different timezone outputs.

atanisoft:
> this setting is process local, it does not effect other processes in the system.

On the ESP32 (and likely other embedded platforms) the TZ setting may be global to all threads (processes). Perhaps we should mask this to only be on Linux (or other full OS platforms)?

bakerstu:
@atanisoft The ESP32 is a single process system. You are correct that the setting is global to all threads in the process, in the ESP32 case, that happens to also be all threads in the system. I don't see any downside to this though, can you explain your concern?

atanisoft:
> can you explain your concern?

Setting TZ sets the *LOCAL* TZ for the ESP32 and impacts all time calculations. Perhaps it is a bug on the ESP32 but it is how they advise setting the TZ (ie: PST8PDT,M3.2.0,M11.1.0) which is used for converting UTC <-> local through the various APIs.

bakerstu:
Interesting. That sounds like a seriously deficient implementation. Definitely not compliant to the POSIX standards.

atanisoft:
> Definitely not compliant to the POSIX standards.

Very likely not the only one.

bakerstu:
I'm not so sure. All the big POSIX OS's have this right (Linux, BSD variants, QNX, VxWorks, etc...), newlib has this right, and Dinkumware has it right. That is 99% of the market. I honestly don't know how Esspresif screwed his up. They must have been trying very hard.

bakerstu (approved):

atanisoft:
I'm not sure how they might have broken it either. The only docs on TZ is available [here](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/system_time.html#timezones).

bakerstu:
I'm okay with the fix proposed for now. I approved the pull request.

bakerstu:
> I'm not sure how they might have broken it either. The only docs on TZ is available [here](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/system_time.html#timezones).

After reading the documentation, I don't see why there would be any issue with changing the timezone. Are you saying that you tried changing the timezone and the result is that the ESP32 system malfunctioned in some way. This would imply that some infrastructure code somewhere is using localtime when it shouldn't.

atanisoft:
The problem is it sets it globally for the esp32 and if you integrate code for SNTP it will get the time offset incorrect either in the fast clock or other usages of time. Calling gmtime() will work even without TZ being set, at least on the esp32, but localtime() will be returned as UTC0 even if your code earlier configured for PST8....

bakerstu:
Normally using localtime to convert from gmtime is ridiculously convenient, I would not want to write that logic myself. That is why I wanted to use it. I'll do some research to see if there is any standard alternative access to that conversion logic without having to globally change the timezone. Perhaps there is an API I can pass a timezone into.

atanisoft:
Perhaps we can add a getenv() call and see if TZ is already set and if not then default to UTC0? It shouldn't impact localtime() or gmtime() if it has been set externally since the platform should still take care of conversion automatically to UTC/GMT.

balazsracz:
I still don't understand why we need to *change* the user's timezone setting. I still maintain that having a library override the user's preferred timezone is undesired behavior, because no application can now link this library and have correct time operations according to the user's expectations. This is because the setting is process-local, therefore it applies to the entire process.

I can only think of two use-cases:
- either we want to use the user's timezone, in which case we should respect the TZ variable, and use `localtime_r()`.
- or we want to use UTC, in which case we should use `gmtime_r()`.

I can see no use-case where we want to override TZ to UTC and then use `localtime`. That behavior seems poisonous for the process.
Also, from the man page of localtime:

> The localtime() function converts the calendar time timep to broken-down time representation, expressed relative to the user's specified  timezone.  The function acts as if it called tzset(3) [...]

Therefore I argue that these two calls are unnecessary and should be removed under all OSes.

bakerstu:
First, I am okay to merge this fix.

I’ve thought about the timezone issue a bit more. I seem to recall that this might have been required in Linux for the gmtime to work properly, though based on the POSIX documentation, this should not be the case. I need to test this, which should be low effort. I think the best solution is to provide interface options that use gmtime or localtime (gmtime is what is always transferred over the LCC wire). Then it is up to the user application, not our BroadcastTime library, to set the timezone and optionally use the localtime interface appropriately, should that be a desirable feature.

I need to spend some cycles on BroadcastTime anyways, I’ve got a hanging pull request. This might be a good excuse to jump back on this.

PR #467, opened 2020-11-21
Stores reserved aliases in the local alias cache

balazsracz:
Updates the code and API of the AliasAllocator to store the list of aliases that are reserved but unused in the local_aliases() cache. This allows a large number of aliases to be pre-allocated using the BulkAliasAllocator without incurring extra memory overhead for keeping a list of Buffer<> objects in a queue. 
This PR makes the output of the BulkAliasAllocator be usable by the IfCan send flows.
The API contract of for alias allocation between the IfCan send flow and the AliasAllocator changes. Previously it was using a QAsync; how there is a direct sync/async API exposed on the AliasAllocator object. This new API simplifies the interaction because the checking for conflicts on reserved aliases is not the responsibility of the IfCan flow anymore but the AliasAllocator.

There is a feature that gets impacted by this change: to keep a certain number of reserved aliases around at all times even if they are not in use by any local node ID. The storage and API to trigger this feature was in the queued object that does not exist anymore. The feature is now triggered instead by an OVERRIDE_CONST. The default behavior will also change from true to false: previously an OpenMRN IO board would allocate two aliases and use only one of them; after this PR only one alias will be allocated and used by a simple IO board. Applications that use many virtual nodes should set the OVERRIDE_CONST instead.

Adds an API to the AliasCache to walk entries in node-ID-ascending order.


src/openlcb/AliasAllocator.hxx
```diff
@@ -126,26 +126,46 @@ public:
...
     /** If there is a pending alias allocation waiting for the timer to expire,
      * finishes it immediately. Needed in test destructors. */
     void TEST_finish_pending_allocation();
 
     /** Adds an allocated aliad to the reserved aliases queue.
         @param alias the next allocated alias to add.
     */
-    void TEST_add_allocated_alias(NodeAlias alias, bool repeat=false);
-    
+    void TEST_add_allocated_alias(NodeAlias alias);
```
atanisoft, line 161: Q, should these TEST_XXX methods be guarded by a GTEST define so that they do not unintentionally get used at runtime?

src/openlcb/AliasAllocator.cxx
```diff
@@ -89,15 +91,65 @@ void AliasAllocator::return_alias(NodeID id, NodeAlias alias)
         if_can()->frame_write_flow()->send(b);
     }
 
-    // This is synchronous allocation, which is not nice.
+    add_allocated_alias(alias);
+}
+
+void AliasAllocator::add_allocated_alias(NodeAlias alias)
+{
+    /** @TODO(balazs.racz): We leak aliases here in case of eviction by the
```
bakerstu, line 99: The correct syntax is @todo for Doxygen.

bakerstu (approved):

src/openlcb/AliasAllocator.hxx
balazsracz: done

src/openlcb/AliasAllocator.cxx
balazsracz: Done -- actually this todo is obsolete. Replaced it with a comment explaining why.

PR #470, opened 2020-11-22
Add support for C++11 compilation of std::make_unique()

atanisoft:
Since arduino-esp32 defaults to C++11 (due to ESP-IDF using the same) the usage of std::make_unique caused build failures for @RobertPHeller (as noted on groups.io). After a bit of digging into how to detect that we are building on an older version of the C++ standard I was able to include a version of std::make_unique as submitted for C++ standards as a header include.

### DIFF
--- a/src/openlcb/BulkAliasAllocator.cxx
+++ b/src/openlcb/BulkAliasAllocator.cxx
@@ -34,6 +34,7 @@
 
 #include "openlcb/BulkAliasAllocator.hxx"
 #include "openlcb/CanDefs.hxx"
+#include "utils/MakeUnique.hxx"
 
 namespace openlcb
 {
--- a/src/utils/MakeUnique.hxx
+++ b/src/utils/MakeUnique.hxx
@@ -1,0 +1,50 @@
+/**
+ * \file MakeUnique.hxx
+ *
+ * C++11 version of std::make_unique which is only available from c++14 or
+ * later.
+ * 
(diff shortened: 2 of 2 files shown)


balazsracz (approved):

bakerstu (approved):
It is unfortunate that in 2020, almost 2021, we need to still support C++11, nearly a decade old.

atanisoft:
Agreed, it would really be good to move up to a minimum of c++14 (maybe even c++17)

RobertPHeller:
At Sun, 22 Nov 2020 10:07:01 -0800 bakerstu/openmrn <reply@reply.github.com> wrote:

> 
> @bakerstu approved this pull request.
> 
> It is unfortunate that in 2020, almost 2021, we need to still support C++11,
> nearly a decade old.

I suspect that since the ESP32 is not as mainstream as, say, AVR or ARM, and
the dev support crew for the ESP32 is likely small (and overworked?), it is
going to take some time for them to vet and/or update the SDK codebase up to
the "current" (2020) standards.

> 
> 
> 

-- 
heller@deepsoft.com       -- Webhosting Services

atanisoft:
It's a bit more complicated. They have an issue tracking the increase to at least c++14 over on esp-idf but it is stalled until v5.0 is branched, which I suspect is unlikely anytime soon. The good news is the code is compatible with at least c++14 now.

PR #473, opened 2020-11-23
ESP32: Prep for Arduino-esp32 v1.0.5

atanisoft:
The ESP-IDF version detection that was in place was no longer working due to ESP-IDF v3.3 (picked up in arduino-esp32 1.0.5) defining `ESP_IDF_VERSION_MAJOR` which used to only be defined in in v4.0+.

Shifting to use __has_include to switch between the new and old include paths with fallback to use old paths when __has_include is not usable (though it always should be with GCC).

I'll be doing further work on this as part of the ESP-IDF v4.x support since there are a few breaking API changes in ESP-IDF v4.1 which I'll submit in another PR.

This has been tested with v1.0.4 and v1.0.5-rc2.

### DIFF
--- a/src/freertos_drivers/esp32/Esp32WiFiManager.cxx
+++ b/src/freertos_drivers/esp32/Esp32WiFiManager.cxx
@@ -48,17 +48,42 @@
 #include <mdns.h>
 #include <tcpip_adapter.h>
 
-// ESP-IDF v4+ has a slightly different directory structure to previous
-// versions.
-#ifdef ESP_IDF_VERSION_MAJOR
-// ESP-IDF v4+
+// Starting in ESP-IDF v4.0 a few header files have been relocated so we need
+// to adjust the include paths accordingly. If the __has_include preprocessor
+// directive is defined we can use it to find the appropriate header files.
+// If it is not usable then we will default the older header filenames.
+#if defined __has_include
+
+// rom/crc.h was relocated to esp32/rom/crc.h in ESP-IDF v4.0
+// TODO: This will need to be platform specific in IDF v4.1 since this is
+// exposed in unique header paths for each supported platform. Detecting the
+// operating platform (ESP32, ESP32-S2, ESP32-S3, etc) can be done by checking
+// for the presence of one of the following defines:
+// CONFIG_IDF_TARGET_ESP32      -- ESP32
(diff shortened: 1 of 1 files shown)


atanisoft:
@RobertPHeller can you give this a test in your env as well?

RobertPHeller:
Compiles just fine -- no errors.

RobertPHeller:
At Mon, 23 Nov 2020 05:00:47 -0800 bakerstu/openmrn <reply@reply.github.com> wrote:

> 
> Thanks @RobertPHeller for a quick test. Question related to WiFi, do you use
> it very much? I'm asking as I'm considering a change in the setup to move
> from only compile time config to CDI based config with compile time
> constants being used as CDI defaults.

Right now, I am not using WiFi at all -- my boards have CAN transceivers and 
RJ45 connectors and my code does include the WiFi logic.

> 

-- 
heller@deepsoft.com       -- Webhosting Services

atanisoft:
Ok sounds like little or no impact on you in that case. I may ask you to test out a new CAN driver in the near future (it's part of my IDF v4 support rework but will be splitting that into smaller chunks for easier review)

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -48,17 +48,42 @@
...
 
-// ESP-IDF v4+ has a slightly different directory structure to previous
-// versions.
-#ifdef ESP_IDF_VERSION_MAJOR
-// ESP-IDF v4+
+// Starting in ESP-IDF v4.0 a few header files have been relocated so we need
+// to adjust the include paths accordingly. If the __has_include preprocessor
+// directive is defined we can use it to find the appropriate header files.
+// If it is not usable then we will default the older header filenames.
+#if defined __has_include
```
balazsracz, line 55: Isn't the syntax for this `#if defined(__has_include)` ?

balazsracz (approved):

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
atanisoft: It appears to work either way, updated it as you have posted above though.

PR #475, opened 2020-11-23
ESP32: Cleanup of compile time constants and ESP32 GPIO wrapper

atanisoft:
* Move the constexpr declarations from only being defined on ESP32 to be defined as constants that can be overriden.
* Introduce ESP32 GPIO wrapper that includes compile time validation of usages which prevents the usage of certain patterns which are known to be problematic (such as using GPIO pins connected to flash or attempting to enable pull-up on a pin that has a hardware pull-down R).
* Switch to using ESP-IDF provided define for Atomic.hxx if available.

### DIFF
--- a/arduino/OpenMRNLite.h
+++ b/arduino/OpenMRNLite.h
@@ -49,7 +49,7 @@
 #include "utils/logging.h"
 #include "utils/Uninitialized.hxx"
 
-#if defined(ESP32)
+#ifdef ESP32
 
 #include <esp_task.h>
 #include <esp_task_wdt.h>
@@ -63,18 +63,18 @@
 // dynamic CDI.xml generation support
 #define HAVE_FILESYSTEM 1
 
-#else
+#else // not ESP32
 
 // Include generic Arduino API compatible GPIO pin support.
 #include "freertos_drivers/arduino/ArduinoGpio.hxx"
 
-#endif // ESP32
+#endif // not ESP32
 
 #ifdef ARDUINO_ARCH_STM32
 
 #include "freertos_drivers/stm32/Stm32Can.hxx"
 
-#endif
+#endif // ARDUINO_ARCH_STM32
 
 namespace openmrn_arduino
 {
@@ -354,14 +354,9 @@ public:
         }
     }
 
-// For single threaded platforms (including ESP32-S2) do not expose the support
-// for spawning a seperate thread for the OpenMRN executor.
-// NOTE: The ESP32 is included here since it is only a single core platform and
-// starting an executor thread will likely prevent the "main" task from
-// executing, additionally there is very low free heap when PSRAM is not
-// available as part of the heap (only WROVER modules typically have PSRAM).
-#if !defined(OPENMRN_FEATURE_SINGLE_THREADED) && \
-    !defined(CONFIG_IDF_TARGET_ESP32S2)
+// For single threaded platforms do not expose the support for creating a
+// seperate thread for the OpenMRN executor.
+#ifndef OPENMRN_FEATURE_SINGLE_THREADED
 
     /// Entry point for the executor thread when @ref start_executor_thread is
     /// called with donate_current_thread set to false.
@@ -395,6 +390,8 @@ public:
     ///
     /// Note: On the ESP32 the watchdog timer is disabled for the PRO_CPU prior
     /// to starting the background task for the @ref Executor.
+    /// Note: For single-core ESP32 modules (ie: ESP32-S2) it may not be
+    /// advantageous to have a dedicated task for the OpenMRN Executor.
     void start_executor_thread()
     {
         haveExecutorThread_ = true;
@@ -421,7 +418,7 @@ public:
             config_arduino_openmrn_stack_size());
 #endif // ESP32
     }
-#endif // !OPENMRN_FEATURE_SINGLE_THREADED && !CONFIG_IDF_TARGET_ESP32S2
+#endif // !OPENMRN_FEATURE_SINGLE_THREADED
 
     /// Adds a serial port to the stack speaking the gridconnect protocol, for
     /// example to do a USB connection to a computer. This is the protocol that
@@ -519,16 +516,10 @@ private:
     /// Callback from the loop() method. Internally called.
     void run() override
     {
-// If we are running on a platform where we can have a dedicated executor
-// thread, check if it has been created and exit early if it has.
-#if !defined(OPENMRN_FEATURE_SINGLE_THREADED) && \
-    !defined(CONFIG_IDF_TARGET_ESP32S2)
-        if (haveExecutorThread_)
+        if (!haveExecutorThread_)
         {
-            return;
+            stack_->executor()->loop_some();
         }
-#endif // !OPENMRN_FEATURE_SINGLE_THREADED && !CONFIG_IDF_TARGET_ESP32S2
-        stack_->executor()->loop_some();
     }
 
     /// Storage space for the OpenLCB stack. Will be constructed in init().
@@ -537,11 +528,8 @@ private:
     /// List of objects we need to call in each loop iteration.
     vector<Executable *> loopMembers_{{this}};
 
-#if !defined(OPENMRN_FEATURE_SINGLE_THREADED) && \
-    !defined(CONFIG_IDF_TARGET_ESP32S2)
     /// True if there is a separate thread running the executor.
     bool haveExecutorThread_{false};
-#endif // !OPENMRN_FEATURE_SINGLE_THREADED && !CONFIG_IDF_TARGET_ESP32S2
 };
 
 } // namespace openmrn_arduino
--- a/arduino/OpenMRNLite.h
+++ b/arduino/OpenMRNLite.h
@@ -38,10 +38,12 @@
 
 #include <Arduino.h>
 
-#include "freertos_drivers/arduino/ArduinoGpio.hxx"
+#include "openmrn_features.h"
+
 #include "freertos_drivers/arduino/Can.hxx"
 #include "freertos_drivers/arduino/WifiDefs.hxx"
 #include "openlcb/SimpleStack.hxx"
+#include "utils/constants.hxx"
 #include "utils/FileUtils.hxx"
 #include "utils/GridConnectHub.hxx"
 #include "utils/logging.h"
@@ -52,28 +54,19 @@
 #include <esp_task.h>
 #include <esp_task_wdt.h>
 
-namespace openmrn_arduino {
-
-/// Default stack size to use for all OpenMRN tasks on the ESP32 platform.
-constexpr uint32_t OPENMRN_STACK_SIZE = 4096L;
-
-/// Default thread priority for any OpenMRN owned tasks on the ESP32 platform.
-/// ESP32 hardware CAN RX and TX tasks run at lower priority (-1 and -2 
-/// respectively) of this default priority to ensure timely consumption of CAN
-/// frames from the hardware driver.
-/// Note: This is set to one priority level lower than the TCP/IP task uses on
-/// the ESP32.
-constexpr UBaseType_t OPENMRN_TASK_PRIORITY = ESP_TASK_TCPIP_PRIO - 1;
-
-} // namespace openmrn_arduino
-
+#include "freertos_drivers/esp32/Esp32Gpio.hxx"
 #include "freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx"
 #include "freertos_drivers/esp32/Esp32HardwareSerialAdapter.hxx"
 #include "freertos_drivers/esp32/Esp32WiFiManager.hxx"
 
 // On the ESP32 we have persistent file system access so enable
 // dynamic CDI.xml generation support
-#define HAVE_FILESYSTEM
+#define HAVE_FILESYSTEM 1
+
+#else
+
+// Include generic Arduino API compatible GPIO pin support.
+#include "freertos_drivers/arduino/ArduinoGpio.hxx"
 
 #endif // ESP32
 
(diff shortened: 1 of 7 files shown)


arduino/OpenMRNLite.h
```diff
@@ -512,10 +519,16 @@ class OpenMRN : private Executable
     /// Callback from the loop() method. Internally called.
     void run() override
     {
-        if (!haveExecutorThread_)
+// If we are running on a platform where we can have a dedicated executor
+// thread, check if it has been created and exit early if it has.
+#if !defined(OPENMRN_FEATURE_SINGLE_THREADED) && \
```
balazsracz, line 524: I think this ifdef is unnecessary code complexity. See below.

arduino/OpenMRNLite.h
```diff
@@ -524,8 +537,11 @@ class OpenMRN : private Executable
     /// List of objects we need to call in each loop iteration.
     vector<Executable *> loopMembers_{{this}};
 
+#if !defined(OPENMRN_FEATURE_SINGLE_THREADED) && \
```
balazsracz, line 540: 1) 
consider adding an 
```
#else
   static constexpr const bool haveExecutorThread_ = false;
#endif
```
because this will make code optimized away.

2)
it would be beneficial to not have to repeat the same complex `#ifdef` in multiple places in the code (3 currently!). This even overrides my desire to have certain ordering of methods vs class variables. Consider moving this variable declaration up into the other ifdef.

src/freertos_drivers/esp32/Esp32Gpio.hxx
```diff
@@ -0,0 +1,467 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * \file Esp32Gpio.hxx
+ *
+ * Helper declarations for using GPIO pins via the ESP-IDF APIs.
+ *
+ * @author Mike Dunston
+ * @date 27 March 2020
```
balazsracz, line 32: check this date, is it true?

src/freertos_drivers/esp32/Esp32Gpio.hxx
```diff
@@ -0,0 +1,467 @@
...
+template <class Defs, bool PUEN, bool PDEN> struct GpioInputPin;
+
+/// Defines a GPIO output pin. Writes to this structure will change the output
+/// level of the pin. Reads will return the pin's current level.
+///
+/// The pin is set to output at initialization time, with the level defined by
+/// `SAFE_VALUE'.
+///
+/// Do not use this class directly. Use @ref GPIO_PIN instead.
+template <gpio_num_t PIN_NUM, bool INVERTED = false>
```
balazsracz, line 62: There is now an InvertedGpio template. Directly supporting inversion in the base template is no longer necessary. This could decrease code complexity in this file.

src/freertos_drivers/esp32/Esp32Gpio.hxx
```diff
@@ -0,0 +1,467 @@
...
+    static_assert(PIN_NUM != 11 && PIN_NUM != 16 && PIN_NUM != 17
+                , "Pin is reserved for flash usage.");
+#else
+    static_assert(!(PIN_NUM >= 6 && PIN_NUM <= 11)
+                , "Pin is reserved for flash usage.");
+#if defined(BOARD_HAS_PSRAM)
+    static_assert(PIN_NUM != 16 && PIN_NUM != 17
+                , "Pin is reserved for PSRAM usage.");
+#endif // BOARD_HAS_PSRAM
+#endif // ESP32_PICO
```
balazsracz, line 90: I find these a bit confusing, because really the block this is closing is the !PICO

src/freertos_drivers/esp32/Esp32Gpio.hxx
```diff
@@ -0,0 +1,467 @@
...
+    /// Sets the output state of the connected GPIO pin.
+    ///
+    /// @param new_state State to set the GPIO pin to.
+    void write(Value new_state) const override
+    {
+        if (INVERTED)
+        {
+            LOG(VERBOSE, "Esp32Gpio(%d) write %s", PIN_NUM,
+                new_state == Value::SET ? "CLR" : "SET");
+            ESP_ERROR_CHECK(gpio_set_level(PIN_NUM, new_state == Value::CLR));
```
balazsracz, line 102: `! new_state`

src/freertos_drivers/esp32/Esp32Gpio.hxx
```diff
@@ -0,0 +1,467 @@
...
+        ESP_ERROR_CHECK(gpio_set_level(PIN_NUM, SAFE_VALUE));
+    }
+
+    /// Sets the output pin @param value if true, output is set to HIGH, if
+    /// false, output is set to LOW.
+    static void set(bool value)
+    {
+        if (INVERT)
+        {
+            ESP_ERROR_CHECK(gpio_set_level(PIN_NUM, !value));
```
balazsracz, line 216: You have two implementations of GPIO here; one static and one with virtual functions and overrides.
This is not generally necessary, although some historic implementations may exist in two copies. Generally is is sufficient to have the one with static methods, then use GpioWrapper<>::instance() as the dynamic implementation.

I think the Stm32Gpio.hxx is a good example for simplification. I should make the TI and CC32xx follow the same pattern some day.

src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
```diff
@@ -94,10 +96,13 @@ public:
         ESP_ERROR_CHECK(can_driver_install(
             &can_general_config, &can_timing_config, &can_filter_config));
 
-        xTaskCreatePinnedToCore(rx_task, "ESP32-CAN RX", OPENMRN_STACK_SIZE,
-            this, RX_TASK_PRIORITY, &rxTaskHandle_, tskNO_AFFINITY);
-        xTaskCreatePinnedToCore(tx_task, "ESP32-CAN TX", OPENMRN_STACK_SIZE,
-            this, TX_TASK_PRIORITY, &txTaskHandle_, tskNO_AFFINITY);
+        xTaskCreate(rx_task, "CAN RX",
+            config_arduino_openmrn_stack_size(), this,
```
balazsracz, line 100: i think these should be different constants. These tasks almost always need a different amount of stack if you wantto be careful about RAM usage.

(feel free to ignore this if you intend to remove these tasks in the future).

src/utils/constants.cxx
```diff
@@ -163,4 +163,19 @@ DEFAULT_CONST(socket_listener_backlog, 1);
...
+#ifdef ESP32
+#include <esp_task.h>
+/// Default thread priority for any OpenMRN owned tasks on the ESP32 platform.
+/// Note: This is set to one priority level lower than the TCP/IP task uses on
+/// the ESP32.
+DEFAULT_CONST_DEFERRED(arduino_openmrn_task_priority, ESP_TASK_TCPIP_PRIO - 1);
+#else
+/// Default thread priority for any OpenMRN owned tasks on the Arduino
+/// platform.
+DEFAULT_CONST(arduino_openmrn_task_priority, 0xdefa01);
```
balazsracz, line 180: I'm not sure this will work for you. You pass it to xTaskCreate. In OpenMRN we have os_task_create which interprets this constant and sets it to the default (middle) priority level.

src/utils/constants.hxx
```diff
@@ -102,19 +107,47 @@
...
 #endif // native C
 
+/// Defines the default value of a constant using an externally defined value
+/// by using deferred expansion to work around pre-compiler limitations. Like
+/// @ref DEFAULT_CONST use this in a single .cxx file and make sure NOT to
+/// include the header that has the respective DECLARE_CONST macro.
+///
+/// @param var name of the constant.
+/// @param value is what the actual value should be.
+#define DEFAULT_CONST_DEFERRED(var, value) DEFAULT_CONST(var, value)
```
balazsracz, line 141: do we really need this complexity in the base header? It's somewhat unclear to me when and why would you want to use it ("deferred expansion" is not a particularly descriptive comment on a macro called "_DEFERRED").

Could this complexity move to the location where you actually need it?

Can you put a single DEFER macro into the value instead of duplicating every macro, like OVERRIDE_CONST(foo, DEFER(bar));

src/freertos_drivers/esp32/Esp32Gpio.hxx
atanisoft: Yes, I started on this file quite some time ago on the old WIP PR and have updated it many times over to get the static_asserts correct for the various chip combinations.

If you prefer, I can update it.

src/freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx
atanisoft: This class is planned to be removed with the merge of the TWAI, there are a couple other bugs in this implementation that went unnoticed for some time (counting rx/tx frames as an example)

src/utils/constants.cxx
atanisoft: Should we default this to -1 instead? I copied this as-is from OpenMRNLite.h.

src/utils/constants.hxx
atanisoft: The usage of this is in expanding of an externally defined value, such as:
```
OVERRIDE_CONST_DEFERRED(gridconnect_buffer_size, CONFIG_LWIP_TCP_MSS);
```

I'll see if I can do this with `OVERRIDE_CONST(gridconnect_buffer_size, DEFER(CONFIG_LWIP_TCP_MSS));` instead. If that works I'll simplify this to just have `DEFER(value)` instead.

src/freertos_drivers/esp32/Esp32Gpio.hxx
atanisoft: The complexity comes from the PICO-D4 allowing usage of pins 9 and 10 whereas all other variants restrict those pins to flash. I'll see if I can clarify this further in comments.

src/freertos_drivers/esp32/Esp32Gpio.hxx
atanisoft: I'll take a look at that.

src/freertos_drivers/esp32/Esp32Gpio.hxx
atanisoft: I'll take a look at Stm32Gpio and see if I can rework this to be closer to that.

src/utils/constants.hxx
balazsracz: another trick I've seen is to add an EMPTY() macro (which expands to nothing) and have this be present in the expansion. If that works, we could have that as the default behavior, i.e. modify DEFAULT_CONST and OVERRIDE_CONST directly.

atanisoft: That sounds like a better option than what I've implemented here. How does `CONSTANT_DEFER` sound for the extra macro defined in constants.hxx and used in the `DEFAULT_CONST` and `OVERRIDE_CONST` macros?

src/freertos_drivers/esp32/Esp32Gpio.hxx
atanisoft: I looked at switching to use `GpioWrapper<>::instance()` directly and it would pose a problem with using the instances with `MultiConfiguredPC`. The problem being `GpioWrapper::set_direction()` assuming that the underlying pin configuration can not be changed at runtime, that is true for some platforms but not necessarily so on the ESP32 (only certain pins have restrictions on them).

I'll look at removing the second implementation so that there is only one implementation (Esp32Gpio).

src/utils/constants.hxx
atanisoft: In testing this I found a problem with the inline expansion of `DEFER(value)`, specifically when `NEED_SIMPLE_CONST` is not defined. The `asm(...)` lines are using concatenation to inject the value into the generated asm code and that will prevent this from working as expected.

One option to fix this would be ensure that NEED_SIMPLE_CONST is used by the ESP32, another would be use it for all platforms (likely not ideal).

Thoughts?

balazsracz: I definitely don't want to use simple const for all platforms. Have you tried EMPTY() ?

atanisoft: The `asm(...)` expansion fails unfortunately:
```
.../src/utils/constants.cpp.obj.d -o esp-idf/OpenMRNLite/CMakeFiles/__idf_OpenMRNLite.dir/src/utils/constants.cpp.obj -c ../components/OpenMRNLite/src/utils/constants.cpp
/tmp/ccUyDoir.s: Assembler messages:
/tmp/ccUyDoir.s:129: Error: junk at end of line, first unrecognized character is `('
ninja: build stopped: subcommand failed.
```
The code used was:
```
DEFAULT_CONST(arduino_openmrn_task_priority, EMPTY(ESP_TASK_TCPIP_PRIO - 1));
```

balazsracz:
that's not actually what i meant. Something like this:
```
#define EMPTY()

#define DEFAULT_CONST(previous arguments) previous definition EMPTY()
```

but i don't think we have a guarantee that a numeric expression is an
acceptable argument to default const and override const.

Does your deferred code actually work on the esp32? Are you sure the
problem is with the macro evaluation and not the #include order?

atanisoft:
> Does your deferred code actually work on the esp32?

Yes, I've been testing it without issues on a couple projects.

The problem is how the pre-processor is expanding the `DEFAULT_CONST` inline with the `asm()` directives.

src/utils/constants.hxx
atanisoft: One more option, I can rename the two current ones to `DEFAULT_CONST_INLINE` and `OVERRIDE_CONST_INLINE` and rename the deferred variant to the current names. This way the ability to override with another compile time constants is available and using the two types of constant value definitions is also preserved.

atanisoft:
Closing for now and will resubmit as part of the arduino-esp32 2.0 fix up PR.

PR #476, opened 2020-11-23
WIP: ESP32: TWAI (CAN) driver supporting select()

atanisoft:
The existing `Esp32HardwareCan` code depends on using tasks to poll the ESP-IDF CAN (renamed to TWAI in ESP-IDF v4.2) which is not optimal. This code replicates the ESP-IDF v4.2 TWAI driver and has been augmented to support ESP-IDF v3.2 and above.

In my testing I've seen between 900-950 frames per second without dropping frames, this was done using Esp32CanLoadTest on the esp32 and load_test on Linux using an LCC-Buffer.

@RobertPHeller If you can please test this in your setup when you get a chance it would be appreciated. The current set of ESP32 PRs can be applied concurrently with this one and I'd suggest apply this one last.

### DIFF
--- a/arduino/OpenMRNLite.h
+++ b/arduino/OpenMRNLite.h
@@ -69,6 +69,7 @@ constexpr UBaseType_t OPENMRN_TASK_PRIORITY = ESP_TASK_TCPIP_PRIO - 1;
 
 #include "freertos_drivers/esp32/Esp32HardwareCanAdapter.hxx"
 #include "freertos_drivers/esp32/Esp32HardwareSerialAdapter.hxx"
+#include "freertos_drivers/esp32/Esp32Twai.hxx"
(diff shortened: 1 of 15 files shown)


RobertPHeller:
OK, it compiles and appears to work.  I don't have things set up enough to do any heavy testing, but  I was able to put the node on a small network on my bench and talk to it and load the CDI.

atanisoft:
Thanks @RobertPHeller for giving it a test run. I'm doing some extended testing on it using a variation on the ESP32 IO Board example and seeing some odd differences between arduino-esp32 (any version) and running under ESP-IDF (what arduino-esp32 is built upon). I haven't yet narrowed down the culprit but there is some sort of discrepancy in the configuration that I haven't uncovered yet.

PR #479, opened 2020-11-27
Adds a method to send out a global alias mapping enquiry to the bus.

balazsracz:
Updates the global AME handler to clear the remote alias cache. This cache
will be re-populated from the responses coming back from the bus after the
AME frame.

### DIFF
--- a/src/openlcb/IfCan.cxx
+++ b/src/openlcb/IfCan.cxx
@@ -718,13 +718,18 @@ void IfCan::send_global_alias_enquiry(Node *source)
         LOG_ERROR("Tried to send global AME witout a local alias.");
         return;
     }
-    auto *b = frame_write_flow()->alloc();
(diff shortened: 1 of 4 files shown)


src/openlcb/IfCan.cxx
```diff
@@ -702,6 +705,28 @@ void IfCan::set_alias_allocator(AliasAllocator *a)
...
+{
+    if (!source->is_initialized())
+    {
+        LOG_ERROR("Tried to send global AME from not initialzied node.");
+        return;
+    }
+    NodeAlias send_alias = local_aliases()->lookup(source->node_id());
+    if (!send_alias)
+    {
+        LOG_ERROR("Tried to send global AME witout a local alias.");
```
bakerstu, line 718: s/witout/without/

bakerstu (approved):

PR #480, opened 2020-11-28
Fix incorrect global AME replies.

balazsracz:
Fixes a bug where a global AME request was also getting replies for reserved
but unallocated aliases. These carry an invalid node ID from the local alias
cache and should not be sent to the bus.

Adds reserved aliases to the test cases and to test that no additional packets
show up.

### DIFF
--- a/src/openlcb/IfCan.cxx
+++ b/src/openlcb/IfCan.cxx
@@ -382,8 +382,9 @@ private:
     {
         while (nextIndex_ < if_can()->local_aliases()->size())
         {
-            if (if_can()->local_aliases()->retrieve(
-                    nextIndex_, nullptr, nullptr))
+            NodeID n;
+            if (if_can()->local_aliases()->retrieve(nextIndex_, &n, nullptr) &&
+                ((n >> (5 * 8)) != 0))
             {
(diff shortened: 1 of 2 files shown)


src/openlcb/IfCan.cxx
```diff
@@ -382,8 +382,9 @@ class AMEGlobalQueryHandler : public StateFlowBase,
     {
         while (nextIndex_ < if_can()->local_aliases()->size())
         {
-            if (if_can()->local_aliases()->retrieve(
-                    nextIndex_, nullptr, nullptr))
+            NodeID n;
+            if (if_can()->local_aliases()->retrieve(nextIndex_, &n, nullptr) &&
```
bakerstu, line 386: Is order of operations within an if statement guaranteed by the language standard? Otherwise, n could be used uninitialized.

bakerstu (approved):

src/openlcb/IfCan.cxx
balazsracz: Yes, an `&&` operator is always evaluated left to right and it has shortcut evaluation by definition of the C standard.

PR #497, opened 2020-12-30
Adds a version of the Queue where the scheduling algorithm can be tuned by weights.

balazsracz:


src/utils/ScheduledQueue.hxx
```diff
@@ -0,0 +1,218 @@
...
+                ret.index = i;
+                --numPending_;
+                return ret;
+            }
+        }
+        DIE("Unexpected nonempty queue");
+        return Result(0, 0);
+    }
+
+    /// @return the lock to use for the _locked() functions.
```
bakerstu, line 134: Why is this public? Is the user supposed to use this lock or is it only used inside of object members? If the user is expected to use this lock, can you add some comments to indicate where and why?

bakerstu (approved):
This looks like it could be extremely useful for polled RS-485 busses.

src/utils/ScheduledQueue.hxx
balazsracz: Added more verbose comments.
```
    /// The caller must acquire this lock before using any of the _locked()
    /// functions. If the caller needs to do many operations in quick
    /// succession, it might be faster to do them under a single lock,
    /// i.e. acquire lock() first, then call xxx_locked() repeatedly, then
    /// unlock.
    /// @return the lock to use for the _locked() functions.
```

This API also exists on Queue classes (it's also public there), and I wanted to keep a compatible API.

balazsracz:
> This looks like it could be extremely useful for polled RS-485 busses.

yes, see https://github.com/bakerstu/openmrn/pull/498

PR #499, opened 2021-01-01
Support eeprom update notification.

balazsracz:
Adds a weak callback to the eepromemulation driver, triggered
after each write is completed.

Shows with an example how to use this to automatically apply changes
that happened from JMRI configuration UI.


bakerstu (approved):
Yes, I need this. Thank you!

PR #504, opened 2021-01-03
Tune DCC end of packet behavior

balazsracz:
Adjusts the end of packet behavior of DCC packet generating:
- adds five extra one bits even if we are not generating a railcom cutout.
- allows appending a half-zero after the fifth bit even if no cutout was generated.

This ensures that switching from DCC to Marklin is possible with keeping a DCC cutout window.

### DIFF
--- a/src/freertos_drivers/ti/TivaDCC.hxx
+++ b/src/freertos_drivers/ti/TivaDCC.hxx
@@ -581,16 +581,18 @@ inline void TivaDCC<HW>::interrupt_handler()
             break;
         case DCC_NO_CUTOUT:
             current_bit = DCC_ONE;
-            if (++preamble_count >= 4) {
+            if (++preamble_count >= 5) {
+                if (HW::generate_railcom_halfzero())
+                {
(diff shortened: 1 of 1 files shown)


bakerstu (approved):
It looks like if there is a cutout, then no DCC bitstream (1 bits for example) is generated during the cutout. This needs to eventually get resolved. Please file an issue.

balazsracz:
> It looks like if there is a cutout, then no DCC bitstream (1 bits for example) is generated during the cutout. This needs to eventually get resolved. Please file an issue.

You observed this on a scope or by evaluating the code?
During the cutout we do not touch the period_timer, so it keeps running in PWM mode, emitting one bits.
In certain revisions of our hardware this caused glitching on the DRV8800 output, so we committed some code into hardware.hxx that turns the pin from PWM pin to GPIO pin to suppress the output signal.

bakerstu:
I apologize. I think I misread some inline comments regarding DCC bitstream during the RailCom cutout.

PR #506, opened 2021-01-10
Limit Timer

bakerstu:


src/utils/LimitTimer.hxx
```diff
@@ -0,0 +1,132 @@
...
+        {
+            return RESTART;
+        }
+    }
+
+    /// cooldown delay in msec
+    uint16_t updateDelayMsec_;
+
+    /// number of available tokens
+    uint8_t bucket_ ;
```
balazsracz, line 121: could you squeeze these into an `uint16_t bucket_ : 7;` , make bucketMax the same, and needUpdate with just one bit?
Then the data fields fit into a single 32 bit memory. I think we may have up to 62 of these in RAM.

src/utils/LimitTimer.hxx
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Destructor.
+    ~LimitTimer()
+    {
+        cancel();
+    }
+
+    /// Attempts to take a token out of the bucket.
+    /// @return true if the take is successful, false if there are no available
+    ///         tokens, in which case there will be an internal key event
```
balazsracz, line 67: replace 'internal key event' with 'callback'

src/utils/LimitTimer.hxx
```diff
@@ -0,0 +1,132 @@
...
+        HASSERT(callback);
+    }
+
+    /// Destructor.
+    ~LimitTimer()
+    {
+        cancel();
+    }
+
+    /// Attempts to take a token out of the bucket.
```
balazsracz, line 65: Please add a comment that this call must be made from the executor passed in on the constructor.

balazsracz (approved):

src/utils/LimitTimer.hxx
bakerstu: Fixed.

src/utils/LimitTimer.hxx
bakerstu: Fixed.

src/utils/LimitTimer.hxx
bakerstu: You are correct, there are up to 61 of these in one of our use cases (index 0 is broadcast, index 1 is reserved). I had considered this myself, and since you are are board, let's do it. Fixed.

Don't forget that we have an std::function, so we don't quite fit into 32-bits of memory.

PR #508, opened 2021-01-19
Adds a generalized shift operation to the Fixed16 class.

balazsracz:


bakerstu:
QQ: This appears to result in pulling in the double precision floating point library. Doesn't this kind of defeat the purpose of having a fixed point (Q-Math) library in the first place? BTW, CMSIS-DSP provides an architecture specific optimized implementation of a Q-math library.

https://arm-software.github.io/CMSIS_5/DSP/html/group__groupMath.html

balazsracz:
> QQ: This appears to result in pulling in the double precision floating point library. Doesn't this kind of defeat the purpose of having a fixed point (Q-Math) library in the first place? BTW, CMSIS-DSP provides an architecture specific optimized implementation of a Q-math library.
> 
> https://arm-software.github.io/CMSIS_5/DSP/html/group__groupMath.html

Only the test is using the double precision math library, the production code does not. The whole purpose of this operation on Fixed16 is to allow not pulling in the math library, which would take about 8 kbytes of code for the `pow()` operation.
In the production code I have inlined the coefficients as constexpr, and checked the generated assembly on an armv7m target. It compiled into a static array of 32-bit fixed point values, and used the already existing Fixed16::operator*= as expected. The total code size used by this implementation is <200 bytes and is inlineable.

I checked the q-math library, and did not find a fixed point type in it. It seems to be dealing with a lot of vector and matrix operations which are probably quite typical in DSP applications.

bakerstu:
I see, the fact that everything is constexpr ensures that the compiler does the floating point work and not the target at runtime.

bakerstu (approved):

PR #521, opened 2021-03-22
Broadcast Time Server Detection

bakerstu:


src/openlcb/BroadcastTimeClient.hxx
```diff
@@ -176,6 +184,9 @@ public:
 
         if (event->state == EventState::VALID)
         {
+            // We can only get here if there is a time server detected
+            serverDetected_ = true;
```
balazsracz, line 188: I think we should exclude the start/stop/query events and the set-time=X set-date=X events from this, and look only for the time-is or date-is events. I guess that is a simple check of `((event_id & 0x8000) == 0)`.
All of those other events could be produced by LCC buttons that are meant to remote control the clock.

balazsracz: Another option might be to only look for the date-is events for server detection (0x2 in the fourth nibble).

balazsracz (approved):

src/openlcb/BroadcastTimeClient.hxx
bakerstu: Fixed, went with the date mask option.

PR #523, opened 2021-04-04
Fix incorrect test for IDF version.

balazsracz:
The IDF version test is not compatible with the esp32 board support package 1.0.6 in Arduino.


atanisoft:
there will be a few more of these soon, the arduino-esp32 2.0 code branch is WIP and bumps the ESP-IDF version to 4.3 (currently master) with support for ESP32-S2 and ESP32-C3. I'll send in a few PRs once the arduino-esp32 2.0 code is available to test with on the ESP32-C3.

balazsracz:
This does not actually merge against master, because the respective code
has already been reworked there. So this PR is only here for cherrypicking
into openmrnlite branch.

PR #526, opened 2021-04-05
WiFi manager uplink disable

balazsracz:
Makes the wifi manager operate in one of three modes:
- uplink only
- hub+uplink
- hub only

Previously the uplink could not be disabled, which causes massive problems
in a hub-only application.


atanisoft (approved):
we might want to bump the version identifiers in the other esp32 examples.

balazsracz:
> we might want to bump the version identifiers in the other esp32 examples.

Did that for the other that uses wifi.

PR #530, opened 2021-04-05
Fixup in the ESP32 arduino examples

balazsracz:
Fixup in the ESP32 arduino examples:
- make sure they all have different Node IDs.
- make sure they all have different canonical versions.
This is useful when the user switches from one example to another to
correctly reset the settings.

Further fixes:
- removes unnecessary "ACDI version number" segment.
- ensures that all segments have a name, which is essential in the new JMRI UI.

### DIFF
--- a/arduino/examples/ESP32CanLoadTest/ESP32CanLoadTest.ino
+++ b/arduino/examples/ESP32CanLoadTest/ESP32CanLoadTest.ino
@@ -69,7 +69,7 @@
 
 /// This is the node id to assign to this device, this must be unique
 /// on the CAN bus.
-static constexpr uint64_t NODE_ID = UINT64_C(0x050101011823);
+static constexpr uint64_t NODE_ID = UINT64_C(0x05010101182b);
 
 #if defined(USE_WIFI)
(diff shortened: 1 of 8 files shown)


atanisoft (approved):

atanisoft:
@balazsracz will the ACDI section be automatically added or is this no longer going to be necessary going forward?

balazsracz:
> @balazsracz will the ACDI section be automatically added or is this no longer going to be necessary going forward?

The ACDI tag is still there. What I removed is the "set the ACDI version number here" section. The correct value is automatically done in create_config_file_if_needed(), so the user does not need to ever interact with this.

atanisoft:
Ok great, I missed that it was a user provided ACDI version info segment. Has JMRI been updated with the new version in a dev release or still hosted on your site?

balazsracz:
> Ok great, I missed that it was a user provided ACDI version info segment. Has JMRI been updated with the new version in a dev release or still hosted on your site?

The latest JMRI test release (4.23.3 i think) has the new UI changes.

PR #532, opened 2021-04-06
Event Identify Global Protocol Helper

bakerstu:


src/openlcb/EventIdentifyGlobal.hxx
```diff
@@ -0,0 +1,153 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * @file EventIdentifyGlobal.hxx
+ *
+ * A complete OpenLCB stack for use in straightforward OpenLCB nodes.
```
atanisoft, line 29: This doesn't look quite correct for this file

bakerstu: Oops, it isn't. Fixed.

src/openlcb/EventIdentifyGlobal.hxx
```diff
@@ -0,0 +1,153 @@
...
+    {
+        AtomicHolder h(this);
+        aborted_ = true;
+    }
+
+    /// Entry/reset point into state machine.
+    Action entry()
+    {
+        // get a pseudo random number between 1.500 and 2.011 seconds
+        long long timeout_msec = 1500 + (node_->node_id() & 0x1FF);
```
balazsracz, line 100: we should try to hash the node ID to do the randomness. The reason for it is that some manufacturers sell hardware in sequential order and mixing bits would be necessary to spread that difference to higher bits of the random selection.
I propose mixing bits with simple arithmetics like this:

```
// This code mixes the bits in the node ID to a hash value in 0 to 511.
uint64_t h = node_->node_id() * 0x1cf9a66d;
uint32_t hh = h ^ (h >> 32);
hh = hh ^ (hh >> 12) ^ (hh >> 24);
timeout_msec = 1500 + (hh & 511);
```

balazsracz (approved):

src/openlcb/EventIdentifyGlobal.hxx
bakerstu: Updated/fixed.

PR #534, opened 2021-04-10
Adds GitHub Action that builds the arduino examples.

balazsracz:
This Action works as a CI ensuring that the Arduino OpenMRNLite library and its contributed examples are building correctly.

The Action triggers automatically upon every pull request, and builds all ESP32*.ino and Stm32*.ino sketches using arduino-cli. The OpenMRNLite library and the example sketches are always taken from HEAD.

An example run is visible here:
https://github.com/balazsracz/openmrn/runs/2314541629?check_suite_focus=true


.github/workflows/ArduinoBuild.yml
```diff
@@ -0,0 +1,49 @@
...
+        platform-url: https://raw.githubusercontent.com/stm32duino/BoardManagerFiles/master/STM32/package_stm_index.json
+        arduino-board-fqbn: STM32:stm32:Nucleo_144:pnum=NUCLEO_F767ZI,upload_method=MassStorage,xserial=generic,usb=CDCgen,xusb=FS,opt=osstd,rtlib=nano
+        sketch-names: Stm32*.ino
+        build-properties: '{ "All": "-DHAL_CAN_MODULE_ENABLED" }'
+        debug-compile: true
+        
+    - name: Compile all ESP32 examples
+      uses: ArminJo/arduino-test-compile@v3.0.0
+      with:
+        platform-url: https://dl.espressif.com/dl/package_esp32_index.json
```
atanisoft, line 44: Take a look at the most recent 1.0.6 release as the json file moved to GitHub IIRC.

atanisoft (approved):

.github/workflows/ArduinoBuild.yml
balazsracz: done

PR #539, opened 2021-05-04
Adds drain ioctl.

balazsracz:
Adds an ioctl to the CC32xx UART driver that allows getting a notification
when transmission has completed.


bakerstu (approved):
Not really a good way to do something similar in Linux. I think for NCE bus we relied on a timer set based on the amount of data recently written.

balazsracz:
> Not really a good way to do something similar in Linux. I think for NCE bus we relied on a timer set based on the amount of data recently written.

yeah I was not a fan of trying to second guess how long the transmission takes.
Under linux there is only the tcdrain(); function, but it is synchronous. So we would need to start a new thread for running it. That's doable when we have no memory concerns.

PR #541, opened 2021-05-09
Console buffer fix

RobertPHeller:
I discovered a buffer indexing problem in Console.cxx.  When things get busy, the start of the line (pos) might not be zero, which the scanning code unconditionally assumes.  I fixed this by capturing the value of pos at the start of Console::Session::process_read() and then use it as the start value for i in the line scanning loop.  This cures the problem.


RobertPHeller:
Is there some reason this pull request is still waiting?

src/console/Console.cxx
```diff
@@ -299,8 +302,8 @@ StateFlowBase::Action Console::Session::process_read()
             /* parse the line input into individual arguments */
             unsigned argc = 0;
             char last = '\0';
-
-            for (size_t i = 0; i < pos; ++i)
+            
+            for (size_t i = start; i < pos; ++i)
```
balazsracz, line 306: I don't think this change is correct code.
When we find an end of line, we have to parse the entire line, just just the bytes that came since the last select call.
I don't know what is the issue you are trying to fix, but this does not seem like a good fix for it. A unittest that produces the problem first would be helpful.

RobertPHeller:
I was encountering problems with my BeagleBoard Command Station code
(https://github.com/RobertPHeller/RPi-RRCircuits/tree/master/BBBCommandStationOpenMRN),
where a fresh command line would land in the buffer someplace other than the
beginning of the buffer:

from Console.hxx, in class Console::Session : public StateFlowBase

        StateFlowBase::Action entry()                                  
        {                                                              
            return read_single(&selectHelper, fdIn, line + pos,        
                                           line_size - pos, 
                                           STATE(process_read));  
        }
                                                                               
       
Data is read starting at line + pos: *usually* when things are quiet, pos is
zero (the parser resets pos to zero when the buffer is used up), but when
things get busy, pos might not be zero. My fix handles that case. The reader
code appears to be properly using the line buffer as a ring buffer, except
that the parse code is not (opps!) properly honoring the ring buffer logic. I
guess in a quiet static situation, there would not be any need for a ring
buffer -- pos gets reset to zero when the buffer is used up (the parser checks
for that). But when things get busy, the ring buffer logic kicks in
(Console::Session::entry() is invoked when pos is non-zero), except that the
parser is (incorrectly?) hardwired not to use the ring buffer logic, because
the parse loop always starts at the beginning of the buffer instead of using
pos as the start-of-line position (0 is hardcoded at the top of the for loop).

My command station code is meant to be "split" between two SBCs communicating
over a Tcp/Ip socket: a Beagle Board (a PocketBeagle) for the low-level (using
the PRUs to real-time bit-bash the DCC signals) and a RPi for the high-end
(typically a TCL/TK GUI program). The two programms communicate via a
simple ASCII text "command" protocol: the RPi sends simple commands and
the Beagle Board either does something or returns status info. The problem
happens when the RPi starts sending commands fast enough to cause the ring
buffer logic to kick in.

With a human entering commands over a serial console, the ring buffer logic is 
unlikely to ever kick in, but with one commputer talking to another over a 
network socket, it is very possible for the ring buffer logic to kick in.  I 
guess that is kind of an edge case?


At Sun, 21 Aug 2022 09:15:41 -0700 bakerstu/openmrn ***@***.***> wrote:

> 
> @balazsracz commented on this pull request.
> 
> 
> 
> > @@ -299,8 +302,8 @@ StateFlowBase::Action Console::Session::process_read()
>              /* parse the line input into individual arguments */
>              unsigned argc = 0;
>              char last = '\0';
> -
> -            for (size_t i = 0; i < pos; ++i)
> +            
> +            for (size_t i = start; i < pos; ++i)
> 
> I don't think this change is correct code.
> When we find an end of line, we have to parse the entire line, just just the bytes that came since the last select call.
> I don't know what is the issue you are trying to fix, but this does not seem like a good fix for it. A unittest that produces the problem first would be helpful.
> 

-- 
***@***.***       -- Webhosting Services

balazsracz:
Were you sending more than one command (with \n separated) at a time, or at
close interval?
The buffer is not meant as a ring buffer, the code sets pos=0 after picking
up a line. But I don't see any handling of having more than one command in
the socket. It seems to assume that once we found a \n there is no more
input data.

RobertPHeller:
OK, it seems that this fix is not necessary.  I'm closing this pull request.

RobertPHeller:
It sends one command at a time, but possibly at a close interval.  OK, then 
there might be some kind of race condition or maybe by tcl code is not 
properly flushing its I/O buffer.  Something to check.

At Sun, 21 Aug 2022 14:35:38 -0700 bakerstu/openmrn ***@***.***> wrote:

> 
> Were you sending more than one command (with \n separated) at a time, or at
> close interval?
> The buffer is not meant as a ring buffer, the code sets pos=0 after picking
> up a line. But I don't see any handling of having more than one command in
> the socket. It seems to assume that once we found a \n there is no more
> input data.
> 
> 

-- 
***@***.***       -- Webhosting Services

balazsracz:
To be fair, I do think that the current code in openmrn is incorrect. I
believe it does not correctly reset the parsing state to the beginning of a
line.

This bug is masked when you are typing, because there is always at least
one cycle of the code running where it sees only the end-of-line character
but nothing from the next line.

All unit tests are lucky, because they only test sending one command at a
time.

However, when you send traffic from a program, there is no reason to
believe that commands are only sent one at a time. And that should work
just fine. But it likely does not. So there *is* a bug.

PR #543, opened 2021-05-29
Fixes rounding problem in DCC speed step representation.

### LINKED ISSUE #356: Velocity::get_dcc_128 might be buggy
We have some convenience conversion routines in the Velocity class. I'm afraid these might be buggy; the code and the test cases need to be reviewed.
It seems to me that the test case covering this code itself might have an overflow error.

balazsracz:
Adds tests for exercising all 126 speed steps.

Fixes #356.


atanisoft:
Looks like this might fix https://github.com/bakerstu/openmrn/issues/356

bakerstu (approved):

PR #544, opened 2021-05-30
Allow Esp32WiFiManager to work with any subclass of SimpleCanStackBase.

toholio:
This allows the WiFi manager to be used in cases such as when a stack type implements a different node type. E.g. SimpleTrainCanStack.

### DIFF
--- a/src/freertos_drivers/esp32/Esp32WiFiManager.cxx
+++ b/src/freertos_drivers/esp32/Esp32WiFiManager.cxx
@@ -86,7 +86,7 @@
 #endif // defined __has_include
 
 using openlcb::NodeID;
-using openlcb::SimpleCanStack;
+using openlcb::SimpleCanStackBase;
 using openlcb::TcpAutoAddress;
 using openlcb::TcpClientConfig;
 using openlcb::TcpClientDefaultParams;
@@ -304,7 +304,7 @@ private:
 // WiFi connection, mDNS system and the hostname of the ESP32.
 Esp32WiFiManager::Esp32WiFiManager(const char *ssid
                                  , const char *password
-                                 , SimpleCanStack *stack
+                                 , SimpleCanStackBase *stack
                                  , const WiFiConfiguration &cfg
                                  , const char *hostname_prefix
                                  , wifi_mode_t wifi_mode
@@ -357,7 +357,7 @@ Esp32WiFiManager::Esp32WiFiManager(const char *ssid
 // With this constructor being used, it will be the responsibility of the
 // application to manage the WiFi and mDNS systems.
 Esp32WiFiManager::Esp32WiFiManager(
-    SimpleCanStack *stack, const WiFiConfiguration &cfg)
+    SimpleCanStackBase *stack, const WiFiConfiguration &cfg)
     : DefaultConfigUpdateListener()
     , cfg_(cfg)
     , manageWiFi_(false)
--- a/src/freertos_drivers/esp32/Esp32WiFiManager.hxx
+++ b/src/freertos_drivers/esp32/Esp32WiFiManager.hxx
@@ -84,7 +84,7 @@ public:
     /// @param ssid is the WiFi AP to connect to. Must stay alive forever.
     /// @param password is the password for the WiFi AP being connected
     /// to. Must stay alive forever.
-    /// @param stack is the SimpleCanStack for this node. Must stay alive
+    /// @param stack is the SimpleCanStackBase for this node. Must stay alive
     /// forever.
     /// @param cfg is the WiFiConfiguration instance used for this node. This
     /// will be monitored for changes and the WiFi behavior altered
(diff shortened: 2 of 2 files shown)


atanisoft (approved):
As part of https://github.com/bakerstu/openmrn/pull/536 I'm shifting this over to use `openlcb::SimpleStackBase`. But in the meantime this looks fine.

How do you intend on using this functionality though?

toholio:
@atanisoft I’ve been experimenting with a native LCC WiFi train mode for controlling one of my live steam locos.

This is change something I ran across earlier on when getting up and running, where I wanted the only node created to be the Train.

I have ended up writing my own stack class rather than using SimpleTrainCanStack since I need to return variable FDI data. I might go back and create an abstract base class for Train Can Stacks so that basic derived classes only need to provide an appropriate FDI MemorySpace instance.

atanisoft:
@toholio I'd be interested to see your changes as I have a few designs in mind for mobile decoders built on top of OpenMRN using a variant of the ESP32 SoCs.

For the FDI portion, you can use the standard SimpleCanStack/SimpleCanTrainStack and add on top of it the necessary memory spaces. You can see one example of this in [this](https://github.com/atanisoft/ESP32CommandStation/tree/master/components/LCCTrainSearchProtocol) code which I used from @balazsracz and modified for the CS usage.

I've also incorporated your update into #536 so it will be preserved when that PR has been merged (still a few WIP areas and pending arduino-esp32 2.0 release)

toholio:
@atanisoft I'll take a look at that code. Thanks.

I'll definitely be sharing my work when it's been tidied up. I've aimed for live steam control at first because I don't have the circuit design skills for motor control. Driving servos is straight forward. I'm working exclusively in G scale and have the luxury of space, so might try using a VNH5019 breakout board or similar to move a battery powered loco.

My current struggle is converting my project from the arduino framework to espidf. There is only have a small amount of code to rewrite that uses Arduino libraries but can't get OpenMRNLite to compile at all in my new workspace. I'm completely stumbling here on trivial things so will probably not bother converting for the moment in favour of getting the actual mobile decoder to a runnable state. 

Your command station code has been very helpful! Just a bit overwhelming to learn from when I'm trying to do much more trivial things.

atanisoft:
If you are going for esp-idf, take a look at my OpenMRNIDF repository. It is a ready to use component and I have a few examples of usage, I'd recommend my Esp32C3OlcbIO repository as reference.

atanisoft:
Also which libraries are you stuck on? I may have a version available, including for servos.

toholio:
Ah, the Esp32C3OlcbIO and esp32olcbhub look like just the sort of examples I need. Thanks for those!

It's not the libraries I've been stuck on so far but I'm sure that's coming next. The difficulty has been in basic understanding of the esp-idf workspace layout and toolchain. I was immediately knee deep in CMakeList files and unsure of what pulls together what else and how to even get OpenMRNLite into the project correctly. Hard to fight the uncertainty when starting with something new.

The reason for switching at all is because of trouble getting nlohmann_json to compile (needs c++11) and also trouble getting the AutoSyncFileFlow to persist data to an SD card correctly. (unsure about what is happening here but the fsync call has no effect). Both of those things can wait for now because they're only really needed for the next step of adding I2S sound support to functions (and maybe chuff input for electric locos). Seemed like a good time to reassess the initial decision to use Arduino.

The examples you gave above have already clarified some things from just a superficial look! It's worktime here in Melbourne so I'll have to dive in properly later.

atanisoft:
If you can describe the issues you are seeing with the SD persistence issue I can take a look at it. I've not had issues with it.

Nlohmann' json library should work fine without modifications since gnu++11 is the default in esp-idf and arduino-esp32, I override this in the above repositories for gnu++14 instead though to enable a few extra features.

It sounds like we are working on similar hardware and software for a mobile decoder. I have a hardware design that should fit into an N scale loco that includes pads for a speaker. Drop me an email and we can collaborate on the firmware and possibly a PCB that can fit both out needs

toholio:
I'll send you an email about the mobile decoder project. I'd be very happy to collaborate.

For the SD persistence issue, the extent of what I'm seeing is that after starting the SD handling with:
`SD.begin()`
and then in the stack manager doing something like:
```c++
int fd = stack_->create_config_file_if_needed( cfg.seg().internal_config(), openlcb::CANONICAL_VERSION, openlcb::CONFIG_FILE_SIZE );
configAutoSync_ = new AutoSyncFileFlow( stack_->service(), fd );
```

The `openlcb_config` file is created on the SD card but is always empty. I put logging into `AutoSyncFileFlow` to confirm that it's triggering but haven't gotten as far as connecting a JTAG debugger to go deeper.

Please let me know if I should report that somewhere other than in this comment chain.

atanisoft:
JTAG and SD will conflict as the pins overlap by default.

The only time I've seen a case like this is when the stack is blocked or not running. We can discuss over email and see if we can narrow it down further.

PR #545, opened 2021-06-22
Slow down the GPIO writes a bit.

balazsracz:


src/freertos_drivers/ti/CC3200GPIO.hxx
```diff
@@ -171,6 +171,7 @@ public:
         volatile uint8_t *ptr = reinterpret_cast<uint8_t *>(
             GPIO_BASE + (((unsigned)GPIO_PIN) << 2));
         *ptr = value ? 0xff : 0;
+        __asm__ volatile("\t nop\n");
```
bakerstu, line 174: Can you add a comment as to why this nop is required?

bakerstu (approved):

src/freertos_drivers/ti/CC3200GPIO.hxx
balazsracz: done

PR #546, opened 2021-07-04
Adds CRC8 software implementations

balazsracz:
Adds implementation for computing and checking CRC-8 according to the Dallas/Maxim specification.


src/utils/Crc.cxxtest
```diff
@@ -30,11 +32,71 @@ TEST(CrcIbmTest, Example3) {
...
+TEST(Crc3Test, Example)
+{
+    uint16_t data[3];
+    crc3_crc16_ibm("12345678", 8, data);
+    EXPECT_EQ(0x3c9d, data[0]);
+    EXPECT_EQ(0x75a8, data[1]);
+    EXPECT_EQ(0x0459, data[2]);
+}
+
+TEST(Crc8Test, Example) {
```
bakerstu, line 44: Did you use well known test vectors? Can you add a comment about where we can go to validate the test vectors?

bakerstu (approved):

src/utils/Crc.cxxtest
balazsracz: Added this:
```
    // This test vector comes from the RCN-218 document by RailCommunity. It is
    // the same as an example message from the BiDiB examples page:
    // http://www.bidib.org/support/example1_e.html
```

PR #553, opened 2021-07-11
Adds routines to encode railcom data for a sender.

balazsracz:
- Adds 6-to-8 encode table
- Adds helper functions to perform encoding
- Adds unittests that validate that encoding and decoding are correctly inverse of each other.

Fixes contants:
- Aligns with RCN-217 on the two different ACK values
- Adds the newly appointed NACK value from RCN-217

Refactoring in the RailCom header:
- Makes RailcomDefs to a struct from a namespace. This is aligned better with the style of the rest of the codebase.

### DIFF
--- a/src/dcc/RailCom.hxx
+++ b/src/dcc/RailCom.hxx
@@ -79,33 +79,94 @@ struct Feedback : public DCCFeedback
 /// Formats a dcc::Feedback message into a debug string.
 std::string railcom_debug(const Feedback& fb);
 
-/// Special constant values returned by the @ref railcom_decode[] array.
-namespace RailcomDefs
-{
-    /// invalid value (not conforming to the 4bit weighting requirement)
-    static const uint8_t INV = 0xff;
-    /// Railcom ACK; the decoder received the message ok. NOTE: some early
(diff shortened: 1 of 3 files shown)


src/dcc/RailCom.hxx
```diff
@@ -79,33 +79,94 @@ struct Feedback : public DCCFeedback
...
-
 /** Table for 8-to-6 decoding of railcom data. This table can be indexed by the
  * 8-bit value read from the railcom channel, and the return value will be
  * either a 6-bit number, or one of the constants in @ref RailcomDefs. If the
  * value is invalid, the INV constant is returned. */
 extern const uint8_t railcom_decode[256];
+/// Table for 6-to-8 encoding of railcom data. The table can be indexed by a
+/// 6-bit value that is the semantic content of a railcom byte, and returns the
+/// matching 8-bit value to put out on the UART. This table only contains the
+/// standard codes, for the special codes like ACK use RailcomDefs above.
```
atanisoft, line 90: Below?

balazsracz: fixed

PR #555, opened 2021-07-16
Reduce the number of times the config file is opened in SimpleStack

toholio:
Reuse the file descriptor for the config file that was opened by the ConfigUpdateFlow.

### DIFF
--- a/src/openlcb/ConfigUpdateFlow.hxx
+++ b/src/openlcb/ConfigUpdateFlow.hxx
@@ -71,14 +71,22 @@ public:
     {
     }
 
-    /// Must be called once before calling anything else. Returns the file
-    /// descriptor.
+    /// Must be called once (only) before calling anything else. Returns the
+    /// file descriptor.
     int open_file(const char *path);
     /// Asynchronously invokes all update listeners with the config FD.
     void init_flow();
     /// Synchronously invokes all update listeners to factory reset.
     void factory_reset();
 
+    /// @return the file descriptor of the configuration file, or -1 if the
+    /// configuration file has not yet been opened.
+    int get_fd()
+    {
+        return fd_;
+    }
+
+#ifdef GTEST
     void TEST_set_fd(int fd)
     {
         fd_ = fd;
@@ -86,6 +94,7 @@ public:
     bool TEST_is_terminated() {
         return is_terminated();
     }
+#endif // GTEST
 
     void trigger_update() override
     {
--- a/src/openlcb/SimpleStack.cxx
+++ b/src/openlcb/SimpleStack.cxx
@@ -103,8 +103,13 @@ void SimpleStackBase::start_stack(bool delay_start)
 {
 #if OPENMRN_HAVE_POSIX_FD
     // Opens the eeprom file and sends configuration update commands to all
-    // listeners.
-    configUpdateFlow_.open_file(CONFIG_FILENAME);
+    // listeners. We must only call ConfigUpdateFlow::open_file() once and it
+    // may have been done by an earlier call to create_config_file_if_needed()
+    // or check_version_and_factory_reset().
+    if (configUpdateFlow_.get_fd() < 0)
+    {
+        configUpdateFlow_.open_file(CONFIG_FILENAME);
+    }
     configUpdateFlow_.init_flow();
 #endif // have posix fd
 
@@ -139,12 +144,12 @@ void SimpleStackBase::default_start_node()
 #if OPENMRN_HAVE_POSIX_FD 
     {
         auto *space = new FileMemorySpace(
-            SNIP_DYNAMIC_FILENAME, sizeof(SimpleNodeDynamicValues));
+            configUpdateFlow_.get_fd(), sizeof(SimpleNodeDynamicValues));
         memoryConfigHandler_.registry()->insert(
             node(), MemoryConfigDefs::SPACE_ACDI_USR, space);
         additionalComponents_.emplace_back(space);
     }
-#endif // NOT ARDUINO, YES ESP32
+#endif // OPENMRN_HAVE_POSIX_FD
     size_t cdi_size = strlen(CDI_DATA);
     if (cdi_size > 0)
     {
@@ -157,12 +162,13 @@ void SimpleStackBase::default_start_node()
 #if OPENMRN_HAVE_POSIX_FD
     if (CONFIG_FILENAME != nullptr)
     {
-        auto *space = new FileMemorySpace(CONFIG_FILENAME, CONFIG_FILE_SIZE);
+        auto *space =
+            new FileMemorySpace(configUpdateFlow_.get_fd(), CONFIG_FILE_SIZE);
         memory_config_handler()->registry()->insert(
             node(), openlcb::MemoryConfigDefs::SPACE_CONFIG, space);
         additionalComponents_.emplace_back(space);
     }
-#endif // NOT ARDUINO, YES ESP32
+#endif // OPENMRN_HAVE_POSIX_FD
 }
 
 SimpleTrainCanStack::SimpleTrainCanStack(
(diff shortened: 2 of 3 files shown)


src/openlcb/SimpleStack.cxx
```diff
@@ -138,8 +138,9 @@ void SimpleStackBase::default_start_node()
     }
 #if OPENMRN_HAVE_POSIX_FD 
     {
-        auto *space = new FileMemorySpace(
-            SNIP_DYNAMIC_FILENAME, sizeof(SimpleNodeDynamicValues));
+        auto *space =
+            new FileMemorySpace(configUpdateFlow_.open_file(CONFIG_FILENAME),
```
balazsracz, line 142: I think this is confusing because it gives the appearance that we open the config file more than once.
We should instead add a get_fd() API to ConfigUpdateFlow. We need to check the returned value for being >= 0.

We should probably also prevent more than one call to ConfigUpdateFlow::open_file because it seems to ignore the  filename in that case.

atanisoft: is there an equiv of "ensure_open()" on ConfigUpdateFlow? Perhaps we can add that and have it return the fd as well as open if needed.

atanisoft: Perhaps it would also make sense to rework this area slightly further to move the CONFIG_FILENAME/SNIP_DYNAMIC_FILENAME into the constructor for ConfigUpdateFlow and have just the call to ensure_open() (or get_fd()) that returns the active FD for the underlying file.

balazsracz: We should not do any file operations (including ::open) in constructors. It makes it unnecessarily complicated to set up applications because the order in which static constructors are executed is undeterministic.

I don't know what "ensure_open" is. Since this ensure_open or get_fd would be called in application initialization and not on demand, the sequence of calls is very deterministic. We don't need to prepare for calls of different sequence. Applications always start with opening or resetting their config file. We also need to make sure not to try to open the file before the application had actually run that code.

toholio: I have been experimenting with this and the code might be cleanest with the suggestion of passing the filename to ConfigUpdateFlow's constructor. The naming ensure_open() makes the functionality clear. ConfigUpdateFlow::open_file() could check the filename matches in the event that it does get called twice.

toholio: @balazsracz I had understood the suggestion as the filename would be passed to the constructor and stored but the file would not be opened until requested. Is that poor form?

atanisoft: > We should not do any file operations (including ::open) in constructors

I agree, I was implying that only the filename would be passed in for future use by `ensure_open`, `get_fd`, `open`, etc..

At the point that global constructors are called it is not guaranteed that all required dependencies will be available (in the case of esp32 this would be VFS, SPIFFS, etc).

toholio: Actually, doing it this way means also passing the configuration filename into the constructor for SimpleStackBase and therefore the classes that inherit from it. Perhaps that isn't so bad but at first glance it seems a bit messy.

I'll try some options in the morning.

atanisoft: `CONFIG_FILENAME` is already a "global" in the openlcb namespace, it will not need to be passed into SimpleStackBase. You don't pass it in today so you won't need to by shifting the reference to a constructor call.

toholio: Ah, right you are. There's a lesson for me about working on things late at night.

I have updated the pull-request with the config file path now a parameter to ConfigUpdateFlow's constructor. I have not tested this properly yet but it shows how it might look.

balazsracz: I'm not supportive of this direction. We are basically copying data from .rodata to RAM in order to use only once and then never free it. This needs to run in extremely memory constrained environments.

I don't see a reason why the configupdateflow should have the filename outside of the local context of the open_file. I don't see why this open would have to be called more than once. It is part of the initialization of the stack, we don't have any ambiguity about where those calls happen.

toholio: Fair points @balazsracz and I’m happy to change it back if that’s the preference. Easy to keep open_file and add get_fd. 

I’ll take a look again in the morning.

balazsracz: sounds good, thanks!

src/openlcb/ConfigUpdateFlow.hxx
```diff
@@ -71,14 +71,19 @@ public:
...
-    /// descriptor.
+    /// Must be called once (only) before calling anything else. Returns the
+    /// file descriptor.
     int open_file(const char *path);
     /// Asynchronously invokes all update listeners with the config FD.
     void init_flow();
     /// Synchronously invokes all update listeners to factory reset.
     void factory_reset();
 
+    int get_fd()
```
balazsracz, line 82:
```
// @return the file descriptor of the configuration file, or -1 if the configuration file has not yet been opened.
```

src/openlcb/SimpleStack.cxx
```diff
@@ -223,24 +229,28 @@ int SimpleStackBase::create_config_file_if_needed(const InternalConfigData &cfg,
     struct stat statbuf;
     bool reset = false;
     bool extend = false;
-    int fd = ::open(CONFIG_FILENAME, O_RDONLY);
+    int fd = configUpdateFlow_.get_fd();
```
balazsracz, line 232: i think we should rather assert that get_fd() < 0.
create_config_file_if_needed is always the first command in appl_main() when we have an app that operates on a filesystem.

It is not necessary to support calling create_config_file_if_needed after a config file is already open.

balazsracz (commented):
Looks good in general, just some comments.

src/openlcb/SimpleStack.cxx
toholio: No problem tidying that up. I'll push it and the `get_fd()` comment through shortly.

I had wondered about whether testing for the file being opened in `SimpleStackBase::create_config_file_if_needed()` made sense as I wrote it. Similar in `SimpleStackBase::start_stack()` but it seems like we can't be completely certain of the order in that case, so it does the test.

src/openlcb/ConfigUpdateFlow.hxx
```diff
@@ -71,14 +71,21 @@ public:
...
     void factory_reset();
 
+    /// @return the file descriptor of the configuration file, or -1 if the
+    /// configuration file has not yet been opened.
+    int get_fd()
+    {
+        return fd_;
+    }
+
     void TEST_set_fd(int fd)
```
atanisoft, line 89: Please wrap this method (and TEST_is_terminated) with:
```
#ifdef GTEST
...
#endif // GTEST
```

src/openlcb/SimpleStack.cxx
```diff
@@ -157,7 +162,8 @@ void SimpleStackBase::default_start_node()
 #if OPENMRN_HAVE_POSIX_FD
     if (CONFIG_FILENAME != nullptr)
     {
-        auto *space = new FileMemorySpace(CONFIG_FILENAME, CONFIG_FILE_SIZE);
+        auto *space =
+            new FileMemorySpace(configUpdateFlow_.get_fd(), CONFIG_FILE_SIZE);
         memory_config_handler()->registry()->insert(
             node(), openlcb::MemoryConfigDefs::SPACE_CONFIG, space);
         additionalComponents_.emplace_back(space);
```
atanisoft, line 169: Please correct the `#endif` line below to have `OPENMRN_HAVE_POSIX_FD` instead of `NOT ARDUINO, YES ESP32` since we moved to a new define.

atanisoft (approved):

balazsracz (approved):

PR #562, opened 2021-08-12
Adds implementation for railcom messages related to logon

balazsracz:
- Adds constants for the automatic logon railcom messages
- Adds helper functions in the RailcomDefs to generate the S-9.2.1.1 / RCN-218 feedback messages.
- Adds a processor for a railcom hub that recognizes these feedback messages and checks for correctness.

### DIFF
--- a/src/dcc/LogonFeedback.hxx
+++ b/src/dcc/LogonFeedback.hxx
@@ -35,6 +35,7 @@
 #define _DCC_LOGONFEEDBACK_HXX_
 
 #include "dcc/RailcomHub.hxx"
+#include "utils/Crc.hxx"
 
 namespace dcc
 {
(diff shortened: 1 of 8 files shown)


src/dcc/LogonFeedback.hxx
```diff
@@ -0,0 +1,272 @@
...
+                break;
+            case PacketType::LOGON_ENABLE:
+                any_error |=
+                    (((data >> 44) & 0xf) != RMOB_LOGON_ENABLE_FEEDBACK);
+                cb_->process_decoder_id(key, any_error, data & PAYLOAD_MASK);
+                break;
+            case PacketType::GET_DATA_START:
+            case PacketType::GET_DATA_CONT:
+            case PacketType::MISC_254:
+            case PacketType::UNKNOWN:
```
bakerstu, line 149: Do you want to add a default case here?

src/dcc/LogonFeedback.hxx
```diff
@@ -0,0 +1,272 @@
...
+        else
+        {
+            data |= ERROR_UNKNOWN;
+        }
+        shift -= 6;
+    }
+
+    /// Parses a concatenated ch1+ch2 response into a single uint64_t. It
+    /// contains error flags in the MSB, the number of valid 6-bit codepoints
+    /// in the second MSB , and 6 bytes data in the LSB, entered MSB-first.
```
bakerstu, line 218: MSB first means wire order?

src/utils/Crc.cxxtest
```diff
@@ -175,3 +175,25 @@ TEST(Crc8Test, Collision)
     LOG(INFO, "s1=%s s2=%s", string_to_hex(s1).c_str(),
         string_to_hex(s2).c_str());
 }
+
+TEST(Crc8Test, LogonExample)
```
bakerstu, line 179: I'm a bit confused. These tests, as well as some above, don't have any "EXPECTs". What is the acceptance criteria? Doesn't crash?

bakerstu (approved):

src/dcc/LogonFeedback.hxx
balazsracz: done

src/dcc/LogonFeedback.hxx
balazsracz: clarified:
```
    /// Parses a concatenated ch1+ch2 response into a single uint64_t. It
    /// contains error flags in the MSB, the number of valid 6-bit codepoints
    /// in the second MSB , and 6 bytes data in the LSB, entered with the first
    /// wire byte in the third MSB and last wire byte in the LSB.
    ///
    /// Any input byte that was missing or contained any other codepoint than a
    /// valid 6-bit data codepoint (including garbage or ACK) is translated
    /// into zero bits in the parse result.
    /// @param fb feedback to parse
    /// @return parse result.
```

src/utils/Crc.cxxtest
balazsracz: added comment:
```
// These examples contain specific messages from other unittests and print
// their CRC. This pattern can be used for developing other tests. There are no
// expectations in these tests.
```

PR #563, opened 2021-08-13
Adds helper functions for generating the logon packets

balazsracz:
Updates dcc::Packet to have helper functions for generating packets related to DCC logon:
- Logon Enable
- Select / get short info
- Logon Assign 

Moves the definitions of DCC packet related constants into a header so they are reusable.


src/utils/Crc.cxxtest
```diff
@@ -197,3 +197,27 @@ TEST(Crc8Test, LogonExample2)
     m.update256(0x22);
     LOG(INFO, "example: 0x%02x", m.get());
 }
+
+TEST(Crc8Test, LogonExample3)
```
bakerstu, line 201: There are no EXPECTs. What is the pass criteria?

bakerstu (approved):

src/utils/Crc.cxxtest
balazsracz: fixed above.

PR #568, opened 2021-08-19
Adds SPI driver for Tiva

balazsracz:


src/freertos_drivers/ti/TivaSPI.hxx
```diff
@@ -0,0 +1,320 @@
...
+
+        switch (bitsPerWord)
+        {
+            default:
+            case 8:
+                return transfer_polled<uint8_t>(msg);
+            case 16:
+                return transfer_polled<uint16_t>(msg);
+            case 32:
+                return transfer_polled<uint32_t>(msg);
```
bakerstu, line 239: 32-bit transfers not supported by hardware, should we have an assert here?

src/freertos_drivers/ti/TivaSPI.hxx
```diff
@@ -0,0 +1,320 @@
...
+    {
+        // Disable port first for reconfiguration.
+        HWREG(base_ + SSI_O_CR1) &= ~SSI_CR1_SSE;
+        
+        HWREG(base_ + SSI_O_CR0) = spiCr0_;
+        HWREG(base_ + SSI_O_CPSR) = spiPrescaler_;
+        HWREG(base_ + SSI_O_CR1) = spiCr1_;
+    }
+
+    OSSem *sem_; /**< reference to the semaphore belonging to this bus */
```
bakerstu, line 305: Does this get used anywhere? It may be dead code. Should probably check the CC3220 SPI driver too.

bakerstu (approved):

src/freertos_drivers/ti/TivaSPI.hxx
balazsracz: I believe it is used to signal the DMA complete interrupt.

balazsracz: Which explains why it seems unused, because DMA is still a TODO.

PR #571, opened 2021-09-11
Fixes a static initialization bug under linux.

balazsracz:
The osGlobalAtomic variable was accessed from the non-default threads while the
static initializers were running. This caused the mutex to be lost sometimes.
After this fix a static Atomic variable does not have a constructor call, but is
linker-initialized from the .data section.

This bug caused the hub application to get sometimes "livelocked" on linux, with the
packets printed to the screen but not being sent to any connected devices actually.


bakerstu (approved):
Looks good.

PR #575, opened 2021-09-14
Fix Minor Entry Model Bounded Bug

bakerstu:
This issue occurs when using EntryModel::putc_inc(). Once wrapping around pas the "max digits", the entry value should be blank. However, the clamping was not handling the wrap around case and setting the value to min or max rather than leaving it blank or "empty".

### DIFF
--- a/src/utils/EntryModel.cxxtest
+++ b/src/utils/EntryModel.cxxtest
@@ -1,0 +1,90 @@
+#include "utils/test_main.hxx"
+#include "utils/EntryModel.hxx"
+
+TEST(EntryModelTest, Create)
+{
+    EntryModel<int64_t, 11> em;
+
+    EXPECT_EQ(0U, em.digits());
+    EXPECT_EQ(0U, em.cursor_index());
+    EXPECT_FALSE(em.has_initial());
+    EXPECT_EQ("", em.parsed(true));
+}
+
+TEST(EntryModelTestBoundedTest, Create)
+{
+    EntryModelBounded<int64_t, 11> em;
+
+    EXPECT_EQ(0U, em.digits());
+    EXPECT_EQ(0U, em.cursor_index());
+    EXPECT_FALSE(em.has_initial());
+    EXPECT_EQ("", em.parsed(true));
+}
+
+TEST(EntryModelTestBoundedTest, InitValue)
+{
+    EntryModelBounded<int64_t, 11> em;
+
+    em.init(2, 10, 24, 1, 28, 0);
+
+    EXPECT_EQ(2U, em.digits());
+    EXPECT_EQ(0U, em.cursor_index());
+    EXPECT_TRUE(em.has_initial());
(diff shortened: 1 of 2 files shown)


balazsracz:
I think this now introduces a different bug around wraparound and clamping. Specifically, at the wraparound the payload is set to empty, but it is not guaranteed that 0 is within the bounds of [min, max]. Therefore it is possible to get into a state where the user accepts a value that is not within the allowed range.

Example:
- min=1, max=11, digits = 2
- initialize without an initial value
- at this point get_value() will return 0 which is not within the range
- press key '1', press key '1', press key '1'
- now again get_value() returns 0 which is not within the range

It seems to me that EntryModelBounded should also define a get_value() function which will have to do clamping.

I see another place where things can go wrong:
- if the max value has fewer digits than digits_ then it might be impossible to get to the clear state

bakerstu:
I know about the get_value returning 0, it is intentional. I would have to change the API to get the some kind of other error code. I however, have already changed, and tested on a branch, the code to call parsed(true) and check for an empty string before accepting the return value of get_value(). I can add this to procedure definition to the comments for get_value().

If you want a get_value() API that clamps, I suggest we modify the API to get_value(unsigned start_index = 0, bool force_clamp = false), and document how it works.

You are correct that if we have max value with digits < digits_, we will get into an undesired state. I suggest that we assert this as an incorrect parameter combination to the init() method. Can this be done with a static_assert so we don't have to incur a runtime check?

bakerstu:
I've made a bunch of changes. Please re-review.

balazsracz (commented):
This code is very confusing. It seems to me that a couple of variables and functions are incorrectly named and the documentation comments don't help clear up this confusion. After staring at this class for 2x1 hours I have very little confidence that I understand what it is trying to do and it is correctly doing what it is trying to do.

Of the naming, probably the most confusing of all is "hasInitial_". This should be named "isAtInitialValue_".
"index_" as a class variable name is meaningless. It should be "size_" or "length_".
"digits_" is also weird. It should be "maxSize_" or "maxLength_".
"putc_inc" is also weird. The same functionality is called "append" or "push_back" in the standard library.
"cursor_index()" seems to be really "size()".

It seems to me that the business logic that this class is doing is inconsistent, because it mixes controller code (like increment, append) and view code (like taking care of right justification or copying bytes to an output buffer).

I think a lot of this code could be simplified by dropping the character buffer and using the following class variables:
T value_;
unsigned numLeadingZeros_ : 5;
unsigned isAtInitialValue_ : 1; //< decides whether right justify is needed
unsigned maxSize_ : 5; //< expressed in digits.
unsigned size_ : 5; //< expressed in digits.

then doing the string render when the caller asks you to do so. There is no performance penalty from doing this considering that this class is actuated only at user interactions (meaning 1 per second or less). The clamping anyway does string parsing all the time which is a lot more expensive than rendering.

PR #576, opened 2021-09-20
WIP Refactor EntryModel

bakerstu:
- Moves to a model of the EntryModel metadata tracking by integer not C string
- Moves to an API model similar to other STL API

### DIFF
--- a/src/utils/EntryModel.hxx
+++ b/src/utils/EntryModel.hxx
@@ -230,7 +230,7 @@ public:
     }
 
     /// Test if cursor is visible.
-    /// @return true if cursor is visiable, else false
+    /// @return true if cursor is visible, else false
     bool cursor_visible()
     {
         return size_ < maxSize_;
@@ -357,7 +357,7 @@ public:
     {
         if (force || !empty_)
         {
-            // puposely do not reset the numLeadingZeros_
+            // purposely do not reset the numLeadingZeros_
             empty_ = false;
             if (value_ < valueMin_)
             {
@@ -401,7 +401,7 @@ public:
     }
 
 protected:
-    /// Set the min and max boundaries supported by the EntryModel parameters.
+    /// Set min and max boundaries supported based on maxSize_ (digit count).
     virtual void set_boundaries()
     {
         valueMax_ = 0;
@@ -433,9 +433,7 @@ protected:
     unsigned size_             : 5; ///< actual number of digits
     unsigned isAtInitialValue_ : 1; ///< true if still has the initial value
     unsigned empty_            : 1; ///< true if the value_ is "empty"
-    unsigned reserved_         : 15; ///< reserved bit space
-
-    int base_; ///< radix base
+    unsigned base_             : 6; ///< radix base
 
     DISALLOW_COPY_AND_ASSIGN(EntryModel);
 };
@@ -447,16 +445,15 @@ template <class T> class EntryModelBounded : public EntryModel<T>
 {
 public:
     /// Constructor.
-    /// @param upper force characters to be upper case
-    EntryModelBounded(bool upper = false)
-        : EntryModel<T>(upper)
+    EntryModelBounded()
+        : EntryModel<T>()
     {
     }
 
     /// Initialize with a value.
     /// @param max_size max number of digits in the base type
     /// @param base base type, 10 or 16
-    /// @param value unsigned value to initialize with
+    /// @param value value to initialize with
     /// @param min minumum value
     /// @param max maximum value
     /// @param default_val default value
@@ -480,7 +477,7 @@ private:
     /// Clamp the value at the min or max.
     /// @param force Normally, clamping doesn't occur if the entry is "empty".
     ///              However, if force is set to true, we will clamp anyways.
-    void clamp(bool force = false)
+    void clamp(bool force = false) override
     {
         if (force && EntryModel<T>::empty_)
         {
@@ -494,7 +491,7 @@ private:
 
     /// Override base class to do nothing. The boundaries come from
     /// EntryModelBounded::init().
-    void set_boundaries()
+    void set_boundaries() override
     {
     }
 
--- a/src/utils/EntryModel.hxx
+++ b/src/utils/EntryModel.hxx
@@ -189,14 +189,7 @@ public:
             isAtInitialValue_ = false;
             clamp();
             // need to compute the size now that the initial value is false
-            if (value_ <= 0)
-            {
-                size_ = 1;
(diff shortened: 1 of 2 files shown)


src/utils/EntryModel.hxx
```diff
@@ -41,389 +41,465 @@
...
+    /// @param c character to append, base 10: 0 - 9, base 16: 0 - F
+    void push_back_char(char c)
     {
-        memset(data_, ' ', digits_);
-        if (data)
+        switch (base_)
         {
-            memcpy(data_, data, strlen(data));
+            case 10:
+                HASSERT(c >= '0' && c <= '9');
```
balazsracz, line 145: it is a bit scary to crash like this, maybe we should just ignore invalid input (do nothing)

src/utils/EntryModel.hxx
```diff
@@ -41,389 +41,465 @@
...
+
+    /// Test if the entry is "empty". Having an initial value is not empty.
+    /// @return true if empty, else false
+    bool empty()
+    {
+        return (!isAtInitialValue_ && empty_);
+    }
+
+    /// Test if cursor is visible.
+    /// @return true if cursor is visiable, else false
```
balazsracz, line 243: typo

src/utils/EntryModel.hxx
```diff
@@ -41,389 +41,465 @@
...
+    /// Clamp the value at the min or max.
+    /// @param force Normally, clamping doesn't occur if the entry is "empty".
+    ///              However, if force is set to true, we will clamp anyways.
+    virtual void clamp(bool force = false)
+    {
+        if (force || !empty_)
         {
-            memmove(data_ + 1, data_, digits_ - 1);
-            data_[0] = '-';
+            // puposely do not reset the numLeadingZeros_
```
balazsracz, line 372: typo

src/utils/EntryModel.hxx
```diff
@@ -41,389 +41,465 @@
...
+            --value_;
+        }
+        clamp();
+        return value_;
     }
 
-    /// Clamp the value at the min or max.
-    void clamp()
+protected:
+    /// Set the min and max boundaries supported by the EntryModel parameters.
```
balazsracz, line 412: /// Set the min and max boundaries supported based on the maxSize_ (digit count).

src/utils/EntryModel.hxx
```diff
@@ -41,389 +41,465 @@
...
 
-    int base_; /**< radix base */
-    char data_[N + 1]; /**< data string */
+    unsigned numLeadingZeros_  : 5; ///< number of leading zeros
+    unsigned maxSize_          : 5; ///< maximum number of digits
+    unsigned size_             : 5; ///< actual number of digits
+    unsigned isAtInitialValue_ : 1; ///< true if still has the initial value
+    unsigned empty_            : 1; ///< true if the value_ is "empty"
+    unsigned upper_            : 1; ///< force characters to be upper case
+    unsigned reserved_         : 15; ///< reserved bit space
```
balazsracz, line 434: please remove reserved_; the compiler will automatically reserve the necessary number of bits.
(This explicit mention of reserved opens up a bug where you reserve too many bits and end up wasting an entire unsigned due to that.)

src/utils/EntryModel.hxx
```diff
@@ -41,389 +41,465 @@
...
+
+    int base_; ///< radix base
```
balazsracz, line 436: It seems like this could fit into the bit field? We only need to support 10 and 16 so 5 or 6 bits should be enough.

src/utils/EntryModel.hxx
```diff
@@ -41,389 +41,465 @@
...
-    {
-        EntryModel<T, N>::set_value(min_);
-    }
-
-    /// Set the value to the maximum.
-    void set_max()
+    /// Initialize with a value.
+    /// @param max_size max number of digits in the base type
+    /// @param base base type, 10 or 16
+    /// @param value unsigned value to initialize with
```
balazsracz, line 457: i think this does not need to be unsigned, it could be signed.

src/utils/EntryModel.hxx
```diff
@@ -41,389 +41,465 @@
...
+        EntryModel<T>::set_value(valueDefault_);
     }
 
-    /// Pre-increment value.
-    T operator ++()
+private:
+    /// Clamp the value at the min or max.
+    /// @param force Normally, clamping doesn't occur if the entry is "empty".
+    ///              However, if force is set to true, we will clamp anyways.
+    void clamp(bool force = false)
```
balazsracz, line 481: override

src/utils/EntryModel.hxx
```diff
@@ -41,389 +41,465 @@
...
         }
-        return value;
     }
 
-private:
-    /// Clamp the value at the min or max.
-    void clamp()
+    /// Override base class to do nothing. The boundaries come from
+    /// EntryModelBounded::init().
+    void set_boundaries()
```
balazsracz, line 495: override

balazsracz (approved):
This is very good, thank you! Now it is easy to see that every single function is correctly performing the logic as needed.

src/utils/EntryModel.hxx
bakerstu: Actually, this is what I prefer. We should get here based on incorrect runtime code in the same executable only.

src/utils/EntryModel.hxx
bakerstu: Fixed.

src/utils/EntryModel.hxx
bakerstu: Fixed.

src/utils/EntryModel.hxx
bakerstu: Fixed.

src/utils/EntryModel.hxx
bakerstu: Fixed.

src/utils/EntryModel.hxx
bakerstu: Fixed.

src/utils/EntryModel.hxx
bakerstu: Fixed.

src/utils/EntryModel.hxx
bakerstu: Fixed.

src/utils/EntryModel.hxx
bakerstu: Fixed.

PR #577, opened 2021-09-27
Adds platform-independent interface for sleeping.

balazsracz:
The intention is to replace all usage of deprecated POSIX function `usleep` with the function defined in this cross-platform header. Then different platforms can provide their implementation.

This will allow us to remove the feature macro hacks that we had to put in to make the code compile on recent releases of armgcc, such as `#define _BSD_SOURCE` or `#define _DEFAULT_SOURCE`. These are harmful, because they often conflict with each other and sometimes with expectations from middleware libraries (e.g. a macro declaring C99 source).

### DIFF
--- a/src/os/sleep.h
+++ b/src/os/sleep.h
@@ -44,7 +44,7 @@
 static void microsleep(uint32_t microseconds) __attribute__((weakref("usleep")));
 
 
-/// Executes a busy loop for a given amount of time. It is recommended ot use
+/// Executes a busy loop for a given amount of time. It is recommended to use
 /// this only for small number of microseconds (e.g. <100 usec).
 /// @param microseconds how long to delay.
 extern void microdelay(uint32_t microseconds);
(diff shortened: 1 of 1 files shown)


src/os/sleep.h
```diff
@@ -0,0 +1,52 @@
...
+#include <inttypes.h>
+
+/// Sleep a given number of microseconds. The granularity of sleep depends on
+/// the operating system, for FreeRTOS sleeping less than 1 msec is not
+/// possible with this function.
+/// @param microseconds how long to sleep.
+static void microsleep(uint32_t microseconds) __attribute__((weakref("usleep")));
+
+
+/// Executes a busy loop for a given amount of time. It is recommended ot use
```
atanisoft, line 47: Typo

balazsracz: Fixed

bakerstu (approved):

PR #585, opened 2021-12-05
Adds support for SPI connected flash chips.

balazsracz:
- Adds a helper class that implements the common operations.
- Adds an instantiation of the SPIFFS driver to use the external flash chip.

### DIFF
--- a/src/freertos_drivers/common/SPIFlash.hxx
+++ b/src/freertos_drivers/common/SPIFlash.hxx
@@ -1,0 +1,168 @@
+/** \copyright
+ * Copyright (c) 2021, Balazs Racz
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are  permitted provided that the following conditions are met:
+ *
+ *  - Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ *  - Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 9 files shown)


src/freertos_drivers/common/SPIFlash.hxx
```diff
@@ -0,0 +1,168 @@
...
+///    .spiMode_ = 3,
+///    .writeCommand_ = 0xff,
+///};
+struct SPIFlashConfig
+{
+    /// Use this frequency to talk to SPI.
+    uint32_t speedHz_ {1000000};
+
+    /// How many bytes is an erase sector.
+    uint32_t sectorSize_ {4 * 1024};
```
bakerstu, line 62: This value should come from a table based on reported ID. A todo and an issue filed for enhancement is good enough for now. Of course, we may need to get this value before instantiating the SPIFFS so that it gets the correct sector size too.

We may find ourselves down the road needing to have a second source for the SPI flash that uses different sector and page sizes.

src/freertos_drivers/common/SPIFlash.hxx
```diff
@@ -0,0 +1,168 @@
...
+
+    /// A page program operation might wrap around a page. This will cause
+    /// bytes to be written to the wrong place. There is a check that prevents
+    /// this.
+    ///
+    /// This variable is the mask on the address bits that define the
+    /// page. Each write operation must start and finish within the same
+    /// address & pageSizeMask_.
+    uint32_t pageSizeMask_ {~(256u - 1)};
```
bakerstu, line 71: same.

bakerstu (approved):

src/freertos_drivers/common/SPIFlash.hxx
balazsracz: These are just the default initialization values. The expected usage is that there is a struct in HwInit which can be filled up for the given hardware. We could possibly put this into RAM and change the values, or select from more than one such structure that are in flash.

bakerstu: Ultimately, this is discoverable at runtime by evaluating the SPI FLASH ID itself. That is how we should get the value.

balazsracz: filed #586

PR #597, opened 2021-12-28
Implements BUS_OFF handling in the TivaCan driver.

balazsracz:
- Adds can bus status ioctl.
- Flushes the TX queue when bus warning or bus error
  is observed by the controller.
- Ensures that all message transmissions continue flowing
  (and are discarded) while we're in an error condition
  like bus off.

### DIFF
--- a/src/freertos_drivers/ti/TivaCan.cxx
+++ b/src/freertos_drivers/ti/TivaCan.cxx
@@ -33,15 +33,16 @@
 
 #include <stdint.h>
 
-#include "inc/hw_types.h"
-#include "inc/hw_memmap.h"
-#include "inc/hw_ints.h"
-#include "inc/hw_can.h"
-#include "driverlib/rom.h"
-#include "driverlib/rom_map.h"
+#include "can_ioctl.h"
 #include "driverlib/can.h"
 #include "driverlib/interrupt.h"
+#include "driverlib/rom.h"
+#include "driverlib/rom_map.h"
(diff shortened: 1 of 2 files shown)


src/freertos_drivers/ti/TivaCan.cxx
```diff
@@ -155,13 +177,20 @@ void TivaCan::interrupt_handler()
...
+            txBuf->flush();
+            txPending = false;
+            txBuf->signal_condition_from_isr();
         }
         if (status & CAN_STATUS_EWARN)
         {
             /* One of the error counters has exceded a value of 96 */
             ++softErrorCount;
-            /* flush and data in the tx pipeline */
-            MAP_CANMessageClear(base, 2);
```
bakerstu, line 164: Are you sure you want to remove this line of code? You could end up with a stale message in the CAN object when the bus recovers.

bakerstu (approved):

src/freertos_drivers/ti/TivaCan.cxx
balazsracz: Yes, I want to keep a message in the transmit object, so that the CAN controller keeps attempting to transmit something periodically. Otherwise we might never realize the cable was plugged in because the bus is quiescent.
I made another change that causes the CAN frame be replaced with newly transmitted frames every time ::write() is called. This way the frame in the transmit object is not going to be super old, provided that there are regular transmissions happening.

PR #599, opened 2022-01-01
Adds binary release script.

balazsracz:
Adds script (in the makefile) to compile a binary release for OpenMRN.
The binary release contains the standalone applications that are useful
outside of OpenMRN. Currently this list is as follows:
- hub
- bootloader_client
- memconfig_utils
- send_datagram

The binary release script compiles these applications using the host g++ compiler, and packages it as a zip file.

Also added is a release-js script which compiles two applications using node.js package for win.exe and macos binaries.
These are:
- js_hub
- bootloader_client

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -34,6 +34,20 @@ The following is out of scope of this documentation:
   
   `sudo npm install -g pkg`
   
+## Release numbering
+
+Release numbers are marked as major.minor.patch, such as `v2.10.1`.
+
+- Major release numbers are incremented once per year (2020 is 0, 2021 is 1,
+  2022 is 2, ...).
+
+- Minor release numbers are 10, 20, 30, 40 for given quarters, then incremented
+  by one if there are multiple releases built within a quarter.
+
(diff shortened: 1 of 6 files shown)


bakerstu (commented):
Can you add a readme in mark down that describes how to run the release script?

RELEASE.md
```diff
@@ -0,0 +1,71 @@
...
+   
+5. On the linux.x86_64 host, build the javascript binaries. Run
+   
+   `make -j5 release-js`
+   
+6. Copy `openmrn/bin/release/applications.win.zip` and
+   `openmrn/bin/release/applications.macos.zip`.
+   
+7. On the OpenMRN GitHub project, select Releases, create a new release, select
+   create a new tag in the form of `release-v2.10.1`. Upload the release
```
bakerstu, line 68: We should describe the versioning pattern.

bakerstu (approved):

RELEASE.md
balazsracz: Done

PR #600, opened 2022-01-09
Minor fixes to SNIP client and SNIP handler

balazsracz:
- Adds proper shutdown to SNIP client. This is needed for unittesting purposes.
- Fixes crashing bug in SNIP client when returned payload is empty.
- Moves SNIP handler's static descriptor to a separate compilation unit to avoid
  it being unexpectedly linked into binaries. This is not only flash space wasted,
  but also presents an unexpected dependency on SNIP_DYNAMIC_FILENAME being available.

### DIFF
--- a/src/openlcb/SNIPClient.hxx
+++ b/src/openlcb/SNIPClient.hxx
@@ -39,6 +39,7 @@
 #include "executor/CallableFlow.hxx"
 #include "openlcb/Defs.hxx"
 #include "openlcb/If.hxx"
+#include "os/sleep.h"
 
 namespace openlcb
 {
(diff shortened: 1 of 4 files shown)


src/openlcb/SNIPClient.hxx
```diff
@@ -88,6 +89,17 @@ public:
     {
     }
 
+    /// Flushes the pending timed operations.
```
bakerstu, line 92: Is this shutdown method only for tests? Should it have an ifdef GTEST wrapper?

bakerstu (approved):

src/openlcb/SNIPClient.hxx
balazsracz: Unfortunately I can't make this test only.
I have another component which has ownership of a SNIPClient. That component must call this function in its destructor. The destructor cannot be made testonly, even if in practice it is only run in tests.

PR #606, opened 2022-01-31
Adds a script for computing optimal CAN-bus timing configurations.

balazsracz:


bakerstu (approved):
Would it be better to move this over to the openlcb/documents repo?

kiwi64ajs:
> Would it be better to move this over to the openlcb/documents repo?

Given it's going to form part of the documentation yes, probably.

Expecting this to end up in a document, it would be easier of the script produced its output in a tabular form with column headings etc - like a Spreadsheet to facilitate easy copy/paste into a table in an ODT file.

Are these numbers ever going to change or is this a one-off calculation, but we're providing the source of the calculations for reference?

balazsracz:
> > Would it be better to move this over to the openlcb/documents repo?
> 
> Given it's going to form part of the documentation yes, probably.
> 
> Expecting this to end up in a document, it would be easier of the script produced its output in a tabular form with column headings etc - like a Spreadsheet to facilitate easy copy/paste into a table in an ODT file.
> 
> Are these numbers ever going to change or is this a one-off calculation, but we're providing the source of the calculations for reference?

I am fine adding this to the documents repo as well.
The outputs don't change unless you change the formulas somehow. There are a couple of reasons why you might want to change the code:
- different CAN controllers have different restrictions on how the timing can be set. For example the best timing we found cannot be set on the MCP2515 because it has a tighter limit on acceptable SJW. Also the MCP2515 can only do up to 8 PROPSEG whereas Tivas can do up to 16 for the sum of PropSeg + PS1. So for different hardware different filters might be worth to implement as an add-on.
- I think there is one constant that can be changed, the TXRX propagation delay. This could be set differently for example for an non-isolated, a single-chip isolated, or a dual-chip isolated setup. But given that we don't have a standard for transceiver delay, maybe it does not make that much difference. I can see today's 297m long setting squeezing into the acceptable range if we look at the transceiver delay more accurately.

kiwi64ajs:
Ok, that’s helpful to have that confirmed as I kinda thought that but not close enough to the detail to know for sure.

What about changing the format of the output to better facilitate pasting the raw output it into a table in a LibreOffice doc without a bunch of manual reformatting?

PR #610, opened 2022-02-05
Fixes interpretation of O_CREAT in the CC32xxDeviceFile.

balazsracz:
When SL_FS_CREATE fails we will still attempt to do an FsOpen with SL_FS_WRITE instead of returning an error or crashing.

### DIFF
--- a/src/freertos_drivers/ti/CC32xxDeviceFile.cxx
+++ b/src/freertos_drivers/ti/CC32xxDeviceFile.cxx
@@ -92,13 +92,16 @@ int CC32xxDeviceFile::open(File* file, const char *path, int flags, int mode)
     else
     {
         /* file not open yet, open and intialize metadata */
-        int32_t result;
+        int32_t result = 0;
         if (flags & O_CREAT)
         {
             result = sl_FsOpen((const unsigned char *)path,
                                SL_FS_CREATE |
                                SL_FS_CREATE_MAX_SIZE(maxSizeOnCreate),
                                nullptr);
+        }
+        if (result > 0)
+        {
             writeEnable = true;
         }
         else if (flags & O_WRONLY)
@@ -163,6 +166,7 @@ void CC32xxDeviceFile::disable()
     {
         sl_FsClose(handle, nullptr, nullptr, 0);
         handle = -1;
+        writeEnable = false;
     }
 }
 


src/freertos_drivers/ti/CC32xxDeviceFile.cxx
```diff
@@ -92,13 +92,16 @@ int CC32xxDeviceFile::open(File* file, const char *path, int flags, int mode)
...
-        int32_t result;
+        int32_t result = 0;
         if (flags & O_CREAT)
         {
             result = sl_FsOpen((const unsigned char *)path,
                                SL_FS_CREATE |
                                SL_FS_CREATE_MAX_SIZE(maxSizeOnCreate),
                                nullptr);
+        }
+        if (result > 0)
```
atanisoft, line 103: this looks like it will break the original if/else if setup. Perhaps moving this block into the `O_CREAT` block? It also looks like `O_WRONLY` may also assume that write should be enabled.

bakerstu: Actually, I think this logic is correct.

bakerstu (approved):

src/freertos_drivers/ti/CC32xxDeviceFile.cxx
atanisoft: Technically the flow is correct but it changes the original behavior of the if/else if on read/write vs write only. Write only also has the same bug of not checking the result code form opening the file and will assume it can write even if the file is not opened. Likely the solution would be move the result check further down and only set write enabled if the file was successfully opened for write mode (read or not)

balazsracz:
At the top of ::open there are checks that exclude a variety of unsupported
combinations. This change here is only valid in the presence of those
checks.
In particular O_RDWR is entirely unsupported. So you have to open with
either O_RDONLY or with O_WRONLY.
O_CREAT | O_RDONLY makes no sense, therefore the only valid use-cases are:
- O_RDONLY
- O_WRONLY
- O_WRONLY | O_CREAT

The idea of this change is that if we are called with O_WRONLY|O_CREAT but
encounter an error during creation (an example such error would be
SL_ERROR_FS_FILE_ALREADY_EXISTS), we fall back to an open for write without
create. Currently we just crash.

PR #611, opened 2022-02-06
Fixes bug in FlashFile O_TRUNC handling.

balazsracz:
There was a bug in how we interpreted the O_TRUNC in FlashFile::open. We were looking for O_WRONLY in the mode bits but O_WRONLY is actually part of the flags bits.

### DIFF
--- a/src/freertos_drivers/common/FlashFile.hxx
+++ b/src/freertos_drivers/common/FlashFile.hxx
@@ -82,7 +82,7 @@ public:
     /// Overrides behavior of open for O_TRUNC.
     int open(File *file, const char *path, int flags, int mode) override
     {
-        if ((mode & O_WRONLY) && (flags & O_TRUNC))
+        if (flags & O_TRUNC)
         {
             // erase entire file.
             flash_->erase(flashStart_, size_);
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/common/FlashFile.hxx
```diff
@@ -82,7 +82,7 @@ public:
     /// Overrides behavior of open for O_TRUNC.
     int open(File *file, const char *path, int flags, int mode) override
     {
-        if ((mode & O_WRONLY) && (flags & O_TRUNC))
+        if (flags & O_TRUNC)
```
bakerstu, line 85: This is not correct. It should only be truncated if O_WRONLY or O_RDWR. If O_RDONLY, O_TRUNC should be ignored. The code should look like this:
`if ((flags & O_TRUNC) && (flags & O_ACCMODE) != O_RDONLY)`

From the standard "The result of using O_TRUNC without either O_RDWR or O_WRONLY is undefined." Though I do not think we should be truncating a file opened for read only in any case.

bakerstu (approved):

src/freertos_drivers/common/FlashFile.hxx
balazsracz: done

PR #614, opened 2022-03-06
Explicit handling of NOT_RESPONDING aliases.

balazsracz:
Ensures that we can enter more than one ID with NOT_RESPONDING alias into the alias cache.
This is critical so that the AddressedWriteFlow can correctly function in the presence
of node IDs that disappeared from the network (or never existed).

We already have a workaround where we add a node ID with the special alias NOT_RESPONDING,
but the AliasCache was not prepared for having more than one such entry, and thus it was
evicting these entries unexpectedly.

This PR makes the NOT_RESPONDING entries unique in the aliasMap by appending part of the
storage offset to the alias.

### DIFF
--- a/src/openlcb/AliasCache.cxx
+++ b/src/openlcb/AliasCache.cxx
@@ -47,6 +47,19 @@ namespace openlcb
 
 #define CONSTANT 0x1B0CA37ABA9 /**< constant for random number generation */
 
+/// This code removes the unique bits in the stored node alias in case this is
+/// a NOT_RESPONDING entry.
+/// @param stored alias in the metadata storage
+/// @return the alias if it's valid or NOT_RESPONDING ifthis is a sentinel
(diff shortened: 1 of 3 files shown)


src/openlcb/AliasCache.cxx
```diff
@@ -399,7 +454,8 @@ bool AliasCache::retrieve(unsigned entry, NodeID* node, NodeAlias* alias)
     Metadata* md = pool + entry;
     if (!md->alias_) return false;
     if (node) *node = md->get_node_id();
-    if (alias) *alias = md->alias_;
+    if (alias)
```
bakerstu, line 457: style, add braces.

bakerstu (approved):

src/openlcb/AliasCache.cxx
balazsracz: done

PR #615, opened 2022-03-19
Simplifies traction node OpenLCB addresses for DCC locomotives.

balazsracz:
Currently we have functions that use an encoding that is more complicated than necessary.
The encoding is not fixed in the standard in any way, so it's rather arbitrary what we put here.

After this PR we have the following encoding:
node_id = 0x060100000000 | address | (is_long ? 0xC000 : 0);

### DIFF
--- a/src/openlcb/TractionCvSpace.cxx
+++ b/src/openlcb/TractionCvSpace.cxx
@@ -282,7 +282,7 @@ StateFlowBase::Action TractionCvSpace::fill_read1_packet()
      * between long and short addresses */
     if (dccAddress_ >= 0x80)
     {
-        b->data()->add_dcc_address(dcc::DccLongAddress(dccAddress_));
+        b->data()->add_dcc_address(dcc::DccLongAddress(dccAddress_ & 0x3FFF));
     }
     else
     {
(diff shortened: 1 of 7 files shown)


src/openlcb/TractionDefs.hxx
```diff
@@ -216,14 +216,7 @@ struct TractionDefs {
...
             case dcc::TrainAddressType::DCC_LONG_ADDRESS:
-                if (addr < 128)
-                {
-                    return NODE_ID_DCC | 0xC000 | addr;
-                }
-                else
-                {
-                    return NODE_ID_DCC | addr;
-                }
+                return NODE_ID_DCC | 0xC000 | addr;
```
atanisoft, line 219: Can we move `0xC000` to a constant that can be used elsewhere?

src/openlcb/TractionCvSpace.cxx
```diff
@@ -282,7 +282,7 @@ StateFlowBase::Action TractionCvSpace::fill_read1_packet()
      * between long and short addresses */
     if (dccAddress_ >= 0x80)
```
atanisoft, line 283: Since we have a known bit indicating long/short address, can we move this check up to `TractionCvSpace::set_node(Node *node)` where we can test the node ID?

atanisoft (approved):
Overall looks good, added a couple comments.

src/openlcb/TractionDefs.hxx
balazsracz: done

src/openlcb/TractionCvSpace.cxx
balazsracz: Yes. I will do this in a separate PR though.

PR #621, opened 2022-04-17
Adds byte_stream.md, a design documentation for a proposed feature.

balazsracz:


doc/byte_stream.md
```diff
@@ -0,0 +1,137 @@
...
+## Requirements
+
+- Memory allocation
+
+  - Allocation has to avoid memory fragmentation. While the expectation is that
+    larger blocks of memory are allocated in one go, these blocks should be
+    reused between different users of Byte Streams and at different times,
+    instead of returning to malloc/free.
+  
+  - Block allocation size should be 1 kbyte. This is small enough that even
```
bakerstu, line 36: What about making the allocation size exactly 1 MTU, where MTU is configurable for the use case?

doc/byte_stream.md
```diff
@@ -0,0 +1,137 @@
...
+  This buffer holds the raw data. This buffer is reference counted, shared
+  between all sinks, and never entered into a `Q`. All sinks consider this
+  buffer as read-only.
+
+- `using ByteBuffer = Buffer<ByteChunk>;`
+
+  This buffer holds a reference to a RawBuffer, and start/end pointers within
+  that describe the exact range of bytes to use. This buffer is not shareable,
+  it can be entered into a queue, i.e., sent to a `StateFlowWithQueue` /
+  `FlowInterface<ByteBuffer>`.
```
bakerstu, line 80: Can a ByteBuffer hold a subset of another ByteBuffer, or only RawBuffer?

doc/byte_stream.md
balazsracz: Making the allocation size a compile time constant (constexpr or `#define`) seems plausible. Making it a run-time constant is a lot more difficult, and probably also a link-time constant is hard. There are places in openmrn that are doing `sizeof(Buffer<T>)` and those will fail when T's size is not a compile time constant.

doc/byte_stream.md
balazsracz: This is described in detail under Implementation > Zero-copy.
But specifically to your question, there is no reason for a second level indirection. When a ByteBuffer is cloned, it takes a new reference to the same RawBuffer (if there is a reference to begin with), and copies the void* and size values from the other ByteBuffer.

doc/byte_stream.md
bakerstu: I understand. A constexpr seems fine to me.

bakerstu (approved):

PR #622, opened 2022-05-02
Adds implementation of byte buffer.

balazsracz:


src/utils/Buffer.cxx
```diff
@@ -32,11 +32,17 @@
...
 
 #include "utils/Buffer.hxx"
+#include "utils/ByteBuffer.hxx"
 
 DynamicPool *mainBufferPool = nullptr;
+Pool *rawBufferPool = nullptr;
 
 Pool* init_main_buffer_pool()
 {
+    if (!rawBufferPool)
```
bakerstu, line 42: There is a potential race condition here between two threads calling this for the "first" time. I suspect in reality, this is initialized before multi-threading is started. It would be good to either make the check/allocate atomic or make a note on the expected startup behavior in the header file comment block.

You could alternatively use pthread_once.

bakerstu (approved):

src/utils/Buffer.cxx
balazsracz: This is true in abstract, but in the concrete case it is unnecessary. As you see from the surrounding code, the mainBufferPool is not protected by a pthread_once either. The reason this is not a problem is that this code runs at least once at static initialization time, and there are no threads at that point.

PR #623, opened 2022-05-02
Adds StreamSenderCan for the streaming protocol.

balazsracz:
Adds an interface state flow for sending a stream. This uses the new bytebuffer API for the data payloads, and handles the state machine and the CAN frame rendering.

This code is compatible with Bootloader.hxx and is essentially a refactoring of BootloaderClient.hxx.


src/openlcb/StreamSender.hxx
```diff
@@ -0,0 +1,555 @@
...
+        return errorCode_;
+    }
+    
+    /// Start of state machine, called when a buffer of data to send arrives
+    /// from the application layer.
+    Action entry()
+    {
+        if (requestInit_)
+        {
+            /// @todo get the destination address somehow.
```
bakerstu, line 191: Do we already have the destination from start_stream()?

bakerstu (approved):

src/openlcb/StreamSender.hxx
balazsracz: yes, this was an outdated todo. removed.

PR #629, opened 2022-06-18
TCAN4550 Instrumentation Addition

bakerstu:
Adds GPIO toggle for error passive and bus-off states.

### DIFF
--- a/src/freertos_drivers/common/TCAN4550Can.hxx
+++ b/src/freertos_drivers/common/TCAN4550Can.hxx
@@ -35,14 +35,16 @@
 #define _FREERTOS_DRIVERS_COMMON_TCAN4550CAN_HXX_
 
 #include "Can.hxx"
+#include "DummyGPIO.hxx"
 #include "SPI.hxx"
 
+#include "os/Gpio.hxx"
 #include "os/OS.hxx"
 #include "utils/Atomic.hxx"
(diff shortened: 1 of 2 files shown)


src/freertos_drivers/common/TCAN4550Can.hxx
```diff
@@ -55,8 +57,10 @@ public:
     /// @param name name of this device instance in the file system
     /// @param interrupt_enable callback to enable the interrupt
     /// @param interrupt_disable callback to disable the interrupt
+    /// @param test_pin test GPIO pin for instrumenting the code
     TCAN4550Can(const char *name,
-                void (*interrupt_enable)(), void (*interrupt_disable)())
+                void (*interrupt_enable)(), void (*interrupt_disable)(),
+                const Gpio *test_pin = DummyPinWithRead::instance())
```
balazsracz, line 63: do you want to ifdef this argument as well? (would avoid a reference to DummyPinWithRead::instance(), which is code in flash, vtable, etc.)

balazsracz (approved):

src/freertos_drivers/common/TCAN4550Can.hxx
bakerstu: Fixed.

PR #637, opened 2022-08-08
Add BroadcastTime Utilities

bakerstu:
- adds set_date_year_str to BroadcastTime (base class) to setthe date from an "MMM dd, yyyy" string format.
- adds query functions for year and month+day.
- adds an API call to distinguish generator and consumer.
- fixes bug where double trigger of a wakeup caused a crash
- adds utility function for parsing and rendering time and model clock rate

### DIFF
--- a/src/openlcb/BroadcastTimeAlarm.hxx
+++ b/src/openlcb/BroadcastTimeAlarm.hxx
@@ -165,7 +165,7 @@ protected:
 
 private:
     // Wakeup helper
-    class Wakeup : public Executable
+    class Wakeup : public Executable, protected Atomic
     {
     public:
         /// Constructor.
@@ -179,9 +179,17 @@ private:
         /// Trigger the wakeup to run.
         void trigger()
         {
-            if (!armed)
+            bool add = false;
             {
-                armed = true;
+                AtomicHolder h(this);
+                if (!armed)
+                {
+                    armed = true;
+                    add = true;
+                }
+            }
+            if (add)
+            {
                 alarm_->service()->executor()->add(this);
             }
         }
--- a/src/openlcb/BroadcastTimeAlarm.hxx
+++ b/src/openlcb/BroadcastTimeAlarm.hxx
@@ -172,13 +172,18 @@ private:
         /// @param alarm our parent alarm that we will awaken
         Wakeup(BroadcastTimeAlarm *alarm)
             : alarm_(alarm)
+            , armed(false)
         {
         }
 
         /// Trigger the wakeup to run.
         void trigger()
         {
-            alarm_->service()->executor()->add(this);
+            if (!armed)
+            {
+                armed = true;
+                alarm_->service()->executor()->add(this);
+            }
         }
 
     private:
@@ -186,10 +191,12 @@ private:
         /// on the CPU.
         void run() override
         {
+            armed = false;
             alarm_->wakeup();
         }
 
         BroadcastTimeAlarm *alarm_; ///< our parent alarm we will wakeup
+        bool armed;
     };
 
     /// Setup, or wait to setup alarm.
--- a/src/openlcb/BroadcastTimeDefs.cxx
+++ b/src/openlcb/BroadcastTimeDefs.cxx
@@ -37,6 +37,8 @@
 
 #include <string>
 
+#include "utils/format_utils.hxx"
+
 namespace openlcb
 {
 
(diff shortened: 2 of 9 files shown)


src/openlcb/BroadcastTimeDefs.hxx
```diff
@@ -341,6 +341,29 @@ struct BroadcastTimeDefs
...
+    /// @param rate rate in the form of rate quarters
+    /// @return rate represented in the form of a string (float)
+    static std::string rate_quarters_to_string(int16_t rate);
+
+    /// Convert a string (hh:mm) to hour and minute component integers.
+    /// @param stime time in the form of a string
+    /// @param hour resulting hour integer (0 to 23)
+    /// @param min resulting minute integer (0 to 59)
+    /// @return true on success, else false on fault
+    static bool string_to_time(std::string stime, int *hour, int *min);
```
balazsracz, line 361: this should be const string&

src/openlcb/BroadcastTimeDefs.cxx
```diff
@@ -0,0 +1,161 @@
...
+#include "BroadcastTimeDefs.hxx"
+
+#include <string>
+
+namespace openlcb
+{
+
+extern "C"
+{
+// normally requires _GNU_SOURCE
```
balazsracz, line 45: when this happens in a cxx file, you can add `#define _WHATEVER_SOURCE` to the top of that cxx file before you include any headers. Then it's guaranteed that you will get the respective declaration.
Sometimes this runs into problems with duplicate definitions of preprocessor symbols. That needs to be fixed in the other occurrence of the `#define`, which is invariably in a hxx file, by adding a wrapper of `#ifndef _WHATEVER_SOURCE  #define _WHATEVER_SOURCE #endif`

src/openlcb/BroadcastTimeDefs.cxx
```diff
@@ -0,0 +1,161 @@
...
+{
+    std::string result;
+    if (rate < 0)
+    {
+        result.push_back('-');
+        rate = -rate;
+    }
+    uint16_t whole = rate >> 2;
+    uint16_t frac = rate & 0x3;
+    if (whole >= 100)
```
balazsracz, line 91: instead of this, you should just use format_utils's integer_to_string

src/openlcb/BroadcastTimeDefs.cxx
```diff
@@ -0,0 +1,161 @@
...
+    if (whole >= 10)
+    {
+        result.push_back('0' + ((whole % 100) / 10));
+    }
+    result.push_back('0' + (whole % 10));
+    switch (frac)
+    {
+        default:
+        case 0:
+            result.append(".00");
```
balazsracz, line 104: should we skip this? If the rate is an integer, rendering it as an integer makes the most sense.

src/openlcb/BroadcastTimeDefs.cxx
```diff
@@ -0,0 +1,161 @@
...
+    return result;
+}
+
+//
+// BroadcastTimeDefs::string_to_time()
+//
+bool BroadcastTimeDefs::string_to_time(std::string stime, int *hour, int *min)
+{
+    struct tm tm;
+    if (strptime(stime.c_str(), "%R", &tm) != nullptr)
```
balazsracz, line 125: reverse this if:
```
if (!...)
{
  return false
}
use output
return true
```

src/openlcb/BroadcastTimeDefs.cxx
```diff
@@ -0,0 +1,161 @@
...
+        }
+        return true;
+    }
+    return false;
+}
+
+//
+// BroadcastTimeDefs::string_to_rate_quaraters()
+//
+int16_t BroadcastTimeDefs::string_to_rate_quarters(std::string srate)
```
balazsracz, line 143: consider adding a unit test that exercised these utility functions, including verifying that they behave as expected around the limit values and errors.

src/openlcb/BroadcastTimeAlarm.hxx
```diff
@@ -172,24 +172,31 @@ private:
...
             : alarm_(alarm)
+            , armed(false)
         {
         }
 
         /// Trigger the wakeup to run.
         void trigger()
         {
-            alarm_->service()->executor()->add(this);
+            if (!armed)
```
atanisoft, line 182: should `armed` be an atomic_bool or otherwise protected from concurrency?

src/openlcb/BroadcastTimeDefs.cxx
bakerstu: https://github.com/bakerstu/openmrn/issues/640

src/openlcb/BroadcastTimeAlarm.hxx
bakerstu: Fixed.

src/openlcb/BroadcastTimeDefs.cxx
bakerstu: Fixed.

PR #641, opened 2022-08-09
Updates the dev kit example to support up to 8 x MCP23017 io expanders for a total of 128 lines.

rilull:
This will wrap up the changes needed to support up to 8 MCP23017 IO extenders on the Nucleo DevKit.


applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/main.cxx
```diff
@@ -474,11 +475,469 @@ constexpr const Gpio *const kPortExt0[] = {
 openlcb::MultiConfiguredPC ext0_pcs(
     stack.node(), kPortExt0, ARRAYSIZE(kPortExt0), cfg.seg().ext0_pc());
 
-#endif // if num extboards > 0
+#endif // if num mcpios = 2
+
+#if NUM_MCPIOS == 4
+
+MCP23017 exp0(&io_executor, 0, 0, 0);
```
balazsracz, line 482: we don't want to repeat those lines here that were already present inside the previous ifdef.
Instead, we can use something like this
```
#if NUM_MCPIOS > 0
first two
#endif
#if NUM_MCPIOS > 2
second two
#endif
... > 4
etc.
```

boards/st-stm32f303re-nucleo-dev-board/HwInit.cxx
```diff
@@ -66,15 +66,16 @@ static Stm32Uart uart0("/dev/ser0", USART2, USART2_IRQn);
 static Stm32Can can0("/dev/can0");
 
 /** EEPROM emulation driver. The file size might be made bigger. */
-static Stm32EEPROMEmulation eeprom0("/dev/eeprom", 4000);
+static Stm32EEPROMEmulation eeprom0("/dev/eeprom", 8192);
```
balazsracz, line 69: let's take 10% off of this maximum. We have a problem when a file size hits the maximum and we will wear out the flash super fast, even though it will appear to work correctly.

src/openlcb/MultiConfiguredPC.hxx
```diff
@@ -69,7 +69,7 @@ CDI_GROUP_END();
 
 /// CDI Configuration for a @ref ConfiguredConsumer.
 CDI_GROUP(PCConfig);
-enum class ActionConfig : uint8_t
+enum class ActionConfig : uint16_t
```
balazsracz, line 72: I don't see a reason for this change. Could you explain why it is needed?
There are only two values.

applications/nucleo_io/targets/freertos.armv6m.st-stm32f091rc-nucleo-dev-board/main.cxx
```diff
@@ -474,11 +475,469 @@ constexpr const Gpio *const kPortExt0[] = {
...
+MCP23017 exp1(&io_executor, 0, 0, 1);
+MCP23017 exp10(&io_executor, 0, 1, 0);
+MCP23017 exp11(&io_executor, 0, 1, 1);
+MCP23017 exp20(&io_executor, 1, 0, 0);
+MCP23017 exp21(&io_executor, 1, 0, 1);
+MCP23017 exp30(&io_executor, 1, 1, 0);
+MCP23017 exp31(&io_executor, 1, 1, 1);
+
+constexpr const MCP23017Gpio IOEXT0_A0(&exp0, MCP23017::PORTA, 0);
```
balazsracz, line 750: this is a super long sequence, maybe we could create something simpler to put here. I'll give a shot.

boards/st-stm32f303re-nucleo-dev-board/HwInit.cxx
balazsracz: done in main.cxx

PR #642, opened 2022-08-14
Additional reasons for the DCC output signal being disabled

bakerstu:


src/dcc/DccOutput.hxx
```diff
@@ -74,6 +74,10 @@ public:
         /// This output should be off due to the conflict between program track
         /// and normal operation mode.
         PGM_TRACK_LOCKOUT = 32,
+        /// A local request for disabling the output
+        LOCAL_DISABLE = 64,
```
balazsracz, line 78: we're running low on bits here. I think a local disable should rather use the CONFIG_SETTING bit and not add a new one.

src/dcc/DccOutput.hxx
```diff
@@ -74,6 +74,10 @@ public:
...
+        /// invalid incoming DCC signal
+        INVALID_SIGNAL_INPUT = 128,
```
balazsracz, line 80: once removing the local_disable, move this up to 64.

balazsracz (approved):

src/dcc/DccOutput.hxx
bakerstu: I'm not so sure about this. Are you implying that these two disable sources would not overlap in the same device?

What do you think of adding:
LOCAL_DISABLE = CONFIG_SETTING,

...also, if we need more bits, we should consider using a uint16_t.

balazsracz (approved):

PR #645, opened 2022-08-20
OpenSSL Use legacy API Name

bakerstu:
The legacy name is EVP_CIPHER_CTX_key_length. The latest OpenSSL has the following define:

#define EVP_CIPHER_CTX_key_length EVP_CIPHER_CTX_get_key_length


balazsracz:
why do we want to switch to the legacy name?

bakerstu:
> why do we want to switch to the legacy name?

Because it works with both old and new versions of OpenSSL, whereas the new name only works with new versions. My distribution of Linux doesn't support the new version of OpenSSL for example.

balazsracz (approved):

PR #649, opened 2022-09-12
Add BitRangeEventP

bakerstu:


src/openlcb/EventHandlerTemplates.hxx
```diff
@@ -794,6 +794,43 @@ private:
     unsigned size_; //< number of bits stored.
 };
 
+/// Producer event handler for a sequence of bits represented by a
+/// dense block of consecutive event IDs.
+class BitRangeEventP : public BitRangeEventPC
```
balazsracz, line 799: This class will probably advertise itself as consumer as well. You might need to override the identify global to avoid doing so.

balazsracz (approved):

src/openlcb/EventHandlerTemplates.hxx
bakerstu: Yes, good point. Fixed.

PR #651, opened 2022-09-15
Allows skipping unneeded event IDs from the factory initialization.

balazsracz:
Implements a heuristic and a manual control to avoid putting unneeded event offsets to the factory initialize event ID list.
- when there is a segment that's not SPACE_CONFIG, all events in that segment will be skipped.
- adds a SkipInit(true) option that can be set to manually force an individual event ID to not be factory initialized.


bakerstu (approved):
Thank you for taking this on. I think this is an elegant solution. I was able to effectively root cause this issue, but I was getting very frustrated coming up with a solution.

PR #655, opened 2022-09-18
Makes "found short" be persistent in programming track backend operations.

balazsracz:
When we get a backend op, we check if the output is already disabled due to short, and if yes, we report that to the caller.

### DIFF
--- a/src/dcc/ProgrammingTrackBackend.hxx
+++ b/src/dcc/ProgrammingTrackBackend.hxx
@@ -191,6 +191,10 @@ public:
     Action entry() override
     {
         request()->resultCode = OPERATION_PENDING;
+        auto* pgm = get_dcc_output(DccOutput::PGM);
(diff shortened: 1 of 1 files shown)


src/dcc/ProgrammingTrackBackend.hxx
```diff
@@ -200,9 +204,11 @@ public:
                 return call_immediately(STATE(exit_service_mode));
 
             case ProgrammingTrackRequest::Type::SEND_RESET:
+                if (has_short) request()->hasShortCircuit_ = 1;
```
bakerstu, line 207: Can we add braces in stead of having single line if statements? Same in the next case statement.

bakerstu (approved):

src/dcc/ProgrammingTrackBackend.hxx
balazsracz: done

PR #658, opened 2022-09-23
Simplify the sending of asynchronous events with new OpenLCB API option

bakerstu:


balazsracz (approved):
nice simplification of code!

PR #666, opened 2022-10-09
Adds a couple of more functions to cortex-M0 atomic operations.

balazsracz:
The atomic operations cannot be compiled by GCC because cortex-m0 does not have the instructions. So GCC calls these library functions, and we have implementation for some of them. This expands the list to a few more.


bakerstu (approved):
Can you add a function header block to these methods that describe the operation?

balazsracz:
> 

done

PR #669, opened 2022-10-16
Remove consumer from PolledProducer.

### LINKED ISSUE #663: openlcb::PolledProducer Incorrectly Generates Consumer Identified Messages
openlcb::PolledProducer is based off of BitEventPC instead of BitEventProducer. This results in it incorrectly generating consumer identified messages.

bakerstu:
Fixes #663

### DIFF
--- a/src/openlcb/PolledProducer.cxxtest
+++ b/src/openlcb/PolledProducer.cxxtest
@@ -61,7 +61,7 @@ protected:
         return hwState_;
     }
 
-    bool hwState_{false};
+    bool hwState_;
     PolledProducer<QuiesceDebouncer, FakeBit> p_;
     RefreshLoop l_;
 };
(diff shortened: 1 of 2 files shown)


src/openlcb/PolledProducer.cxxtest
```diff
@@ -61,7 +61,7 @@ protected:
         return hwState_;
     }
 
-    bool hwState_;
+    bool hwState_{false};
```
balazsracz, line 64: please remove theother initialization on line 45

balazsracz (approved):

src/openlcb/PolledProducer.cxxtest
bakerstu: Ah, I did not see that there. I'll roll back my change instead.

PR #679, opened 2022-11-28
SimpleStack Delay Node Initialization

bakerstu:
This fixes an issue whereby delaying the startup of a SimpleStack based node caused all of the initialization messages to be absent from the physical CAN bus if the CAN interface had not been bound to the node prior to the SimpleStack construction.

### DIFF
--- a/src/openlcb/DefaultNode.cxx
+++ b/src/openlcb/DefaultNode.cxx
@@ -41,15 +41,26 @@ namespace openlcb
 
 extern void StartInitializationFlow(Node* node);
 
-DefaultNode::DefaultNode(If* iface, NodeID node_id)
+DefaultNode::DefaultNode(If* iface, NodeID node_id, bool init)
     : nodeId_(node_id), isInitialized_(0), iface_(iface)
 {
     iface_->add_local_node(this);
-    StartInitializationFlow(this);
+    if (init)
+    {
+        StartInitializationFlow(this);
+    }
 }
 
 DefaultNode::~DefaultNode()
 {
+}
+
+void DefaultNode::initialize()
+{
+    // Ensures the caller's logic is statefully correct.
+    HASSERT(!isInitialized_);
+
+    StartInitializationFlow(this);
 }
 
 } // namespace openlcb
--- a/src/openlcb/DefaultNode.cxx
+++ b/src/openlcb/DefaultNode.cxx
@@ -39,28 +39,18 @@
 namespace openlcb
 {
 
-extern void StartInitializationFlow(Node* node);
-
 DefaultNode::DefaultNode(If* iface, NodeID node_id, bool init)
     : nodeId_(node_id), isInitialized_(0), iface_(iface)
 {
     iface_->add_local_node(this);
     if (init)
     {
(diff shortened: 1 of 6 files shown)


src/openlcb/SimpleStack.cxx
```diff
@@ -116,6 +116,7 @@ void SimpleStackBase::start_stack(bool delay_start)
     if (!delay_start)
     {
         start_iface(false);
+        node()->initialize();
```
balazsracz, line 119: i would expect to iterate the iface()->local_nodes list and call initialize on all of them. This would ensure that all virtual nodes get initialized as well.

src/openlcb/SimpleStack.cxx
```diff
@@ -193,6 +194,7 @@ void SimpleTrainCanStack::start_node()
 void SimpleStackBase::start_after_delay()
 {
     start_iface(false);
+    node()->initialize();
```
balazsracz, line 197: i would expect to iterate the iface()->first_local_node() / next_local_node() list and call initialize on all of them. This would ensure that all virtual nodes get correctly initialized as well.

src/openlcb/DefaultNode.cxx
```diff
@@ -41,15 +41,26 @@ namespace openlcb
...
+    {
+        StartInitializationFlow(this);
+    }
 }
 
 DefaultNode::~DefaultNode()
 {
 }
 
+void DefaultNode::initialize()
```
balazsracz, line 58: please add this function into DefaultTrainNode as well (in TractionTrain.hxx).
DefaultTrainNode does not inherit from defaultnode (on purpose). It does inherit from Node, so an alternative option would be to move this logic into Node.

balazsracz (approved):

src/openlcb/SimpleStack.cxx
bakerstu: I have made this update as suggested.

src/openlcb/Node.cxx
```diff
@@ -0,0 +1,50 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * @file Node.cxx
+ *
+ * Node definition for asynchronous NMRAnet nodes.
```
balazsracz, line 29: OpenLCB

balazsracz (approved):

src/openlcb/Node.cxx
bakerstu: Fixed.

PR #680, opened 2022-11-28
Add an EntryModel option to disable automatic clamping

bakerstu:


src/utils/EntryModel.cxxtest
```diff
@@ -630,6 +630,64 @@ TEST(EntryModelBoundedTest, SetMinMax)
     EXPECT_FALSE(em.empty());
 }
 
+TEST(EntryModelBoundedTest, SetMinMaxUnclamped)
```
balazsracz, line 633: should we have a test where min..max does not include 0? I think that was the specific bug we needed to address.

balazsracz (approved):

src/utils/EntryModel.cxxtest
bakerstu: Good idea, I will add this.

bakerstu: Added.

PR #682, opened 2022-12-04
Abstract the "generic" portions of DccAccConsumer into a "base" class

bakerstu:
Passes all test cases.

### DIFF
--- a/src/openlcb/DccAccyConsumer.hxx
+++ b/src/openlcb/DccAccyConsumer.hxx
@@ -43,17 +43,15 @@
 namespace openlcb
 {
 
-class DccAccyConsumer : public SimpleEventHandler
+/// Base (generic protocol) implementation of a DCC accessory consumer.
+class DccAccyConsumerBase : public SimpleEventHandler
 {
-public:
+protected:
     /// Constructs a listener for DCC accessory control.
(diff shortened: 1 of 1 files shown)


src/openlcb/DccAccyConsumer.hxx
```diff
@@ -67,7 +65,8 @@ public:
         memset(isStateKnown_, 0, sizeof(isStateKnown_));
     }
 
-    ~DccAccyConsumer()
+    /// Destructor.
+    ~DccAccyConsumerBase()
     {
         EventRegistry::instance()->unregister_handler(this);
     }
```
balazsracz, line 72: we should probably add a public: after this so that the virtual public interafce methods of the simpleeventhandler do not have their visibility changed. Then another protected: before the class variables.

balazsracz (approved):

src/openlcb/DccAccyConsumer.hxx
bakerstu: I don't agree that this is required. As long as the virtual methods are declared public in the SimpleEventHandler class, they can always be called on that object type. We don't want anyone calling these methods on the DccAccyConsumerBase object type, so they should not be public.

PR #684, opened 2022-12-09
Adds EEPROMFile driver for TI Tiva microcontrollers.

balazsracz:
The EEPROMFile allows using the built-in eeprom peripheral of the Tiva. There are some limitations, for example only 4-byte aligned reads and writes are supported.

### DIFF
--- a/src/freertos_drivers/ti/TivaEEPROMFile.hxx
+++ b/src/freertos_drivers/ti/TivaEEPROMFile.hxx
@@ -52,7 +52,8 @@ public:
         , byteSize_(byte_size)
     {
         MAP_SysCtlPeripheralEnable(SYSCTL_PERIPH_EEPROM0);
-        MAP_EEPROMInit();
+        // Due to Erratum MEM#11 we must never use ROM_EEPROMInit.
+        EEPROMInit();
         // This is for Tiva 129. For 123 it would be okay to have 2*64.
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/ti/TivaEEPROMFile.hxx
```diff
@@ -0,0 +1,103 @@
...
+class TivaEEPROMFile : public DeviceFile
+{
+public:
+    TivaEEPROMFile(const char *name, unsigned byte_offset, unsigned byte_size)
+        : DeviceFile(name)
+        , byteOffset_(byte_offset)
+        , byteSize_(byte_size)
+    {
+        MAP_SysCtlPeripheralEnable(SYSCTL_PERIPH_EEPROM0);
+        MAP_EEPROMInit();
```
bakerstu, line 55: There is an errata on ROM_EEPROMInit(). Use EEPROMInit() without the MAP_ prefix to ensure that the ROM version never gets linked in. When making this change, please add an inline comment referencing the Errata.

bakerstu (approved):

src/freertos_drivers/ti/TivaEEPROMFile.hxx
balazsracz: done

PR #687, opened 2022-12-25
Adds stream read support to the memory config handler

balazsracz:
The new stream-related functionality is factored out into separate classes such that it is possible to link against OpenMRN without including the code size impact of streams.

- Refactors the MemoryConfigHandler into a base class with a bunch of utilities and the actual command handlers.
- Adds an alternate child class with the command handler for stream reads.
- Registers the alternate implementation with the base handler and forwards the stream read/write commands to it.
- Creates a new class StreamTransport which will be available as a pointer from the If* object. This is optional to instantiate, and keeps the stream-related objects of the interface.
- Adds an async allocator to StreamTransport to acquire stream sender objects. This is similar to how DatagramService deals with datagram send clients.
- Makes stream sender and receiver not be tied to a Node* object, but only to an If* object. The node_ is then taken only when the transmit/receive is invoked. This makes it possible to operate with a single stream sender/receiver across multiple virtual nodes.
- Adds APIs for specifying and querying the stream dst IDs on/from the StreamSender.
- Adds ability to send a CAN frame to the local interface (as if it came from the network).
- Updates Stream Sender to be able to send streams to localhost (either same or different virtual node). This uses the frame loopback functionality of CanIf.
- Ignores loopback stream data packets when checking for alias conflicts.
- fixed bug in alias conflict handler which misinterpreted a stream frame as a CID frame.
- Expands byte buffer API for direct (zero-copy) appends.
- Updates stream initiate response error flags to match the error codes of other standards. This needs to be brought into the draft text as well.
- creates async_stream_test_helper.hxx with refactoring some of the common classes from the stream receive test.

### DIFF
--- a/src/openlcb/StreamDefs.hxx
+++ b/src/openlcb/StreamDefs.hxx
@@ -54,7 +54,10 @@ struct StreamDefs
     {
         FLAG_CARRIES_ID = 0x01,
         FLAG_REJECT_OUT_OF_ORDER = 0x02,
-        FLAG_PERMANENT_ERROR = 0x40,
+        /// @todo synchronize the draft text in the standards repository with
+        /// these definitions. These definitions match other OpenLCB protocols.
+        FLAG_PERMANENT_ERROR = 0x10,
+        FLAG_TEMPORARY_ERROR = 0x20,
         FLAG_ACCEPT = 0x80,
     };
 
--- a/src/openlcb/StreamSender.hxx
+++ b/src/openlcb/StreamSender.hxx
@@ -81,10 +81,9 @@ public:
 class StreamSenderCan : public StreamSender
 {
 public:
-    StreamSenderCan(Service *service, IfCan *iface, Node *node)
+    StreamSenderCan(Service *service, IfCan *iface)
         : StreamSender(service)
         , ifCan_(iface)
-        , node_(node)
         , sleeping_(false)
         , requestClose_(false)
         , requestInit_(false)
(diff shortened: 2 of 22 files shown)


src/openlcb/StreamDefs.hxx
```diff
@@ -54,7 +54,10 @@ struct StreamDefs
     {
         FLAG_CARRIES_ID = 0x01,
         FLAG_REJECT_OUT_OF_ORDER = 0x02,
-        FLAG_PERMANENT_ERROR = 0x40,
+        /// @todo synchronize the draft text in the standards repository with
```
bakerstu, line 57: Did you file an issue in the OpenLCB documents repo?

bakerstu (approved):

src/openlcb/StreamSender.hxx
```diff
@@ -94,20 +93,26 @@ public:
     /// Initiates using the stream sender. May be called only on idle stream
     /// senders.
     ///
+    /// @param src Source virtual node on the local interface.
     /// @param dst Destination node ID to send the stream to.
     /// @param source_stream_id 8-bit stream ID to use on the this (the source)
     /// side.
+    /// @param dst_stream_id 8-bit stream ID to use on the this (the source)
```
atanisoft, line 100: is this description correct? Shouldn't it to be the remote side?

src/openlcb/StreamDefs.hxx
balazsracz: https://github.com/openlcb/documents/issues/56

src/openlcb/StreamSender.hxx
balazsracz: fixed

PR #692, opened 2023-01-29
Refactors fast time state into a separate base class from BroadcastTime.

balazsracz:


bakerstu (commented):
I'm struggling to understand what purpose this serves. The whole point of the BroadcastTime object was to have a server/client neutral location for this logic. Do you have a use case in mind that requires this refactor?

balazsracz:
> I'm struggling to understand what purpose this serves. The whole point of the BroadcastTime object was to have a server/client neutral location for this logic. Do you have a use case in mind that requires this refactor?

I had two use-cases in mind.
- Writing a DCC time client. The input would be DCC packets coming on the track with the fast clock packets. However, I would like the API to the software components to stay the same.
- I had a need for remembering the state of a clock to detect time jumps. In the end I solved this need in a different way though, but have not tested the new way yet.

bakerstu (approved):

bakerstu:
> * Writing a DCC time client. The input would be DCC packets coming on the track with the fast clock packets. However, I would like the API to the software components to stay the same.

This makes sense. I wondered if this is something you had in mind.

atanisoft:
> * The input would be DCC packets coming on the track with the fast clock packets.

Do you have an example of these sort of packets? I'm curious on the formatting and usages.

balazsracz:
They are in the newest DCC standard S9.2.1 (
https://www.nmra.org/sites/default/files/standards/sandrp/pdf/s-9.2.1_dcc_extended_packet_formats.pdf
)
Section 2.3.6.2

atanisoft:
Very cool, I haven't looked at this newest revision on the spec. I can certainly think of a few interesting use cases for stationary decoders, but they could also be implemented as LCC nodes as well.

PR #695, opened 2023-02-05
Fix build of Arduino STM32 CAN driver.

balazsracz:
It was broken by #644 which added a new feature for which there is no API on the Arduino.

### DIFF
--- a/src/freertos_drivers/st/Stm32Can.cxx
+++ b/src/freertos_drivers/st/Stm32Can.cxx
@@ -145,7 +145,7 @@ int Stm32Can::ioctl(File *file, unsigned long int key, unsigned long data)
     }
     return -EINVAL;
 }
-#endif
+#endif // !ARDUINO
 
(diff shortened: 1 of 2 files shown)


atanisoft (dismissed):
Do we need / want to add a comment on the `#endif` line for the define being checked for?

src/freertos_drivers/st/Stm32Can.cxx
```diff
@@ -132,6 +132,7 @@ Stm32Can::Stm32Can(const char *name)
 #endif
 }
 
+#ifndef ARDUINO
```
atanisoft, line 135: please add a similar block for `can_ioctl.h` include at top of tile

atanisoft (changes requested):

balazsracz:
> Do we need / want to add a comment on the `#endif` line for the define being checked for?

good point, done

atanisoft (approved):

PR #702, opened 2023-03-12
Adds STM32 RailcomDriver

balazsracz:
This driver is based on a fork of the TivaRailcomDriver; the hardware interaction is rewritten to use the STM32 HAL libraries and in part direct access to the USART peripheral.

A major difference is that unlike the Tiva, the STM32 devices do not have a FIFO on the UART hardware. This makes it impossible to only perform reads from the UART data registers when the cutout is done.
To work around this issue, the STM32 Railcom Driver uses DMA to copy the data from the UART hardware register into the buffer in the driver. This requires a free DMA channel for each railcom channel that is in use.

Misc changes:
- Implements toggle() in STM32 GPIO driver.
- Adds a limit to how many buffers the RailcomToOpenlcbDebugProxy is willing to allocate. When we run out of buffers, we drop the packet to the floor.

### DIFF
--- a/src/dcc/RailcomBroadcastDecoder.cxx
+++ b/src/dcc/RailcomBroadcastDecoder.cxx
@@ -48,15 +48,18 @@ bool RailcomBroadcastDecoder::process_packet(const dcc::Feedback &packet)
 {
     if (packet.ch1Size)
     {
-        return process_data(packet.ch1Data, packet.ch1Size);
-    }
-    else if (packet.ch2Size)
-    {
-        return process_data(packet.ch2Data, packet.ch2Size);
+        return process_data(packet.ch1Data, packet.ch1Size) &&
+            (packet.ch2Size == 0);
     }
     else
     {
-        return true; // empty packet.
+        // No channel1 data.
+        notify_empty();
+        if (!packet.ch2Size)
+        {
+            return true; // empty packet.
+        }
+        return false;
     }
 }
 
@@ -82,7 +85,7 @@ bool RailcomBroadcastDecoder::process_data(const uint8_t *data, unsigned size)
         {
             case dcc::RMOB_ADRLOW:
                 if (currentL_ == payload) {
-                    if (countL_ < REPEAT_COUNT) ++countL_;
+                    if (countL_ < MAX_REPEAT_COUNT) countL_ += 2;
                 } else {
                     currentL_ = payload;
                     countL_ = 0;
@@ -90,7 +93,7 @@ bool RailcomBroadcastDecoder::process_data(const uint8_t *data, unsigned size)
                 break;
             case dcc::RMOB_ADRHIGH:
                 if (currentH_ == payload) {
-                    if (countH_ < REPEAT_COUNT) ++countH_;
+                    if (countH_ < MAX_REPEAT_COUNT) countH_ += 2;
                 } else {
                     currentH_ = payload;
                     countH_ = 0;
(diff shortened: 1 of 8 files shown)


src/dcc/RailcomBroadcastDecoder.cxx
```diff
@@ -82,15 +85,15 @@ bool RailcomBroadcastDecoder::process_data(const uint8_t *data, unsigned size)
         {
             case dcc::RMOB_ADRLOW:
                 if (currentL_ == payload) {
-                    if (countL_ < REPEAT_COUNT) ++countL_;
```
bakerstu, line 85: The brace style in this section (and below) is incorrect per the style guidelines.

src/dcc/RailcomBroadcastDecoder.cxx
```diff
@@ -82,15 +85,15 @@ bool RailcomBroadcastDecoder::process_data(const uint8_t *data, unsigned size)
...
+                    if (countL_ < MAX_REPEAT_COUNT) countL_ += 2;
```
bakerstu, line 88: Please place if statement body in braces on its own line.

src/dcc/RailcomBroadcastDecoder.hxx
```diff
@@ -86,9 +86,15 @@ private:
...
+    /// Notifies the state machine that there is no occupancy detected.
+    void notify_empty();
+    
     /// How many times we shall get the same data out of railcom before we
     /// believe it and report to the bus.
-    static const uint8_t REPEAT_COUNT = 3;
+    static const uint8_t MIN_REPEAT_COUNT = 6;
+    /// Maximum number of repeats we remember. This is how many empty packets
+    /// we need to forget the current address when we're getting empty packets.
+    static const uint8_t MAX_REPEAT_COUNT = 8;
```
bakerstu, line 97: Can you enhance this comment somehow to indicate the intent to count up at 2x the rate that we count down?

src/freertos_drivers/st/Stm32Railcom.hxx
```diff
@@ -24,57 +24,85 @@
...
- * Then open /dev/railcom and read dcc::Feedback structures from it. Each read
- * much be exactly sizeof(dcc::Feedback) length. If there are no more packets to
- * read, you'll get return=0, errno==EAGAIN. Add an ioctl CAN_READ_ACTIVE to
- * get a notification when there is something to read.
+ * Device driver for STM32 chips to read one or more UART inputs for railcom
+ * data.
  *
  * @author Balazs Racz
- * @date 6 Jan 2015
+ * @date 6 Mar 2023
```
bakerstu, line 33: Update the copyright year on line 2 above.

bakerstu (approved):

src/freertos_drivers/st/Stm32Railcom.hxx
balazsracz: done

src/dcc/RailcomBroadcastDecoder.cxx
balazsracz: fixed

src/dcc/RailcomBroadcastDecoder.cxx
balazsracz: fixed

src/dcc/RailcomBroadcastDecoder.hxx
balazsracz: i changed the names and meaning instead to be more intuitive and not have double counts.

PR #703, opened 2023-03-28
bit bang i2c master

bakerstu:


src/freertos_drivers/common/BitBangI2C.cxx
```diff
@@ -0,0 +1,353 @@
...
+        }
+        case StateTx::DATA_7_SCL_CLR:
+        case StateTx::DATA_6_SCL_CLR:
+        case StateTx::DATA_5_SCL_CLR:
+        case StateTx::DATA_4_SCL_CLR:
+        case StateTx::DATA_3_SCL_CLR:
+        case StateTx::DATA_2_SCL_CLR:
+        case StateTx::DATA_1_SCL_CLR:
+        case StateTx::DATA_0_SCL_CLR:
+            gpio_clr(scl_);
```
balazsracz, line 241:
```
// I2C is specified such that the data on the SDA line must be stable during the high period of the clock, 
// and the data line can only change when SCL is Low.
//
// This means that the sequence always has to be 
// 1a. SCL := low
// 1b. set/clear SDA
// 2. wait a cycle for lines to settle
// 3a. SCL := high
// 3b. this edge is when the receiver samples
// 4. wait a cycle, lines are stable
 ```

This also means that we need to start the state machine by clearing the clock.

Comment (balazsracz) [reply src/freertos_drivers/common/BitBangI2C.cxx:241]: image

File: src/freertos_drivers/common/BitBangI2C.cxx [hunk 2]
```diff
@@ -0,0 +1,353 @@
...
+        case StateRx::DATA_0_SCL_CLR:
+            *data <<= 1;
+            if (sda_->read() == Gpio::Value::SET)
+            {
+                *data |= 0x01;
+            }
+            gpio_clr(scl_);
+            ++stateRx_;
+            break;
+        case StateRx::ACK_SDA_SCL_SET:
```
balazsracz, line 306: image

src/freertos_drivers/common/BitBangI2C.cxx
```diff
@@ -0,0 +1,353 @@
...
+        case StateTx::DATA_5_SCL_CLR:
+        case StateTx::DATA_4_SCL_CLR:
+        case StateTx::DATA_3_SCL_CLR:
+        case StateTx::DATA_2_SCL_CLR:
+        case StateTx::DATA_1_SCL_CLR:
+        case StateTx::DATA_0_SCL_CLR:
+            gpio_clr(scl_);
+            ++stateTx_;
+            break;
+        case StateTx::ACK_SDA_SCL_SET:
```
balazsracz, line 244: we need to start by clear scl set sda.
then wait a cycle.
then next cycle we set scl and sample sda for the ACK bit.
then wait a cycle

then we do the TX/RX state machine which will start by clearing SCL and (on tx) setting SDA.
the RX state machine should sample before the next high to low transition, where it should validate that the clock is actually high.

bakerstu:
Note: At this point, I will resolve the review comment to use a "HW" struct. This will result in the BitBangI2C.cxx file no longer existing since it will become a template class. That will cause the previous review history to get a bit wonky.

bakerstu:
Here are read and write screenshots from a logic analyzer (including analog plots) for the resulting output:

image
image

balazsracz (approved):
thank you for the extensive refactoring. i reviewed the state machines and i think we're good now.

src/freertos_drivers/common/BitBangI2C.hxx
```diff
@@ -171,26 +140,121 @@ private:
...
+/// BitBangI2C<I2CHwDefs> bitBangI2C("/dev/i2c0", disable_tick, enable_tick);
+/// @endcode
+///
+/// @tparam HW hardware interface to the access the SCL and SDA I/O lines
+template <class HW> class BitBangI2C : protected BitBangI2CStates, public I2C
+{
+public:
+    /// Constructor.
+    /// @param name name of this device instance in the file system
+    /// @param enable_tick callback to enable ticks
```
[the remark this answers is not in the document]
bakerstu: Fixed.

src/freertos_drivers/common/BitBangI2C.hxx
```diff
@@ -231,57 +276,462 @@ private:
...
+                }
+            }
+            break;
+        case State::DATA_RX:
+            if (state_rx(msg_->buf + count_, (count_ + 1 >= msg_->len)))
+            {
+                if (count_ < 0)
+                {
+                    // Some error occured, likely an unexpected NACK
+                    exit = true;
```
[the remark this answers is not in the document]
bakerstu: Fixed.

PR #706, opened 2023-04-15
Fix bug that convoluted the file mode for the UART hardware mode

bakerstu:


balazsracz (approved):

balazsracz (changes requested):
same bug seems to exist in TivaUart.cxx as well.

balazsracz:
> same bug seems to exist in TivaUart.cxx as well.

or maybe not. but it would be good to keep the two implementations somewhat consistent, so renaming the mode variable there to uartMode_ would be meaningful.

bakerstu:
@balazsracz I have updated the Tiva UART driver as instructed. I have ensured that it builds, but I don't really have an easy way to test it at the moment. ...so a careful review would be helpful.

dpharris:
Is there a way I can test this ... I do have some Tivas.
David

kiwi64ajs:
I think I have a Tiva with an Ethernet port, which I have previously used as a CAN-to-Ethernet gateway. I could dust it off and see if I can load the code.

Is there a Docker that will create the dev environment or a binary I can FLASH, as that might make testing more straightforward, as it's been a long time since I built any of that code… ;)

Alex

balazsracz (approved):

balazsracz:
Looks good to me.
Generally in order to test this you would want to build an example project like applications/io_board/targets/[the one for your tiva, either 123 or 129]. In the main.cxx near the top, make sure you enable SNIFF_ON_SERIAL.
Then plug in the DEBUG port to your computer. Use JMRI to interact with the device as if it were a usb-can adapter. You should see one device on the LCC network. It should work including for large amount of data transfer (like loading the CDI).

bakerstu:
> Generally in order to test this you would want to build an example project like applications/io_board/targets/[the one for your tiva, either 123 or 129]. In the main.cxx near the top, make sure you enable SNIFF_ON_SERIAL.
> Then plug in the DEBUG port to your computer. Use JMRI to interact with the device as if it were a usb-can adapter. You should see one device on the LCC network. It should work including for large amount of data transfer (like loading the CDI).

This test will not exercise a "mode" change, which is where the change is.

PR #707, opened 2023-04-22
Adds hex_to_string helper function to format_utils.

balazsracz:


src/utils/format_utils.cxx
```diff
@@ -242,6 +242,58 @@ string string_to_hex(const string &arg)
...
+        return b - 'a' + 10;
+    }
+    if ('A' <= b && b <= 'F')
+    {
+        return b - 'A' + 10;
+    }
+    return 0xff;
+}
+
+ssize_t hex_to_string(
```
bakerstu, line 262: I don't see any instance where the return value may be negative. Should we make the return type size_t and avoid any need to cast the "len" type?

src/utils/format_utils.cxx
```diff
@@ -242,6 +242,58 @@ string string_to_hex(const string &arg)
...
+        }
+        else
+        {
+            output->push_back(b);
+            b = 0;
+            next_high = true;
+            continue;
+        }
+    }
+    return ofs;
```
bakerstu, line 294: Looks like this could just as easily be "return len;" since there are no non-return breaks in the for loop above. That would also allow the ofs to be declared as the first input parameter to the for loop, local scope.

bakerstu (approved):

src/utils/format_utils.cxx
balazsracz: done

src/utils/format_utils.cxx
balazsracz: done

PR #713, opened 2023-05-06
Adds 9bit reception support to the TI uart drivers.

balazsracz:
Implements a 9-bit reception mode for Tiva and CC32xx UART drivers.
The nine-bit reception works by reading two byte payloads from the fd for each incoming byte, LSB-first. The second byte contains the parity bit. The special mode is enabled with an ioctl, and only works for RX, not for TX.

### DIFF
--- a/include/freertos/tc_ioctl.h
+++ b/include/freertos/tc_ioctl.h
@@ -53,7 +53,6 @@
 /// Two stop bits
 #define TCSTOPTWO   IO(TERMIOS_IOC_MAGIC, 0xF9)
 
-
 /// Argument is a Notifiable* pointer. This notifiable will be invoked when all
 /// bytes have completed transferring and the transmit engine is idle.
 #define TCDRAINNOTIFY   IOW(TERMIOS_IOC_MAGIC, 0xE0, 4)
--- a/include/freertos/tc_ioctl.h
+++ b/include/freertos/tc_ioctl.h
@@ -46,10 +46,13 @@
 #define TCPAREVEN   IO(TERMIOS_IOC_MAGIC, 0xF2)
 #define TCPARONE    IO(TERMIOS_IOC_MAGIC, 0xF3)
 #define TCPARZERO   IO(TERMIOS_IOC_MAGIC, 0xF4)
+/// Use 9-bit reception mode
+#define TCNINEBITRX IO(TERMIOS_IOC_MAGIC, 0xF5)
 /// One stop bit
 #define TCSTOPONE   IO(TERMIOS_IOC_MAGIC, 0xF8) 
 /// Two stop bits
 #define TCSTOPTWO   IO(TERMIOS_IOC_MAGIC, 0xF9)
+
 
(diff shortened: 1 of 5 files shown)


include/freertos/tc_ioctl.h
```diff
@@ -46,11 +46,14 @@
...
 #define TCPARONE    IO(TERMIOS_IOC_MAGIC, 0xF3)
 #define TCPARZERO   IO(TERMIOS_IOC_MAGIC, 0xF4)
+/// Use 9-bit reception mode
+#define TCNINEBITRX IO(TERMIOS_IOC_MAGIC, 0xF5)
 /// One stop bit
 #define TCSTOPONE   IO(TERMIOS_IOC_MAGIC, 0xF8) 
 /// Two stop bits
 #define TCSTOPTWO   IO(TERMIOS_IOC_MAGIC, 0xF9)
 
+
```
bakerstu, line 56: Extra whitespace or intentional?

src/freertos_drivers/ti/CC32xxUart.cxx
```diff
@@ -289,14 +297,34 @@ void CC32xxUart::interrupt_handler()
...
-            rxBuf->signal_condition_from_isr();
+            else
+            {
+                // parity error bit is moved to the ninth bit, then two bytes
+                // are written to the buffer.
+                long bit9 = (data & 0x200) >> 1;
+                data &= 0xff;
+                data |= bit9;
+                rxBuf->put((uint8_t *)&data, 2);
+                rxBuf->signal_condition_from_isr();
```
bakerstu, line 314: The if/else [overrun] logic for signal_condition_from_isr() is not the same for both 9 and 8 bit data paths. I think technically either is actually fine, but we should be consistent. Pick one and modify the other to match.

src/freertos_drivers/ti/TivaUart.cxx
```diff
@@ -218,14 +219,34 @@ void TivaUart::interrupt_handler()
...
+            }
+            else
+            {
+                // parity error bit is moved to the ninth bit, then two bytes
+                // are written to the buffer.
+                long bit9 = (data & 0x200) >> 1;
+                data &= 0xff;
+                data |= bit9;
+                rxBuf->put((uint8_t *)&data, 2);
+                rxBuf->signal_condition_from_isr();
```
bakerstu, line 236: Same here.

bakerstu (approved):

src/freertos_drivers/ti/CC32xxUart.cxx
balazsracz: fixed

include/freertos/tc_ioctl.h
balazsracz: fixed

src/freertos_drivers/ti/TivaUart.cxx
balazsracz: fixed

PR #719, opened 2023-07-09
Esp32 Bootloader updates from comments

atanisoft:
Fixing comments from https://github.com/bakerstu/openmrn/pull/536.

### DIFF
--- a/src/freertos_drivers/esp32/Esp32BootloaderHal.hxx
+++ b/src/freertos_drivers/esp32/Esp32BootloaderHal.hxx
@@ -56,15 +56,15 @@
 
 // Enable streaming support for the bootloader
 #define BOOTLOADER_STREAM
-// Set the buffer size to half the sector size to minimize the flash writes.
+
+#ifndef WRITE_BUFFER_SIZE
+// Set the buffer size to half of the sector size to minimize the flash writes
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/esp32/Esp32BootloaderHal.hxx
```diff
@@ -399,13 +424,16 @@ uint16_t flash_complete(void)
         if (res != ESP_OK)
         {
             LOG_ERROR("[Bootloader] Failed to update the boot partition!");
+            return openlcb::Defs::ERROR_FIRMWARE_CSUM;
         }
     }
-    else if (res == ESP_ERR_OTA_VALIDATE_FAILED)
+    else
```
balazsracz, line 430: consider moving this else above the then part by inverting the condition. Then the entire function should look like this:
```
res = (call api)
if (res != ESP_OK) {
  log error
  return error code
}
res = (call api)
if (res != ESP_OK) {
  log error
  return error code
}
```
which saves on indent and is much easier to read.

balazsracz (approved):

src/freertos_drivers/esp32/Esp32BootloaderHal.hxx
atanisoft: Done, also removed calls to `ESP_ERROR_CHECK_WITHOUT_ABORT` which are redundant since the error cases already include this information.

PR #726, opened 2023-08-05
Workaround for stm32can stuck in error passive state

balazsracz:
Under some circumstances the stm32 CAN controller can get stuck in error passive state and never hit bus off state even though there is no other bus participant. This PR changes the queuing and error handling logic to detect this state and drop the transmit queue entries.

The specific condition is when there is an operational can transceiver, but there is no other bus participant, meaning that all transmit attempts end up with no-ack error. In this situation the TEC rises to 128, the device goes error passive, but the TEC does not get incremented any further.

Removes flag clear statements for error flags in the ESR register, because these flags are not writeable (can not be cleared in software).

### DIFF
--- a/src/freertos_drivers/st/Stm32Can.cxx
+++ b/src/freertos_drivers/st/Stm32Can.cxx
@@ -219,6 +219,19 @@ void Stm32Can::disable()
  */
 void Stm32Can::tx_msg()
 {
+    // If we are error passive, the last transmission ended with an error, and
+    // there are no free TX mailboxes, then we flush the input queue. This is a
+    // workaround because the STM32 CAN controller can get stuck in this state
+    // and never get to bus off if the TX attempts end up with no-ack (meaning
+    // the controller is alone on the bus).
+    if ((CAN->ESR & CAN_ESR_EPVF) && ((CAN->ESR & CAN_ESR_LEC_Msk) != 0) &&
+        ((CAN->TSR & (CAN_TSR_TME0 | CAN_TSR_TME1 | CAN_TSR_TME2)) == 0))
+    {
+        txBuf->flush();
+        txBuf->signal_condition();
+        return;
+    }
+
     /* see if we can send anything out */
     struct can_frame *can_frame;
 
--- a/src/freertos_drivers/st/Stm32Can.cxx
+++ b/src/freertos_drivers/st/Stm32Can.cxx
@@ -448,31 +448,35 @@ void Stm32Can::sce_interrupt_handler()
         /* error interrupt has occured */
         CAN->MSR |= CAN_MSR_ERRI; // clear flag
 
+        bool cancel_queue = false;
+        
         if (CAN->ESR & CAN_ESR_EWGF)
         {
             /* error warning condition */
             state_ = CAN_STATE_BUS_WARNING;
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/st/Stm32Can.cxx
```diff
@@ -219,6 +219,19 @@ void Stm32Can::disable()
...
+    // there are no free TX mailboxes, then we flush the input queue. This is a
+    // workaround because the STM32 CAN controller can get stuck in this state
+    // and never get to bus off if the TX attempts end up with no-ack (meaning
+    // the controller is alone on the bus).
+    if ((CAN->ESR & CAN_ESR_EPVF) && ((CAN->ESR & CAN_ESR_LEC_Msk) != 0) &&
+        ((CAN->TSR & (CAN_TSR_TME0 | CAN_TSR_TME1 | CAN_TSR_TME2)) == 0))
+    {
+        txBuf->flush();
+        txBuf->signal_condition();
+        return;
```
bakerstu, line 232: In this case, should we actually return, or should we allow the data to be queued up in the space created. I think it probably doesn't matter, and the return is okay. The data is likely to get stale anyways.

src/freertos_drivers/st/Stm32Can.cxx
```diff
@@ -435,31 +448,35 @@ void Stm32Can::sce_interrupt_handler()
         /* error interrupt has occured */
         CAN->MSR |= CAN_MSR_ERRI; // clear flag
 
+        bool cancel_queue = false;
+        
         if (CAN->ESR & CAN_ESR_EWGF)
         {
             /* error warning condition */
             state_ = CAN_STATE_BUS_WARNING;
-            CAN->ESR &= ~CAN_ESR_EWGF;
```
bakerstu, line 442: Are we sure we want to remove this flag clearing, and the two below?

bakerstu (approved):

src/freertos_drivers/st/Stm32Can.cxx
balazsracz: I'll add this from the git commit to the PR description:
> Removes flag clear statements for error flags in the ESR register, because these flags are not writeable (can not be cleared in software).

src/freertos_drivers/st/Stm32Can.cxx
balazsracz: There are two invariants that we care about and should hold:
- there shall be space in the txBuf for new frames to enter. We need this so that the stack is happy. This will be true after the flush and signal.
- there shall be frames in the controller's TX mailboxes (no free tx mailbox). This is necessary so that when the cable is plugged in, the retries will eventually reach a tx success (even though the frame is outdated). The tx success will then exit the condition in two ways: the TEC will decrement and there will be an empty tx mailbox.

Once we are in here, there is no free tx mailbox, so return makes sense.
Another CAN driver (https://github.com/bakerstu/openmrn/blob/master/src/freertos_drivers/ti/TivaCan.cxx#L184) does the same, after flushing the queue they do nothing.

I guess we could rotate in the newer frames into some TX mailbox, but this would be pretty complex logic, and the frames in the TX mailbox would still be outdated when they do get transmitted, just somewhat less outdated.

PR #730, opened 2023-08-17
Ensure tick_disable() upon timeout; fix locking.

### LINKED ISSUE #729: BitBangI2C tick_disable() on timeout
If a timeout occurs, tick_disable() may not get called.

bakerstu:
Fixes #729.

### DIFF
--- a/src/freertos_drivers/common/BitBangI2C.hxx
+++ b/src/freertos_drivers/common/BitBangI2C.hxx
@@ -40,7 +40,7 @@
 
 /// Consolidate all the state machine enumerations for easy operator
 /// overloading.
-class BitBangI2CStates
+class BitBangI2CStates : protected Atomic
 {
 protected:
     /// Default constructor.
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/common/BitBangI2C.hxx
```diff
@@ -730,6 +730,7 @@ inline int BitBangI2C<HW>::transfer(struct i2c_msg *msg, bool stop)
         // be true on first entry at start to "reset" the bus to a known state.
         // On a timeout, it may not have been reset back to false.
         stop_ = false;
+        HW::tick_disable();
```
bakerstu, line 733: Should we wrap this tick_disable() call in an atomic holder? Same for the tick_enable() just above.

balazsracz: Yes, good point. We should. In fact we should protect also the writes to the class variables like stop_.

On the tick_enable side we have to move the semaphore flush loop higher up and put an atomic across all the lines from `msg_ = ...` until `tick_enable()`. 
We have to verify that after tick_disable() no pending interrupt will be executed. (It's not enough to pause the timer, we have to clear any pending interrupt too. If we stop by masking, we have to mask in the NVIC, not in the hardware.)

balazsracz (approved):

src/freertos_drivers/common/BitBangI2C.hxx
bakerstu: Fixed.

src/freertos_drivers/common/BitBangI2C.hxx
```diff
@@ -40,7 +40,7 @@
 
 /// Consolidate all the state machine enumerations for easy operator
 /// overloading.
-class BitBangI2CStates
+class BitBangI2CStates : protected Atomic
```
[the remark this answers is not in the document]
bakerstu: Fixed.

PR #736, opened 2023-08-30
Fix test makefiles and test target

balazsracz:
This PR fixes a variety of annoying bugs related to how our makefiles were running tests, cleans up unnecessary stuff, and adds separate `cov` and `test` targets under openmrn to be able to control whether we compile and link with or without -ftest-coverage. 

Structural changes:
- Deletes old and defunct `openmrn/targets/test` and `etc/test.mk`. These were unused.
- Replaces it with a fork of `cov`, skipping the coverage compilation options.

Bugfixes:
- Fixes bug where doing a make test in an application target directory fails after a clean by not finding things like `lib/libopenlcb.a`. Adding the appropriate dependency makes this compile now.
- Fixes bug where making a file change in an openmrn source would rebuild just that file, but not the test binaries when using make -jN.
- Adds -Wno-maybe-uninitialized to test compilation, because newest GCC throws an error on gtest-all.c

Cleanups:
- Removes `lib` from core_target (openmrn/targets/foo) SUBDIRS. We don't need this, because we never want to build a lib/liblib.a -- but that's exactly what was happening.
- Deletes all unnecessary openmrn/target/*/lib/Makefile files.
- Fixes outdated documentation comments
- Fixes weird/confused includes on targets/{cov,test}/Makefile


bakerstu (approved):
Thanks for cleaning this up. I look forward to using the coverage.

Issue #711, opened 2023-04-29
Possible future compiler problem with MultiConfiguredPC.hxx: std::allocator::construct & std::allocator::destroy going away


RobertPHeller:
When compiling with -std=gnu++20 (the default for ESP-IDF-V5.0.1), MultiConfiguredPC.hxx gets (fatal) compiler errors involving the use of std::allocator::construct & std::allocator::destroy.  Not a problem now if -std=gnu++14, but will probably be a problem someday...

bakerstu:
As a general note, we don't typically like to use the "gnu" language version options. "gnu" tends to pull in components that may be experimental and/or not portable to other non GCC compilers, which we also support. So we prefer -std=c++14 rather than -std=gnu++14.

atanisoft:
@bakerstu Unless things have changed, ESP-IDF depends on a few GNU extensions. ESP-IDF defaults to using whichever of these works first: gnu++20 gnu++2a gnu++17 gnu++14

PR #737, opened 2023-09-06
Fix MultiConfiguredPC.hxx compilation with C++20

atanisoft:
Due to the C++20 standard removing the construct and destruct methods from std::allocator, it was necessary to adjust the code to use std::allocator_traits instead which has provided these methods since C++11.

Fixes https://github.com/bakerstu/openmrn/issues/711

### DIFF
--- a/src/openlcb/MultiConfiguredPC.hxx
+++ b/src/openlcb/MultiConfiguredPC.hxx
@@ -41,6 +41,7 @@
 #include "openlcb/RefreshLoop.hxx"
 #include "utils/ConfigUpdateListener.hxx"
 #include "utils/ConfigUpdateService.hxx"
+#include "utils/Debouncer.hxx"
 #include "utils/format_utils.hxx"
 
 namespace openlcb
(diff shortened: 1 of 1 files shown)


src/openlcb/MultiConfiguredPC.hxx
```diff
@@ -138,10 +138,20 @@ public:
         ConfigUpdateService::instance()->register_update_listener(this);
         producedEvents_ = new EventId[size * 2];
         std::allocator<debouncer_type> alloc;
+// C++20 standard removed the construct and destruct methods from
+// std::allocator which are now available in std::allocator_traits requiring
+// passing in the allocator to these methods.
+#if __cplusplus >= 202002L
```
balazsracz, line 144: We don't need to make this code conditionalized. std::allocator_traits exists since c++11, and we are using c++14 currently (never intend to go back to below c++11).

src/openlcb/MultiConfiguredPC.hxx
```diff
@@ -138,10 +138,20 @@ public:
...
+        std::allocator_traits<std::allocator<debouncer_type>> alloc_traits;
```
balazsracz, line 145: alloc_traits is a static class, so we don't need to make an instance. Use this pattern:
```
using atraits = std::allocator_traits<std::allocator<debouncer_type>>;
...
atraits::construct(alloc, ...)
```

src/openlcb/MultiConfiguredPC.hxx
```diff
@@ -138,10 +138,20 @@ public:
...
+#endif // __cplusplus >= 202002L
         debouncers_ = alloc.allocate(size_);
```
balazsracz, line 147: let's do this through the traits as well, like `atraits::allocate(alloc, size_);`

src/openlcb/MultiConfiguredPC.hxx
```diff
@@ -138,10 +138,20 @@ public:
         ConfigUpdateService::instance()->register_update_listener(this);
         producedEvents_ = new EventId[size * 2];
         std::allocator<debouncer_type> alloc;
+// C++20 standard removed the construct and destruct methods from
```
balazsracz, line 141: this comment is not needed to be added here; let's put it into the PR description and git log will remember it for us.

atanisoft:
@balazsracz PTAL I've tested the updated code with C++20 locally, can you test with C++14?

balazsracz:
Verified that this compiles correctly with the 2018-Q4 armgcc.

PR #743, opened 2023-10-01
Adds support for producer/consumer identified message in the CallbackEventHandler

balazsracz:
When the user registers this callback, we proxy the producer and consumer identified messages to it. This is useful for a node that needs to discover remote state and sends out identify producer / consumer queries for doing so.


src/openlcb/CallbackEventHandler.hxx
```diff
@@ -69,6 +69,22 @@ public:
     /// be used to figure out which bit this is and whether it is on or off.
     typedef std::function<EventState(const EventRegistryEntry &registry_entry,
         EventReport *report)> EventStateHandlerFn;
+    /// This function (signature) is called when a producer/consumer identified
+    /// message arrives.
+    ///
+    /// @param mti MTI_PRODUCER_IDENTIFIED_UNKNOWN or
+    /// MTI_CONSUMER_IDENTIFIED_UNKNOWN. The state bits can be read from
```
bakerstu, line 76: Can we not pass in the correct MTI based on pre-decoding the state of the report?

bakerstu (approved):

src/openlcb/CallbackEventHandler.hxx
balazsracz: I think we don't want to. This way you can test for producer / consumer with a single ==. Otherwise the test is much more complicated.

PR #746, opened 2023-11-12
Limit the amount of input we read into memory in arduino sketches.

balazsracz:
Adds a pushback mechanism to not read faster from a serial bridge than how fast the stack, or a CAN-bus output can sink packets. This is especially important for USBSerial inputs, which can supply data as fast as we are willing to read it.

The implementation relies on two mechanisms:
- setting the gridconnect converter to keep track of the CAN frames and only let a fixed number of them (5) exist at any time in the CAN hub.
- adding a notifiable to the SerialBridge that keep track of the last txt buffer we sent to the gridconnect bridge. We do not read more bytes from the serial port in the read loop until the previous buffer is processed by the gridconnect parser.

### DIFF
--- a/arduino/OpenMRNLite.h
+++ b/arduino/OpenMRNLite.h
@@ -117,8 +117,6 @@ constexpr UBaseType_t OPENMRN_TASK_PRIORITY = ESP_TASK_TCPIP_PRIO - 1;
 
 #endif
 
-OVERRIDE_CONST(gridconnect_bridge_max_incoming_packets, 5);
-
 namespace openmrn_arduino
 {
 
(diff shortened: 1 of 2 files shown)


atanisoft (approved):

arduino/OpenMRNLite.h
```diff
@@ -116,6 +117,8 @@ constexpr UBaseType_t OPENMRN_TASK_PRIORITY = ESP_TASK_TCPIP_PRIO - 1;
 
 #endif
 
+OVERRIDE_CONST(gridconnect_bridge_max_incoming_packets, 5);
```
atanisoft, line 120: should this be moved to OpenMRNLite.cpp or does it not matter that it is included via header?

balazsracz: good catch, fixed.

PR #748, opened 2023-11-14
Adds TinyUsb based CDC virtual Com port driver with an STM32 port.

balazsracz:
- Adds TinyUsb to path.mk
- Adds a target directory to compile it for the stm32f072xb MCU
- Adds a common base class for a TinyUsb based USB stack with CDC device driver
- Instantiates this for STM32
- Adds a CAN-USB application target for the STM32F072xB.

The TinyUSBCdc driver is select-enabled. It does not use a Serial base class, because the read and write fifo's are provided by the TinyUsb codebase. We are exercising the application reads and writes directly from those fifos; this saves one unnecessary copy of the data in memory. (There are still two memcpy's; one in the interrupt to the fifo, one in the application code to the application buffer.)

### DIFF
--- a/applications/usb_can/targets/freertos.armv6m.stm32f072cb-gol/HwInit.cxx
+++ b/applications/usb_can/targets/freertos.armv6m.stm32f072cb-gol/HwInit.cxx
@@ -45,6 +45,7 @@
 #include "os/OS.hxx"
 #include "Stm32Uart.hxx"
 #include "Stm32Can.hxx"
+#include "Stm32UsbCdc.hxx"
 #include "Stm32EEPROMEmulation.hxx"
 #include "hardware.hxx"
 
@@ -67,6 +68,10 @@ static Stm32Can can0("/dev/can0");
 static Stm32EEPROMEmulation eeprom0("/dev/eeprom", 512);
 
 const size_t EEPROMEmulation::SECTOR_SIZE = 2048;
+
+/** USB-Serial instance */
+static Stm32UsbCdc serUSB0("/dev/serUSB0");
+
 
 extern "C" {
 
--- a/applications/usb_can/targets/freertos.armv6m.stm32f072cb-gol/HwInit.cxx
+++ b/applications/usb_can/targets/freertos.armv6m.stm32f072cb-gol/HwInit.cxx
@@ -278,4 +278,4 @@ void usb_interrupt_handler(void)
     dcd_int_handler(0);
 }
 
-}
+} // extern "C"
--- a/applications/usb_can/targets/freertos.armv6m.stm32f072cb-gol/HwInit.cxx
+++ b/applications/usb_can/targets/freertos.armv6m.stm32f072cb-gol/HwInit.cxx
@@ -154,6 +154,35 @@ static void clock_setup(void)
     RCC->CFGR = (RCC->CFGR & (~RCC_CFGR_SW)) | RCC_CFGR_SW_PLL;
     while (!((RCC->CFGR & RCC_CFGR_SWS) == RCC_CFGR_SWS_PLL))
         ;
+
+    RCC_PeriphCLKInitTypeDef  PeriphClkInitStruct = {0};
+    /* Select HSI48 as USB clock source */
+    PeriphClkInitStruct.PeriphClockSelection = RCC_PERIPHCLK_USB;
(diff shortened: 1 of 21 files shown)


src/freertos_drivers/tinyusb/TinyUsbCdcImpl.hxx
```diff
@@ -0,0 +1,436 @@
...
+ * Same VID/PID with different interface e.g MSC (first), then CDC (later) will possibly cause system error on PC.
+ *
+ * Auto ProductID layout's Bitmap:
+ *   [MSB]         HID | MSC | CDC          [LSB]
+ */
+#define _PID_MAP(itf, n)  ( (CFG_TUD_##itf) << (n) )
+#define USB_PID           (0x4000 | _PID_MAP(CDC, 0) | _PID_MAP(MSC, 1) | _PID_MAP(HID, 2) | \
+                           _PID_MAP(MIDI, 3) | _PID_MAP(VENDOR, 4) )
+
+#define USB_VID   0xCafe
```
bakerstu, line 253: Why mixing upper and lower case hex characters?

src/freertos_drivers/tinyusb/TinyUsbCdcImpl.hxx
```diff
@@ -0,0 +1,436 @@
...
+#define USB_BCD   0x0200
+
+// Invoked when received GET DEVICE DESCRIPTOR
+// Application return pointer to descriptor
+uint8_t const *tud_descriptor_device_cb(void) {
```
bakerstu, line 258: Incorrect open brace style, here and below.

src/freertos_drivers/tinyusb/TinyUsbCdcImpl.hxx
```diff
@@ -0,0 +1,436 @@
...
+            break;
+        default:
+        case 0:
+            /* we don't support any exceptions */
+            break;
+    }
+    return retval;
+}
+
+inline void TinyUsbCdc::rx_available() {
```
bakerstu, line 216: Incorrect open brace style, here and below.

bakerstu (approved):

applications/usb_can/targets/freertos.armv6m.stm32f072cb-gol/HwInit.cxx
```diff
@@ -0,0 +1,281 @@
...
+}
+
+extern void dcd_int_handler(int);
+
+void usb_interrupt_handler(void)
+{
+    dcd_int_handler(0);
+}
+
+}
```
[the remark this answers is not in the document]
balazsracz: done

src/freertos_drivers/tinyusb/TinyUsbCdc.hxx
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+#ifndef _FREERTOS_DRIVERS_TINYUSB_TINYUSBCDC_HXX_
+#define _FREERTOS_DRIVERS_TINYUSB_TINYUSBCDC_HXX_
+
+#include "Devtab.hxx"
+#include "os/OS.hxx"
+#include "utils/Singleton.hxx"
+
+class TinyUsbCdc : public Node, public Singleton<TinyUsbCdc> {
```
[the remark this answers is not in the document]
balazsracz: It's not super easy to create multiple CDC endpoints. For reference, the TivaUsbCdc driver doesn't support it either. These are the steps that need to be taken before we can instantiate a second object:
- change the tinyusbconfig.h (which will have a memory footprint effect)
- change the USB descriptors
- update how the thread is started / running
- update all callbacks to be routed to the right instance
- figure out what happens on the host OS if there is no second object but the descriptor says there should be

Ultimately I think having two CDCs is a feature request that's not high priority, and at this point it's more important to prevent accidentally creating two instances, which the singleton will do fine.

src/freertos_drivers/tinyusb/TinyUsbCdcImpl.hxx
balazsracz: fixed

src/freertos_drivers/tinyusb/TinyUsbCdcImpl.hxx
balazsracz: fixed all whitespace

PR #750, opened 2023-11-19
Add common app test coverage rules.

bakerstu:


etc/app_cov.mk
```diff
@@ -0,0 +1,88 @@
+CUSTOM_EXEC=1
```
balazsracz, line 1: `# Include this file into an application/target/cov/Makefile to build and run coverage for the app.`

balazsracz (approved):

etc/app_cov.mk
bakerstu: Added.

PR #756, opened 2023-12-21
Adds local loopback to TractionThrottle.

balazsracz:
This PR improves behavior when there is more than one TractionThrottle in the same openlcb::node_. Specifically, when these are assigned to the same dst_ locomotive, the Train Node will never send back echoes of the packets. This means that the simultaneous updates are missing.

This PR makes all TractionThrottle objects lined up in a linked list upon creation. When a locomotive control commands (speed, estop, fn) is sent out, a local loopback will walk the linked list and identify if there are any other tractionthrottle objects assigned to the same loco. If so, the message will be handed over to it for update callbacks.

### DIFF
--- a/src/openlcb/TractionThrottle.hxx
+++ b/src/openlcb/TractionThrottle.hxx
@@ -287,17 +287,57 @@ private:
     {
         if (listenConsist_)
         {
-            handler_.wait_for_response(
-                NodeHandle(dst_), TractionDefs::RESP_CONSIST_CONFIG, &timer_);
-            send_traction_message(
-                TractionDefs::consist_del_payload(node_->node_id()));
-            return sleep_and_call(
-                &timer_, TIMEOUT_NSEC, STATE(release_listener_response));
+            // Checks if there is another throttle listening to the same
+            // train. If so, we will not unregister ourselves. The last local
+            // throttle to release will do the unregistering.
+            bool found_other_throttle = false;
+            {
+                AtomicHolder h(LinkedObject<TractionThrottle>::head_mu());
+                for (TractionThrottle *p =
+                         LinkedObject<TractionThrottle>::link_head();
+                     p && !found_other_throttle;
+                     p = p->LinkedObject<TractionThrottle>::link_next())
+                {
+                    if (p == this)
+                    {
+                        // self, ignore
+                        continue;
+                    }
+                    if (p->node_ != node_)
+                    {
+                        // Different virtual node, will get regular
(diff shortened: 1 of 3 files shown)


src/openlcb/TractionThrottle.hxx
```diff
@@ -653,16 +710,109 @@ private:
...
+    void loopback_traction_message(Buffer<GenMessage>* b)
+    {
+        auto rb = get_buffer_deleter(b);
+        // Walks all TractionThrottle objects.
+        TractionThrottle *p = nullptr;
+        do
+        {
+            {
+                // finds next instance that's interesting
+                AtomicHolder h(LinkedObject<TractionThrottle>::head_mu());
```
bakerstu, line 743: I'm not crazy about holding an atomic section for the loop through of an unbounded list. Is there an existing mutex we could use instead? What about a task scheduler lock instead?

bakerstu (approved):

src/openlcb/TractionThrottle.hxx
balazsracz: Under linux an Atomic is actually a mutex.
Under freertos it is a critical section, which means that (on an M3/M4) only priority zero interrupts can go through.
So the important question is how many iterations are there here and how long do they take.
- For almost all devices there are at most two iterations. Throttles typically have one or two such objects.
- For command stations there can be up to 64 such objects. (63 for NCE, 31 for XNet). Definitely not unbounded, because each tractionthrottle is about 100 bytes or more.

One iteration is just a pointer chasing and comparing some memory bytes. I expect it to be say 0.5 to 1 usec (few dozen instructions).
As a result, I expect a non-p0 interrupt jitter of about 0.05 to 0.1 millisecond to be caused by this.

Non-P0 interrutps are things like CAN-bus (1 msec per interrupt) or UART (where we have a FIFO , unless in NCE, where the interrupt is one per msec as well).

It's not super practical to replace this with a mutex or a scheduler lock:
- LinkedObject<> is used elsewhere, if we want to replace the locking, we need to copy the code or impact every other use as well (examples are Executor<> and StateFlowWithQueue<>)
- scheduler lock doesn't exist in linux. We don't have an abstraction for it. It's also unclear what it would do in a multi-core ESP32.
- Another alternative might be to jump to the main executor for sending outgoing packets, but this is also not exciting, because it would probably break all unit tests. It also doesn't guarantee that the links don't change if an object gets constructed or deleted on another thread.

PR #758, opened 2023-12-29
Limits end to end buffering in gridconnect bridge.

balazsracz:
When a dispatcher is used to send traffic to a gridconnect bridge, an infinite number of packets were allocated in the dispatcher when more than one client was connected. This is a typical situation when a hub is operating.

By adding a limitedpool to the bridge object, we can ensure that the dispatcher is blocked when the output queue is full and pushes back.

This PR respects the config option gridconnect_bridge_max_outgoing_packets.


bakerstu (approved):
Very nice. Simple and flexible.

PR #759, opened 2023-12-29
Adds configuration options for TCP sockets, tuning for Hub app.

balazsracz:
Adds several tuning parameters for TCP sockets:
- SO_SNDBUF
- SO_RCVBUF
- TCP_NOTSENT_LOWAT

Refactors where the fd socket options are set to a single place.
Applies some reasonable (but not great) values for the Hub application.
Switches the Hub application to use select(). Tunes the parameters to limit buffer size in the hub.
Fixes a crashing bug where a GridconnectBridge was deleted too early when packets owned by its pool were still in flight.


bakerstu (approved):
Interesting logging mechanism. I'll have to squirrel that away for possible later use.

PR #766, opened 2024-01-20
Adds a helper function to decode a 14-bit 9.2.1.1 address into address type and raw address.

balazsracz:


bakerstu (approved):
Should we not have a unit test for this?

balazsracz:
> Should we not have a unit test for this?

done

PR #769, opened 2024-02-01
Adds support for DCC extended accessories 

balazsracz:
- Adds helper functions to generate dcc extended accessory packets in dcc::Packet
- Adds well-known event range (from the current draft eventidentifiers standard)
- Adds DccExtAccyConsumer, which listens to events and turns them into DCC packets.

Unlike the basic dcc accy consumer, the extended consumer does not remember whatthe last set state was. This is because it takes a lot more memory (2 kbytes), and it's a lot less useful, since different decoders may be interpreting this last set command very differently (such as a signal, a turntable, or a timed turnout decoder).


src/openlcb/DccAccyConsumer.hxx
```diff
@@ -260,6 +260,150 @@ private:
...
+    }
+
+private:
+    /// Send the actual accessory command.
+    void send_accy_command() override
+    {
+        dcc::TrackIf::message_type *pkt;
+        mainBufferPool->alloc(&pkt);
+        pkt->data()->add_dcc_ext_accessory(dccAddress_, aspect_);
+        pkt->data()->packet_header.rept_count = 3;
```
atanisoft, line 398: should we consider moving the repeat count to a constant since we generally request three repeats in most packets?

balazsracz: This is not actually super easy. There is no common base class for these two. I would have to put the constant into a global namespace, which is not nice. 
The repeat count is a bit field in the packet header, so it is pretty static. It is a two-bit field, so supports a repetition of 1 to 4.

bakerstu: I think the way that we currently have implemented provides good flexibility.

bakerstu (approved):

PR #770, opened 2024-02-02
Update libify to support IDF export with symlinks

balazsracz:
Adds two features to libify.sh:
- adds support for creating or updating an OpenMRNIDF repository. This means skipping arduino-specific stuff, like examples or OpenMRNLite.h, rpelacing the readme.md and library.properties files with a different one.
- Adds support for creating relative symlinks. This allows using OpenMRNIDF that links back to openmrn using relative symlinks -- these can be checked into git for example.
- Adds the few toplevel files for OpenMRNIDF like cmakefile


arduino/libify.sh
```diff
@@ -118,20 +161,22 @@ function copy_dir() {
...
+        cp -faxr ${USE_LINK} ${INVERSE_DIR}${OPENMRNPATH}/$2 .
     fi
 
     popd >/dev/null
 }
 
-copy_file . arduino/{library.json,library.properties,keywords.txt,README.md,LICENSE,CONTRIBUTING.md}
-copy_dir . arduino/examples
+if [ "x$TARGET_IDF" == "x" ]; then
+    copy_file . arduino/{library.json,library.properties,keywords.txt,README.md,LICENSE,CONTRIBUTING.md}
```
atanisoft, line 177: we should probably have `README.md,LICENSE,CONTRIBUTING.md` on the IDF side, perhaps a slightly different version though.

arduino/libify.sh
```diff
@@ -8,12 +8,14 @@
...
-    echo 'usage: libify.sh path/to/arduino/library/output path/to/openmrn [-f] [-l]'
+    echo 'usage: libify.sh path/to/arduino/library/output path/to/openmrn [-f] [-l] [-i] [-r]'
     echo 'exports OpenMRN code as an arduino library.'
-    echo 'example: libify.sh ~/Arduino/libraries/OpenMRN .. -l'
+    echo 'example: libify.sh ~/Arduino/libraries/OpenMRNLite .. -l'
     echo '(options must come after the path specification)'
     echo '-f will erase the target library before exporting.'
     echo '-l will create symlinks instead of copying files.'
+    echo '-i will create OpenMRNIDF repository instead of arduino.'
+    echo '-r will create relative symlinks. OPenMRNPath has to be a relative path from the library export directoty back to openmrn, starting with ../'
```
atanisoft, line 18: OPenMRNPath -> OpenMRNPath

arduino/libify.sh
```diff
@@ -118,20 +161,22 @@ function copy_dir() {
...
 }
 
-copy_file . arduino/{library.json,library.properties,keywords.txt,README.md,LICENSE,CONTRIBUTING.md}
-copy_dir . arduino/examples
+if [ "x$TARGET_IDF" == "x" ]; then
+    copy_file . arduino/{library.json,library.properties,keywords.txt,README.md,LICENSE,CONTRIBUTING.md}
+    copy_dir . arduino/examples
+fi
 
 copy_file src arduino/OpenMRNLite.{h,cpp} arduino/CDIXMLGenerator.hxx \
```
atanisoft, line 181: Should we consider moving `arduino/OpenMRNLite.{h,cpp}` to the arduino block above since these have arduino dependencies?

balazsracz: You have them in your repo. I guess you don't need it and never build it via cmake?

atanisoft: I think they came over as a mass import from the exported tree create by the script. They are excluded from build via cmake: https://github.com/atanisoft/OpenMRNIDF/blob/5.1.0/CMakeLists.txt

arduino/libify.sh
balazsracz: added

arduino/libify.sh
balazsracz: fixed

arduino/libify.sh
balazsracz: done

arduino/libify.sh
balazsracz: I didn't add contributing.md because you don't have it, and the arduino version references the OpenMRNLite library in the text. So we'd need to first write a new one, but this is really not high priority now.

PR #772, opened 2024-02-02
Fix compiler warnings in openmrn when using new GCC's.

balazsracz:
There are a few places where a newer GCC used by esp-idf emits a warning for OpenMRN code. This PR fixes those.

The fall through warning unfortunately has false positives, but that's still not a good reason to disable it, we rather adjust the comments here to suit the taste of the compiler.


atanisoft (approved):

src/freertos_drivers/esp32/Esp32Gpio.hxx
```diff
@@ -554,112 +554,11 @@ template <class Defs> struct GpioInputPUPD : public GpioInputPin<Defs, true, tru
     {                                                                          \
         static const gpio_num_t PIN_NUM = (gpio_num_t)NUM;                     \
     public:                                                                    \
-        static const gpio_num_t pin()                                          \
+        static gpio_num_t pin()                                          \
```
atanisoft, line 557: please align the trailing \

bakerstu (approved):

src/freertos_drivers/esp32/Esp32Gpio.hxx
balazsracz: done

PR #783, opened 2024-02-10
Ensures that we don't start STA wifi when AP-only is requested.

balazsracz:
This PR fixes codepaths that accidentally started STA mode wifi when AP-only mode was requested.

### DIFF
--- a/src/freertos_drivers/esp32/Esp32WiFiManager.cxx
+++ b/src/freertos_drivers/esp32/Esp32WiFiManager.cxx
@@ -845,6 +845,12 @@ void Esp32WiFiManager::start_mdns_system()
 
 void Esp32WiFiManager::on_station_started()
 {
+    if (!(wifiMode_ == WIFI_MODE_STA || wifiMode_ == WIFI_MODE_APSTA))
+    {
+        LOG(INFO, "[Station] Did not want station.");
+        return;
+    }
+
     // Set the generated hostname prior to connecting to the SSID
     // so that it shows up with the generated hostname instead of
     // the default "Espressif".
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -1287,15 +1307,8 @@ StateFlowBase::Action Esp32WiFiManager::WiFiStackFlow::init_wifi()
...
-    wifi_mode_t requested_wifi_mode = parent_->wifiMode_;
-    if (parent_->wifiMode_ == WIFI_MODE_AP)
-    {
-      // override the wifi mode from AP only to AP+STA so we can perform wifi
-      // scans on demand.
-      requested_wifi_mode = WIFI_MODE_APSTA;
-    }
     // Set the requested WiFi mode.
-    ESP_ERROR_CHECK(esp_wifi_set_mode(requested_wifi_mode));
+    ESP_ERROR_CHECK(esp_wifi_set_mode(parent_->wifiMode_));
```
atanisoft, line 1311: Have you tested that when in AP mode that WiFi scan works? I had tested it long ago and found that it is required to use APSTA mode for scan to work this the comment in the section that was removed above.

balazsracz:
That's fine, but then if you want to scan, you need to start in APSTA mode,
not in AP mode.

atanisoft:
That breaks existing user code that expects WiFi scan to work when operating in AP mode. It's not as simple as changing the mode since that would attempt to start STA mode and rightfully crash during init. I'm fine with not creating the AP / STA interface up front but it would still be good to support scanning while in AP mode as part of a reconfiguration tool.

PR #788, opened 2024-03-10
BLE Basic Infrastructure

bakerstu:
Adds manufacturer-independent infrastructure for a BLE-based connection protocol, and OpenLCB's implemnentation classes and constants.

- Adds abstract classes for defining BLE Advertisements and Services.
- Adds concrete implementation of these with encoding the specific advertisements for GATT-based central and peripheral for OpenLCB, including the necessary UUID constants.
- Adds a Port implementation for DirectHub for sending text data (e.g. gridconnect frames) over BLE.

Since this is all manufacturer-independent, there are no external dependencies involved, but it also means that in itself, this basic infrastructure does not work. A significant amount of additional code is needed to tie this todether with a specific manufacturer's BLE stack.

Misc changes:
- Makes SimpleStack be able to expose the PIP bytes via a direct API
- Fixes a compiler warning on LinkedDataBuffer


src/openlcb/BLEHubPort.hxx
```diff
@@ -60,6 +62,10 @@ public:
     ///
     using SendFunction = std::function<void(const uint8_t *data, size_t len)>;
 
+    /// How big can a single attribute write be? ESP's BLE implementation says
+    /// 600 bytes. We keep some buffer.
+    static constexpr size_t MAX_BYTES_PER_WRITE = 500;
```
atanisoft: fyi, ble core specification is 512b. I'd likely drop to ~450b as a default (or move to a config constant)

balazsracz: There is apparently also an MTU negotiation happening after the connection gets established. So this likely won't work just as is. Then the negiatiated MTU is not the maximum value that you can transmit, because there is some protocol headers as well.

balazsracz (approved):

PR #791, opened 2024-08-05
Bug fixes in DataBuffer

balazsracz:
There were several corner cases in appending to a Datauffer that were incorrectly handled and causing crashes in DirectHub.
This PR fixes bugs and adds tests including several fuzz tests for databuffers.

### DIFF
--- a/src/utils/DataBuffer.hxx
+++ b/src/utils/DataBuffer.hxx
@@ -500,12 +500,20 @@ public:
     /// this. This tries to do `*this += o`. It will succeed if o.head() ==
     /// this->tail() and the bytes in these buffers are back to back.
     /// @param o a LinkedDataBuffer with data payload.
+    /// @param add_link creates a tail-to-head link if none exist yet between
+    /// *this and o.head_. This is fundamentally dangerous, do it only if there
+    /// is no shared ownership of this->tail_.
     /// @return true if append succeeded. If false, nothing was changed.
-    bool try_append_from(const LinkedDataBufferPtr &o)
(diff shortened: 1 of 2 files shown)


src/utils/DataBuffer.hxx
```diff
@@ -176,8 +199,13 @@ private:
 
 using DataBufferPtr = std::unique_ptr<DataBuffer, BufferDelete<uint8_t[]>>;
 
+class LinkedDataBufferPtr;
```
atanisoft, line 202: this declaration seems unnecessary since the class body is just below. Any particular reason this is needed here?

src/utils/DataBuffer.hxx
```diff
@@ -456,37 +531,70 @@ public:
...
+    /// is no shared ownership of this->tail_.
     /// @return true if append succeeded. If false, nothing was changed.
-    bool try_append_from(const LinkedDataBufferPtr &o)
+    bool try_append_from(const LinkedDataBufferPtr &o, bool add_link = false)
     {
         if (!o.size())
         {
             return true; // zero bytes, nothing to do.
         }
+        if (!size_) {
```
atanisoft, line 544: brace formatting?

bakerstu (approved):
Please address @atanisoft comments. I agree with them.

src/utils/DataBuffer.hxx
balazsracz: removed

src/utils/DataBuffer.hxx
balazsracz: fixed all

PR #793, opened 2024-08-14
Fixes a standards compliance issue with the alias conflict handler.

balazsracz:
The standard requires that at the time when an alias conflict is detected, the device to issue an AMR frame with the conflicted alias.

This PR adds the missing AMR frame. Adjusts tests to heck for this frame. Fixes a bug in the Global AME handler that was unexpectedly emitting AMD frames for removed aliases, because the cache entry was not correctly wiped when the remove was processed, and the iteration also picked up such deleted entries.

### DIFF
--- a/src/openlcb/IfCan.cxx
+++ b/src/openlcb/IfCan.cxx
@@ -156,10 +156,15 @@ public:
          * into the empty_alias_allocator() instead to be picked up by the
          * alias allocator flow.  */
         g_alias_use_conflicts++;
-        if_can()->local_aliases()->remove(alias);
-        return exit();
+        alias_ = alias;
+        return allocate_and_call(
+            if_can()->frame_write_flow(), STATE(send_release_alias));
     }
 
+    /// Sends an RID with the alias alias_.
+    ///
+    /// This is done in response to a CID frame seen from the bus with the
+    /// given alias.
     Action send_reserved_alias()
     {
         auto *b = get_allocation_result(if_can()->frame_write_flow());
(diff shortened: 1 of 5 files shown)


src/openlcb/IfCan.cxx
```diff
@@ -170,6 +175,37 @@ class AliasConflictHandler : public CanFrameStateFlow
         return exit();
     }
 
+    /// Sends an AMR (alias mapping release) frame with the alias alias_ and
+    /// the loca node ID that we have for it. Then deletes the mapping from the
```
atanisoft, line 179: loca -> local

bobjacobsen (approved):
> The standard requires that at the time when an alias conflict is detected, the device to issue an AMR frame with the conflicted alias.
>
> This PR adds the missing AMR frame. 

The standard says to send the AMR if the alias is for a node in Permitted state.  As I read it, if that alias hasn't been used yet to go to Permitted state, it's not really "in use" and you just drop it.  

Does this PR do that?

src/openlcb/IfCan.cxx
balazsracz: fixed

balazsracz:
> > The standard requires that at the time when an alias conflict is detected, the device to issue an AMR frame with the conflicted alias.
> > This PR adds the missing AMR frame.
> 
> The standard says to send the AMR if the alias is for a node in Permitted state. As I read it, if that alias hasn't been used yet to go to Permitted state, it's not really "in use" and you just drop it.
> 
> Does this PR do that?

There was indeed a bug there. Thanks for flagging!
I added a unittest to verify this property.

PR #795, opened 2024-08-16
Fixes race conditions in HubDeviceSelect.

balazsracz:
These were problematic when the shutdown/close happened on a different thread than the executor. Caused flakey tests.

Updates a test helper to use heap memory because that plays nicer with Valgrind's data race checker.

### DIFF
--- a/src/utils/HubDeviceSelect.hxx
+++ b/src/utils/HubDeviceSelect.hxx
@@ -173,8 +173,8 @@ public:
         , dst_(dst)
         , skipMember_(skip_member)
     {
-        this->start_flow(STATE(allocate_buffer));
         set_limit_input(shouldThrottle_);
+        this->start_flow(STATE(allocate_buffer));
     }
 
     void set_limit_input(bool should_throttle)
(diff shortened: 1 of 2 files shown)


src/utils/HubDeviceSelect.hxx
```diff
@@ -340,29 +341,30 @@ public:
...
-                executor()->sync_run([this, &completed]()
-                                 {
-                                     if (barrier_.is_done()) completed = true;
-                                 });
-            }
+        }
+        bool completed = false;
+        while (!completed)
+        {
+            executor()->sync_run([this, &completed]() {
```
bakerstu, line 362: Brace style.

src/utils/HubDeviceSelect.hxx
```diff
@@ -485,10 +492,27 @@ protected:
...
+            fd = fd_;
+            if (fd < 0) {
+                return;
+            }
             fd_ = -1;
         }
+        // This is a workaround that sometimes my linux kernel gets stuck in
+        // ::read when I closed the fd like this, even though the fd is
+        // O_NONBLOCK.
+        executor()->add(new CallbackExecutable([this, fd]() {
```
bakerstu, line 511: Brace style.

bakerstu (approved):

src/utils/HubDeviceSelect.hxx
balazsracz: done

PR #798, opened 2024-08-23
Handles unhandled-addressed-messages by generating an OIR reply.

balazsracz:
This PR fixes a stadnards compliance issue. The standard requires that when an addressed message arrives for a node which is not equipped for handling said message (including any unknown but addressed MTI arriving), then an Optional IOnteraction Rejected (OIR) message has to be sent back to the originating node, identifying that this target node is not able to handle that message.

- Adds a "fallback handler" to the Dispatcher object. The fallback handler is invoked when no handler matched on the incoming message.
- Implements a handler that responds with an OIR for all local nodes. Registers this handler as the fallback handler for the MessageDIspatcher in IfCan and IfTcp. This OIR response will be suppressed for incoming OIR and TDE errors to avoid infinite loops of error messages. (In a normal setup OpenMRN does not listen to these messages. Without this case two OpenMRN nodes will have an infinite message loop chatting with each other sending OIR messages back and forth.)

Fixes some bugs exposed by this behavior.
- The TCP parser had forgotten to clear the message it was filling, causing messages to unexpectedly be marked as addressed messages.
- A number of tests had issues when OIR messages were generated, this PR fixes these issues as well. Sometimes the MTI was mistyped, sometimes a message was verified at the CAN level but not actually processed at the destination node.

### DIFF
--- a/src/openlcb/IfImpl.hxx
+++ b/src/openlcb/IfImpl.hxx
@@ -293,6 +293,14 @@ public:
             // Destination is not a local node.
             return release_and_exit();
         }
+        auto mti = message()->data()->mti;
+        if (mti == Defs::MTI_OPTIONAL_INTERACTION_REJECTED ||
+            mti == Defs::MTI_TERMINATE_DUE_TO_ERROR) {
+            // We don't generate an OIR for an incoming error report, as this
+            // generally would cause an infinite bouncing of error reports back
+            // and forth between two OpenMRN nodes.
+            return release_and_exit();
+        }
         return allocate_and_call(
             iface()->addressed_message_write_flow(), STATE(fill_oir));
     }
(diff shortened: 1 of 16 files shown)


src/openlcb/IfImpl.hxx
```diff
@@ -266,6 +266,78 @@ private:
     If::VNodeMap::Iterator it_;
 #endif
 };
+
+/// Message handler that is registered as a fallback handler in the interface's
+/// message dispatcher. This means that all incoming messages that were not
+/// matching the MTI / mask of any existing instantiated handler will end up
+/// here.
+///
+/// THe standard requires that when an addressed message is not handled by a
```
atanisoft, line 275: `THe` -> `The`

src/openlcb/IfImpl.hxx
```diff
@@ -266,6 +266,78 @@ private:
...
+        if (!message()->data()->dstNode)
+        {
+            // Destination is not a local node.
+            return release_and_exit();
+        }
+        return allocate_and_call(
+            iface()->addressed_message_write_flow(), STATE(fill_oir));
+    }
+
+    Action fill_oir()
```
bakerstu, line 300: Could we add Doxygen header blocks to this method, and the others that are undocumented in this file?

src/openlcb/IfImpl.hxx
```diff
@@ -266,6 +266,78 @@ private:
...
+            error_payload(Defs::ERROR_UNIMPLEMENTED, inm->mti));
+        iface()->addressed_message_write_flow()->send(rb.release());
+        return release_and_exit();
+    }
+};
+
+/// This is an addressed message handler that catches errors (OIR and TDE) but
+/// ignores them. While this sounds not super useful, it ensures that these
+/// messages never trigger the UnhandledAddressedMessageHandler.
+class NullErrorHandler : public IncomingMessageStateFlow
```
bakerstu, line 317: This does seem super not useful because the party that initiated the interaction will not be notified, it just times out. Should we have some way of notifying the initiator of the interaction? Perhaps at least file this as an enhancement request issue?

bakerstu (approved):

src/openlcb/IfImpl.hxx
balazsracz: done

src/openlcb/IfImpl.hxx
balazsracz: done

src/openlcb/IfImpl.hxx
balazsracz: I actually deleted this handler and replaced it with an if (mti != OIR && mti != TDE) in the other handler. This has the same result but takes less RAM and FLASH space.

PR #800, opened 2024-10-20
ESP-IDF CMakeLists

bakerstu:
The goal of this pull request is to no longer require a separate OpenMRNIDF repository.

- Provides a CMakeLists file such that openmrn can be included as a registered idf component as part of a standard ESP-IDF application build.
- Provides an esp-idf specific include path with appropriate files linked to it.
- Makes a few minor code tweaks to support the ESP-IDF build.


etc/esp-idf/CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+
+idf_component_register(SRCS "${SRCS}"
+                       INCLUDE_DIRS "${INCLUDE_DIRS}"
+                       REQUIRES "${IDF_DEPS}")
+
+###############################################################################
+# Suppress compilation warnings in OpenMRN
+###############################################################################
+
+target_compile_options(${COMPONENT_LIB} PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-Wno-volatile>)
```
bakerstu, line 157: Does this need to be PUBLIC?

etc/esp-idf/CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+    $ENV{OPENMRNPATH}/src/freertos_drivers/esp32/Esp32HardwareI2C.cxx
+    $ENV{OPENMRNPATH}/src/freertos_drivers/esp32/Esp32HardwareTwai.cxx
+    $ENV{OPENMRNPATH}/src/freertos_drivers/esp32/Esp32Ledc.cxx
+    $ENV{OPENMRNPATH}/src/freertos_drivers/esp32/Esp32SocInfo.cxx
+    $ENV{OPENMRNPATH}/src/freertos_drivers/esp32/Esp32WiFiManager.cxx
+
+    $ENV{OPENMRNPATH}/src/openlcb/AliasAllocator.cxx
+    $ENV{OPENMRNPATH}/src/openlcb/AliasCache.cxx
+    $ENV{OPENMRNPATH}/src/openlcb/BLEAdvertisement.cxx
+    $ENV{OPENMRNPATH}/src/openlcb/BLEService.cxx
```
bakerstu, line 30: These BLE files have not been merged back to master yet. I tested this on a branch with BLE, which is why they are here. We need to also add the src/ble subdirectory files to the SRCS list.

etc/esp-idf/CMakeLists.txt
atanisoft: I opted for `PUBLIC` primarily to suppress a warning from Executor.hxx/cxx showing up if you include the hxx in your project. We might be able to get away with `INTERFACE` though, I'll check and confirm.

src/utils/SocketClient.cxx
```diff
@@ -232,7 +232,7 @@ std::unique_ptr<SocketClientParams> SocketClientParams::from_static(
     std::unique_ptr<DefaultSocketClientParams> p(new DefaultSocketClientParams);
     p->staticHost_ = std::move(hostname);
     p->staticPort_ = port;
-    return std::move(p);
+    return p;
```
atanisoft, line 235: I'm curious why this was changed, any particular reason?

etc/esp-idf/README.md
```diff
@@ -0,0 +1,7 @@
+# ESP-IDF Build Instructions
+The ESP-IDF has its own CMake based build environment for application developers. Using the standard ESP-IDF application build environment, The openmrn/etc/esp-idf directory can be linked into the ESP application "components" directory.
+```
+cd <application path>/components
+ln -s <openmrn path>/etc/esp-idf/ openmrn
+```
+This will result in openmrn being compiled as a registered idf component with the name "openmrn".
```
atanisoft, line 7: We should also document using idf_component.yml which is the approach recommended by Espressif for dependency management. 

A version of this could be:
```
dependencies:
  OpenMRN:
    version: {branch, hash, tag, etc}
    git: https://github.com/bakerstu/openmrn.git
    path: etc/esp-idf
```
The `path` line overrides where `idf.py` looks for the CMakeLists.txt within the referenced repository (it defaults to root of the repository).

If you include `override_path` it will skip looking at the repository and use a local copy instead, which can use environment variables, such as: `${OPENMRNPATH}/etc/esp-idf`.

The  `version` line is optional, when not present `idf.py build` will collect the latest git hash and record it in `dependencies.lock`. If there is a newer version available in git it will present the following during build:
```
-- Building ESP-IDF components for target esp32s3
NOTICE: 
Following dependencies have new versions available:
Dependency "OpenMRNIDF": "5d7b0b3157e3411b3f71ccfacedaef6cb1202bff" -> "63e2b2e3f6ba34719f50194240113247793ac3ac"
Consider running "idf.py update-dependencies" to update your lock file.
```

etc/esp-idf/CMakeLists.txt
```diff
@@ -0,0 +1,162 @@
...
+    esp_adc
+    esp_app_format
+    esp_system
+    esp_wifi
+    espcoredump
+    hal
+    heap
+    vfs
+    mdns
+    bt)
```
atanisoft, line 147: please add pthread since it is used in OSSelectWakeup for VFS registration.

etc/esp-idf/CMakeLists.txt
```diff
@@ -0,0 +1,162 @@
...
+)
+
+if(CONFIG_ENABLE_ARDUINO_DEPENDS)
+    list(APPEND SOURCE_DIRS src)
+endif()
+
+set(IDF_DEPS
+    app_update
+    bootloader_support
+    driver
```
atanisoft, line 137: I'm going to put this on my update list for ESP-IDF v5.3 support updates (hopefully a relatively small update), this `driver` component was split up and can be replaced with:

- esp_driver_gpio
- esp_driver_i2c

but there is currently a compilation failure in `Esp32HardwareI2C.hxx` due to a dependency on the older driver format. I'll take care of migrating to the newer driver code and cleanup of this specific line as optional based on ESP-IDF version (pre-5.3 it will include driver and current Esp32HardwareI2C code, 5.3+ it will depend on the newer I2C driver).

etc/esp-idf/CMakeLists.txt
```diff
@@ -0,0 +1,162 @@
...
+
+idf_component_register(SRCS "${SRCS}"
+                       INCLUDE_DIRS "${INCLUDE_DIRS}"
+                       REQUIRES "${IDF_DEPS}")
+
+###############################################################################
+# Warn all and make all warnings into errors in OpenMRN
+###############################################################################
+target_compile_options(${COMPONENT_LIB} PRIVATE -Wall)
+target_compile_options(${COMPONENT_LIB} PRIVATE -Werror)
```
atanisoft, line 157: ESP-IDF (v5.3) defaults to passing in `-Wall -Werror=all -Wextra` by default. Should we include `-Wextra` here as well?

etc/esp-idf/README.md
```diff
@@ -0,0 +1,7 @@
+# ESP-IDF Build Instructions
```
atanisoft, line 1: please consider adding an `idf_component.yml` in this directory with the following contents:
```
dependencies:
  idf: "<5.3"
```
This will restrict compilation of OpenMRN to ESP-IDF version 5.2.x or earlier by default until I can get a PR setup for the driver updates in ESP-IDF v5.3

include/esp-idf/can_ioctl.h
```diff
@@ -0,0 +1 @@
+../freertos/can_ioctl.h
```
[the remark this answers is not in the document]
balazsracz: no we should generally have more smaller PRs. If something can be done in a separate PR, it should be done in a separate PR.

src/utils/SocketClient.cxx
balazsracz: I believe this is covered by C++ copy elision. If it weren't covered, the new code will not compile, since copying a unique_ptr is not possible.

etc/esp-idf/README.md
balazsracz: We don't use the IDF component manager for several reasons:
- it becomes impossible to compile the code when the espressif servers are not reachable over the internet (have you tried compiling code while you are on an airplane?)
- the compilation becomes not deterministic from the git hash

Both of these are big problems for a professional project. Reproducible compilation results are essential for finding bugs and retroactive maintenance. We are talking about 10y+ of support needs in this product domain, not a 6-month product development cycle like many other domains.

Having said that, it should be possible to use ther component manager if someone wants it. It is just a requirement for the compilation and dependencies to work correctly if someone chooses not to use it.

atanisoft: Reproducible builds is a great call-out and has a rather simple solution. Limit your external dependencies (beyond esp-idf itself) *AND* embed the github link and hash (which typically can not change) and the build system will only use that specific version for external dependencies.

As for offline (airplane) usage, this is doable with small adjustments to the idf_component.yml and pre-staging the external dependencies locally. In the idf_component.yml file you would leverage the `override_path` or `path` element to shortcut the lookup of the component with a local path to where the dependency actually lives. You might need to remove version/git/etc elements though (I forget now on that).

Cloning the dependencies locally into a components tree (as git submodules) would also work for this, optionally adding the following snippet to the top level CMakeLists.txt file:
```
list(APPEND EXTRA_COMPONENT_DIRS "../common_components/Filesystem/")
```
This allows simplified dependencies mapping in component / main CMakeLists.txt files.

atanisoft: also note that the above snippet would *ONLY* check the ESP-IDF version and does not specifically deal with dependencies. If the ESP-IDF version check does not match the declared version it will abort the build with a message for the user to fix their ESP-IDF version selection.

etc/esp-idf/CMakeLists.txt
atanisoft: As noted elsewhere, we can reduce this to `PRIVATE`

etc/esp-idf/CMakeLists.txt
```diff
@@ -0,0 +1,162 @@
+set(SRCS
+    $ENV{OPENMRNPATH}/src/dcc/dcc_constants.cxx
```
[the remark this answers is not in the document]
atanisoft: `${COMPONENT_DIR}` can also be used for referencing the location of CMakeLists.txt

src/utils/SocketClient.cxx
bakerstu: The reason for the change is that the std::move statement is redundant. According to the language standards, a return is always a move. The compiler was producing a "redundant move in return statement" warning.

etc/esp-idf/CMakeLists.txt
bakerstu: This is required because of a warning generated by the usage of Executor::sequence_. We should revisit this at some point.

etc/esp-idf/CMakeLists.txt
bakerstu: This is resolved by merging the BLE basics branch into master and then merging master into this branch.

etc/esp-idf/CMakeLists.txt
```diff
@@ -0,0 +1,162 @@
...
+###############################################################################
+# Warn all and make all warnings into errors in OpenMRN
+###############################################################################
+target_compile_options(${COMPONENT_LIB} PRIVATE -Wall)
+target_compile_options(${COMPONENT_LIB} PRIVATE -Werror)
+
+###############################################################################
+# Suppress compilation warnings in OpenMRN
+###############################################################################
+target_compile_options(${COMPONENT_LIB} PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-Wno-volatile>)
```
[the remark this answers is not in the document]
bakerstu: Agree that we should revisit this.

etc/esp-idf/CMakeLists.txt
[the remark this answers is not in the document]
bakerstu: I'm actually not in favor of changing this. We have an established history of a top level "config" file that setups up environment variables for a repository. @balazsracz and I have also decided to use a Makefile for the top level project build that invokes CMake. We find this a bit more intuitive as it allows us to simply use the command "make" and then all the CMake magic happens under the hood.

etc/esp-idf/CMakeLists.txt
```diff
@@ -0,0 +1,162 @@
...
+    $ENV{OPENMRNPATH}/src/utils/StringPrintf.cxx
+)
+
+set(INCLUDE_DIRS
+    $ENV{OPENMRNPATH}/src
+    $ENV{OPENMRNPATH}/include/esp-idf
+    $ENV{OPENMRNPATH}/include
+)
+
+if(CONFIG_ENABLE_ARDUINO_DEPENDS)
```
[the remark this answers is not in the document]
bakerstu: Removed.

etc/esp-idf/CMakeLists.txt
bakerstu: Done.

etc/esp-idf/CMakeLists.txt
bakerstu: Noted.

etc/esp-idf/CMakeLists.txt
bakerstu: I changed -Werror to -Werror=all and added -Wextra. Luckily it does not seem to have created any new errors. Phew!

etc/esp-idf/CMakeLists.txt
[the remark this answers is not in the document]
atanisoft: In that case, OpenMRNIDF will continue to live on as it's not feasible to pass in env vars when using idf_component.yml to manage dependencies.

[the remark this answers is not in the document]
bakerstu: I'm revisiting this. It looks like ${COMPONENT_DIR} does not follow the symbolic link to the origin path. I'm exploring options.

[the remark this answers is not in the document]
bakerstu: @atanisoft I pushed a possible solution. Can you take a look?

etc/esp-idf/README.md
bakerstu: @atanisoft I've added an idf_component.yml file. This causes a warning for my non-managed build, but we can choose to ignore it. Can you have a look and provide some feedback?

bakerstu:
> Overall looks good.
> 
> One other warning that we should review:
> 
> ```
> [1112/1124] Building CXX object esp-idf/main/CMakeFiles/__idf_main.dir/main.cpp.obj
> In file included from .../main/ThermalMonitor.hxx:20,
>                  from .../main/main.cpp:55:
> ..../OpenMRNIDF/src/utils/Fixed16.hxx: In member function 'float Fixed16::to_float() const':
> ..../OpenMRNIDF/src/utils/Fixed16.hxx:274:17: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
>   274 |         return *reinterpret_cast<float *>(&f);
> ```
> 
> I'm not 100% certain how we should handle this one. This is coming up in ESP-IDF v5.3 (tested with v5.3.1 locally).

Probably this is intentional. If so, we should consider an inline disabling of this warning using the push/pop method.

etc/esp-idf/idf_component.yml
```diff
@@ -0,0 +1,6 @@
+## IDF Component Manager Manifest File
+dependencies:
+  OpenMRN:
```
atanisoft, line 3: Close, we can comment out this OpenMRN section as it would create a circular dependency which may trip up the dependency logic in esp-idf.

etc/esp-idf/README.md
atanisoft: commented on the file above, we should just include the basic snippet listed above to only restrict the idf version check.

src/freertos_drivers/esp32/Esp32Gpio.hxx
```diff
@@ -35,7 +35,7 @@
 #ifndef _DRIVERS_ESP32GPIO_HXX_
 #define _DRIVERS_ESP32GPIO_HXX_
 
-#include "freertos_drivers/arduino/GpioWrapper.hxx"
+#include "freertos_drivers/common/GpioWrapper.hxx"
```
atanisoft, line 38: should we update `arduino/libify.sh` since it handles copying from common -> arduino? [ref](https://github.com/bakerstu/openmrn/blob/eb8ed8b3d71d12b9123d4f0a3a0467693c612d96/arduino/libify.sh#L212-L219)

etc/esp-idf/README.md
atanisoft: @bakerstu it looks like no variant of declaration in idf_component.yml will work correctly with the current version of the idf component manager that I can find. The `path: etc/esp-idf` bit tries to force the git checkout to that path rather than reference that path for where to find the top level directory.

I suspect it only applies to when publishing the component to their registry and not for when consuming it. For now let's drop the file and I'll see if I can find a suitable way to make it work correctly or log an issue for them to fix it

src/freertos_drivers/esp32/Esp32Gpio.hxx
atanisoft: in libify they are copied to arduino from common.

atanisoft: let's track this via an issue to unify the libify bits so we can avoid ifdef etc

PR #809, opened 2025-01-16
Memory Config Timeout Support (Datagrams)

bakerstu:
- Added support on the server side.
- Added support on the client side.
- Added unit test coverage.

### DIFF
--- a/src/openlcb/MemoryConfig.hxx
+++ b/src/openlcb/MemoryConfig.hxx
@@ -110,12 +110,12 @@ public:
     virtual address_t max_address() = 0;
 
     /// Write data to the address space.
-    /// @param source memory space offset address to write to
+    /// @param destination memory space offset address to write to
     /// @param data data to write
     /// @param len length of write data in bytes
     /// @param error Error code that can be passed back. if != 0, the operation
     ///        has failed. If the operation needs to be continued, then set
-    ///        error to ERROR_EAGAIN, and later call the Notifiable to continue.
+    ///        error to ERROR_AGAIN, and later call the Notifiable to continue.
     /// @param again notify() when the caller should call the write again, with
     ///        the offset (source) adjusted with the previously returned bytes.
     /// @return the number of bytes successfully written (before hitting the end
@@ -134,7 +134,7 @@ public:
     /// @param len length of requested read data in bytes
     /// @param error Error code that can be passed back. if != 0, the operation
     ///        has failed. If the operation needs to be continued, then set
-    ///        error to ERROR_EAGAIN, and later call the Notifiable to continue.
+    ///        error to ERROR_AGAIN, and later call the Notifiable to continue.
     /// @param again notify() when the caller should call the read again, with
     ///        the offset (source) adjusted with the previously returned bytes.
     /// @return the number of bytes successfully read (before hitting the end of
--- a/src/openlcb/MemoryConfig.hxx
+++ b/src/openlcb/MemoryConfig.hxx
@@ -375,7 +375,7 @@ protected:
 
     Action ok_response_sent() OVERRIDE
     {
-    if (!response_.empty())
+        if (!response_.empty())
         {
             return allocate_and_call(
                 STATE(client_allocated), dg_service()->client_allocator());
--- a/src/openlcb/MemoryConfig.hxx
+++ b/src/openlcb/MemoryConfig.hxx
@@ -70,7 +70,7 @@ public:
     /** This error code signals that the operation was only partially
      * completed, the again notify was used and will be notified when the
      * operation can be re-tried). */
-    static const errorcode_t ERROR_AGAIN = 0x3FFF;
+    static constexpr errorcode_t ERROR_AGAIN = 0x3FFF;
 
     /// Specifies which node the next operation pertains. If it returns false,
     /// the operation will be rejected by "unknown memory space ID".
@@ -86,14 +86,16 @@ public:
     }
 
     /// Get the read timeout. Default is no timeout.
-    /// @return The read timeout to reply with.
+    /// @return The read timeout to reply with, for example
+    ///         DatagramDefs::TIMEOUT_16 for 16 seconds.
     virtual DatagramDefs::Flag get_read_timeout()
     {
         return DatagramDefs::TIMEOUT_NONE;
     }
 
     /// Get the write timeout. Default is no timeout.
-    /// @return The write timeout to reply with.
+    /// @return The write timeout to reply with, for example
+    ///         DatagramDefs::TIMEOUT_16 for 16 seconds.
     virtual DatagramDefs::Flag get_write_timeout()
     {
         return DatagramDefs::TIMEOUT_NONE;
@@ -109,15 +111,21 @@ public:
      */
     virtual address_t max_address() = 0;
 
-    /// Write data to the address space.
+    /// Write data to the address space. Called by the memory config service for
+    /// incoming write requests.
     /// @param destination memory space offset address to write to
     /// @param data data to write
     /// @param len length of write data in bytes
-    /// @param error Error code that can be passed back. if != 0, the operation
-    ///        has failed. If the operation needs to be continued, then set
-    ///        error to ERROR_AGAIN, and later call the Notifiable to continue.
-    /// @param again notify() when the caller should call the write again, with
-    ///        the offset (source) adjusted with the previously returned bytes.
+    /// @param error The output argument for the error code. If the operation
+    ///        succeeded, sets *error to zero. If the operation failed, sets
+    ///        *error to non-zero. If the operation needs to be continued, then
+    ///        sets error to ERROR_AGAIN, and later calls the Notifiable when
+    ///        ready to continue. The caller preset *error to zero, such that
+    ///        the callee is not required to also set it to zero upon "success".
+    /// @param again Used only when *error was set to ERROR_AGAIN. Calls
+    ///        again->notify() when the operation is ready to be continued. The
+    ///        caller should then call the write again, with the offset (source)
+    ///        adjusted with the previously returned bytes.
     /// @return the number of bytes successfully written (before hitting the end
     ///         of the space)
     virtual size_t write(address_t destination, const uint8_t *data, size_t len,
@@ -128,15 +136,21 @@ public:
         DIE("Unimplemented");
     }
 
-    /// Read data from the address space.
+    /// Read data from the address space. Called by the memory config service
+    /// for incoming read requests.
     /// @param source memory space offset address to read from
     /// @param dst location to fill in with read data
     /// @param len length of requested read data in bytes
-    /// @param error Error code that can be passed back. if != 0, the operation
-    ///        has failed. If the operation needs to be continued, then set
-    ///        error to ERROR_AGAIN, and later call the Notifiable to continue.
-    /// @param again notify() when the caller should call the read again, with
-    ///        the offset (source) adjusted with the previously returned bytes.
+    /// @param error The output argument for the error code. If the operation
+    ///        succeeded, sets *error to zero. If the operation failed, sets
+    ///        *error to non-zero. If the operation needs to be continued, then
+    ///        sets error to ERROR_AGAIN, and later calls the Notifiable when
+    ///        ready to continue. The caller preset *error to zero, such that
+    ///        the callee is not required to also set it to zero upon "success".
+    /// @param again Used only when *error was set to ERROR_AGAIN. Calls
+    ///        again->notify() when the operation is ready to be continued. The
+    ///        caller should then call the read again, with the offset (source)
+    ///        adjusted with the previously returned bytes.
     /// @return the number of bytes successfully read (before hitting the end of
     ///         the space)
     virtual size_t read(address_t source, uint8_t *dst, size_t len,
@@ -925,11 +939,13 @@ private:
         char c = 0;
         response_.assign(response_len, c);
         inline_respond_ok(static_cast<uint8_t>(DatagramClient::REPLY_PENDING) |
-            static_cast<uint8_t>(space->get_read_timeout()));
+            (static_cast<uint8_t>(space->get_read_timeout()) &
+            static_cast<uint8_t>(DatagramDefs::TIMEOUT_MASK)));
         return call_immediately(STATE(try_read));
     }
 
-    Action try_read() {
+    Action try_read()
+    {
         MemorySpace *space = get_space();
         int read_len = get_read_length();
         address_t address = get_address();
@@ -1008,7 +1024,8 @@ private:
         }
         currentOffset_ = 0;
         inline_respond_ok(static_cast<uint8_t>(DatagramClient::REPLY_PENDING) |
-            static_cast<uint8_t>(space->get_write_timeout()));
+            (static_cast<uint8_t>(space->get_write_timeout()) &
+            static_cast<uint8_t>(DatagramDefs::TIMEOUT_MASK)));
         return call_immediately(STATE(try_write));
     }
 
--- a/src/openlcb/MemoryConfig.hxx
+++ b/src/openlcb/MemoryConfig.hxx
@@ -114,7 +114,10 @@ public:
     /// Write data to the address space. Called by the memory config service for
     /// incoming write requests.
     /// @param destination memory space offset address to write to
-    /// @param data data to write
+    /// @param data data to write. The callee may capture and begin processing
+    ///        this data but should only acount for what it has captured in the
+    ///        return value once it has been successfully written, and not just
+    ///        "in process".
     /// @param len length of write data in bytes
     /// @param error The output argument for the error code. If the operation
     ///        succeeded, sets *error to zero. If the operation failed, sets
--- a/src/openlcb/MemoryConfig.hxx
+++ b/src/openlcb/MemoryConfig.hxx
@@ -924,8 +924,8 @@ private:
         currentOffset_ = 0;
         char c = 0;
         response_.assign(response_len, c);
-        inline_respond_ok(
-            DatagramClient::REPLY_PENDING | space->get_read_timeout());
+        inline_respond_ok(static_cast<uint8_t>(DatagramClient::REPLY_PENDING) |
+            static_cast<uint8_t>(space->get_read_timeout()));
         return call_immediately(STATE(try_read));
     }
 
@@ -1007,8 +1007,8 @@ private:
             return respond_reject(Defs::ERROR_INVALID_ARGS);
         }
         currentOffset_ = 0;
-        inline_respond_ok(
-            DatagramClient::REPLY_PENDING | space->get_write_timeout());
+        inline_respond_ok(static_cast<uint8_t>(DatagramClient::REPLY_PENDING) |
+            static_cast<uint8_t>(space->get_write_timeout()));
         return call_immediately(STATE(try_write));
     }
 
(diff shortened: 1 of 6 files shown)


src/openlcb/MemoryConfig.hxx
```diff
@@ -94,23 +109,36 @@ public:
...
     virtual address_t max_address() = 0;
 
-    /** @returns the number of bytes successfully written (before hitting end
-     * of space). If *error is set to non-null, then the operation has
-     * failed. If the operation needs to be continued, then sets error to
-     * MemorySpace::ERROR_AGAIN, and calls the Notifiable @param again when a
-     * re-try makes sense. The caller should call write once more, with the
-     * offset adjusted with the previously returned bytes. */
+    /// Write data to the address space.
+    /// @param source memory space offset address to write to
```
atanisoft, line 113: source -> destination

src/openlcb/MemoryConfig.hxx
```diff
@@ -94,23 +109,36 @@ public:
...
+    /// @param data data to write
+    /// @param len length of write data in bytes
+    /// @param error Error code that can be passed back. if != 0, the operation
+    ///        has failed. If the operation needs to be continued, then set
+    ///        error to ERROR_EAGAIN, and later call the Notifiable to continue.
```
atanisoft, line 118: ERROR_EAGAIN -> ERROR_AGAIN

src/openlcb/MemoryConfig.hxx
```diff
@@ -94,23 +109,36 @@ public:
...
-     * should call read once more, with the offset adjusted with the previously
-     * returned bytes. */
+
+    /// Read data from the address space.
+    /// @param source memory space offset address to read from
+    /// @param dst location to fill in with read data
+    /// @param len length of requested read data in bytes
+    /// @param error Error code that can be passed back. if != 0, the operation
+    ///        has failed. If the operation needs to be continued, then set
+    ///        error to ERROR_EAGAIN, and later call the Notifiable to continue.
```
atanisoft, line 137: ERROR_EAGAIN -> ERROR_AGAIN

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -529,8 +538,17 @@ private:
         if (responseCode_ & DatagramClient::OPERATION_PENDING)
         {
             isWaitingForTimer_ = 1;
+            // Retrieve the timeout flags
```
atanisoft, line 541: can/should we consider moving this to an inline method to reduce duplication of this block of code?

atanisoft (approved):
Looks good, will do some testing over this weekend.

@bobjacobsen can you confirm if JMRI is similarly handling timeout hints on datagram responses so I can test through that as well?

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -391,8 +391,17 @@ private:
...
+            uint8_t timeout_flags =
+                (dgClient_->result() >> DatagramClient::RESPONSE_FLAGS_SHIFT) &
+                    DatagramClient::REPLY_TIMEOUT_MASK;
+            // Convert the timeout flags to seconds. Timeout flags == 0 means
+            // no timeout, which we interpret as the OpenLCB intrinsic default
+            // of 3 seconds.
+            long long timeout_nsec = timeout_flags == 0 ?
+                 SEC_TO_NSEC(3) : SEC_TO_NSEC(0x1 << timeout_flags);
```
atanisoft, line 402: should we use the existing constant defined [here](https://github.com/bakerstu/openmrn/blob/1064f5edbe3e526d3eb1c2fb8f596747f07d10ff/src/openlcb/Datagram.cxx#L42)?

atanisoft: The update looks good and satisfies the consolidation of the usage of magic numbers for this use case to a single location. It would be good to similarly remove the referenced constant, likely in a similar manner.

bakerstu: If you look at the solution, I did remove the "magic" number to a commented constexpr

atanisoft:
With the latest code I'm seeing the following warning/deprecation notice:
```
src/openlcb/MemoryConfig.hxx:1011:43: warning: bitwise operation between different enumeration types ‘openlcb::DatagramClient::ResponseFlag’ and ‘openlcb::DatagramDefs::Flag’ is deprecated [-Wdeprecated-enum-enum-conversion]
 1011 |             DatagramClient::REPLY_PENDING | space->get_write_timeout());
```

Not sure how best we should handle this other than using casting to an integer and then back to ResponseFlag.

bakerstu:
> With the latest code I'm seeing the following warning/deprecation notice:
> 
> ```
> src/openlcb/MemoryConfig.hxx:1011:43: warning: bitwise operation between different enumeration types ‘openlcb::DatagramClient::ResponseFlag’ and ‘openlcb::DatagramDefs::Flag’ is deprecated [-Wdeprecated-enum-enum-conversion]
>  1011 |             DatagramClient::REPLY_PENDING | space->get_write_timeout());
> ```
> 
> Not sure how best we should handle this other than using casting to an integer and then back to ResponseFlag.

@atanisoft This is a good find. We should fix this. I can think of three possible solutions:

1. Static cast the operation.
2. Create a custom operator override.
3. Rework the enumerated types to be a bit more consistent and compatible, maybe even built off each other a bit. Possibly consider structs with bitfields.
   1. This would still likely require some custom operator overrides.
   2. One problem with bitfields is that there is no standard bit ordering enforced by the C/C++ standards. It happens that pretty much everyone uses the same order though, so maybe this is not so much an issue.

I very much would like to cleanup these enumerated types. I think they are not very well laid out, and could have been better designed. That said, I think it is too much of a change at this point. Therefore, I prefer solution 1. It is brute force, but at least self documenting in what the operation is doing.

WDYT?

atanisoft:
I'm fine with opting for static_cast for now and seeing what we can do outside this PR as a better fit. I know C++20 has a bit better handling of enums, which may rely on extending standard variable types like uint8_t or similar.

bakerstu:
I added the static_cast for both read and write timeout cases.

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -79,6 +114,26 @@ public:
...
+    {
+        for (size_t i = 0; i < len && (i + offset) < dataContents2_.size(); ++i)
+        {
+            dst[i] = dataContents2_[i + offset];
+        }
+    }
+
+    /// Consume the payload from a write.
+    void AcceptPayload(
+        MemorySpace::address_t offset, const uint8_t* dst, size_t len)
```
balazsracz, line 128: I think this is `src`, not `dst`.

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+
+    //
+    // Pass back timeout "none". Force a timeout (3 seconds).
+    //
+    expect_any_packet();
+    EXPECT_CALL(tSpace_, get_read_timeout())
+        .WillRepeatedly(Return(DatagramDefs::TIMEOUT_NONE));
+    EXPECT_CALL(tSpace_, read(0, _, 64, _, NotifyCapture(&done)))
+        .WillOnce(DoAll(
+            SetArgPointee<3>(0x3FFF),
```
balazsracz, line 231: let's write `ERROR_AGAIN` here

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+            Return(0)));
+    b = invoke_client_no_block(
+        MemoryConfigClientRequest::READ, NodeHandle(TEST_NODE_ID), 0x52);
+    wait();
+    clock.advance(MSEC_TO_NSEC(3200));
```
balazsracz, line 236: colnsider splitting this delay in two:
```
clock.advance(MSEC_TO_NSEC(2900));
wait();
EXPECT_FALSE(clientTwo_.is_waiting());
EXPECT_EQ(MemoryConfigClient::OPERATION_PENDING, b->data()->resultCode);
clock.advance(MSEC_TO_NSEC(400));
wait();
```

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+    wait();
+    EXPECT_EQ(Defs::OPENMRN_TIMEOUT, b->data()->resultCode);
+    ASSERT_EQ(0U, b->data()->payload.size());
+    EXPECT_TRUE(clientTwo_.is_waiting());
+    testing::Mock::VerifyAndClearExpectations(&tSpace_);
+    expect_any_packet();
```
balazsracz, line 242:
```
/// The code here shows that the memory config handler (on the server side) 
/// does not recover from a memory space timing out and never calling `done`
/// notifiable.
```

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+    EXPECT_CALL(tSpace_, read(0, _, 64, _, NotifyCapture(&done)))
+        .WillOnce(DoAll(
+            SetArgPointee<3>(0x3FFF),
+            Return(0)));
+    b = invoke_client_no_block(
+        MemoryConfigClientRequest::READ, NodeHandle(TEST_NODE_ID), 0x52);
+    wait();
+    clock.advance(SEC_TO_NSEC(15));
+    wait();
+    ASSERT_EQ(0U, b->data()->payload.size());
```
balazsracz, line 269: also add:
EXPECT_EQ(MemoryConfigClient::OPERATION_PENDING, b->data()->resultCode);

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+            SetArgPointee<3>(MemoryConfigDefs::ERROR_OUT_OF_BOUNDS),
+            Return(24)));
+    done->notify();
+    wait();
+    ASSERT_EQ(0U, b->data()->payload.size());
+    testing::Mock::VerifyAndClearExpectations(&tSpace_);
+}
+
+// Tests using the MemoryConfigClient to write a payload that fits into one
+// datagram.
```
balazsracz, line 319: i think this comment is wrong

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+    EXPECT_CALL(tSpace_, write(0, _, test_payload.size(), _, _))
+        .WillOnce(DoAll(WithArgs<0, 1, 2>(
+                Invoke(this, &MemoryConfigClientTest::AcceptPayload)),
+            Return(test_payload.size())));
+    auto b = invoke_flow(&clientTwo_, MemoryConfigClientRequest::WRITE,
+        NodeHandle(TEST_NODE_ID), 0x52, 0, test_payload);
+    wait();
+    EXPECT_EQ(0, b->data()->resultCode);
+    ASSERT_EQ(dataContents2_.size(), b->data()->payload.size());
+    EXPECT_FALSE(
```
balazsracz, line 354: `EXPECT_EQ(0, memcmp())` would be better readable

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+    // Pass back timeout "none". Force a timeout (3 seconds).
+    //
+    expect_any_packet();
+    init_data_contents2();
+    EXPECT_CALL(tSpace_, get_write_timeout())
+        .WillRepeatedly(Return(DatagramDefs::TIMEOUT_NONE));
+    EXPECT_CALL(
+        tSpace_, write(0, _, test_payload.size(), _, NotifyCapture(&done)))
+            .WillOnce(DoAll(
+                SetArgPointee<3>(0x3FFF), Return(test_payload.size())));
```
balazsracz, line 369: ERROR_AGAIN

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+    b = invoke_client_no_block(MemoryConfigClientRequest::WRITE,
+        NodeHandle(TEST_NODE_ID), 0x52, 0, test_payload);
+    wait();
+    clock.advance(MSEC_TO_NSEC(3200));
```
balazsracz, line 373: let's add a test for pending here as well after 2900 msec.

src/openlcb/MemoryConfigClient.cxxtest
balazsracz, line 369: I think this should be Return(0) since you are not actually accepting the payload

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+    wait();
+    EXPECT_EQ(Defs::OPENMRN_TIMEOUT, b->data()->resultCode);
+    EXPECT_TRUE(
```
balazsracz, line 376: EXPECT_NE(0, memcmp())

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+    expect_any_packet();
+    EXPECT_CALL(tSpace_, write(_, _, _, _, _)).Times(0);
+    done->notify();
+    wait();
+    EXPECT_TRUE(
+        memcmp(&dataContents2_[0], test_payload.data(), test_payload.size()));
+    testing::Mock::VerifyAndClearExpectations(&tSpace_);
+
+    //
+    // Pass back timeout "32 seconds". Wait 31 seconds to reply.
```
balazsracz, line 389: this replies inline instead of after 31 seconds. You would need to capture done, wait 31 seconds, then call done notifiable.

This would need two expected calls on write(...) with the first one returning 0.

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -131,6 +188,255 @@ TEST_F(MemoryConfigClientTest, create)
...
+    // Pass back timeout "32 seconds". Force a timeout.
+    //
+    expect_any_packet();
+    init_data_contents2();
+    EXPECT_CALL(tSpace_, get_write_timeout())
+        .WillRepeatedly(Return(DatagramDefs::TIMEOUT_32));
+    EXPECT_CALL(
+        tSpace_, write(0, _, test_payload.size(), _, NotifyCapture(&done)))
+            .WillOnce(DoAll(
+                SetArgPointee<3>(0x3FFF), Return(test_payload.size())));
```
balazsracz, line 420: Return(0)

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -402,7 +405,8 @@ private:
 
     Action read_response_timeout()
     {
-        if (responseCode_ & DatagramClient::OPERATION_PENDING)
+        if (responseCode_ & DatagramClient::OPERATION_PENDING ||
+            !timer_.is_triggered())
```
balazsracz, line 409: if the read finishes upon the first attempt, the timer is never started.
If this code is really needed, then we have to call timer_.set_triggered() on that path.

src/openlcb/MemoryConfig.hxx
```diff
@@ -84,6 +84,21 @@ public:
     {
         return true;
     }
+
+    /// Get the read timeout. Default is no timeout.
+    /// @return The read timeout to reply with.
```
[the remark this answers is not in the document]
bakerstu: Added.

src/openlcb/MemoryConfig.hxx
```diff
@@ -94,23 +109,36 @@ public:
      */
     virtual address_t max_address() = 0;
 
-    /** @returns the number of bytes successfully written (before hitting end
-     * of space). If *error is set to non-null, then the operation has
-     * failed. If the operation needs to be continued, then sets error to
-     * MemorySpace::ERROR_AGAIN, and calls the Notifiable @param again when a
-     * re-try makes sense. The caller should call write once more, with the
-     * offset adjusted with the previously returned bytes. */
+    /// Write data to the address space.
```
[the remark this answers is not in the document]
bakerstu: Added.

src/openlcb/MemoryConfigClient.cxxtest
bakerstu: I have revisited this. Both flows are possible, but the program track does actually use your flow. I updated the test to use your flow.

src/openlcb/MemoryConfigClient.cxxtest
bakerstu: Disregard my comment, fixed.

src/openlcb/MemoryConfigClient.hxx
```diff
@@ -540,7 +547,8 @@ private:
 
     Action write_response_timeout()
     {
-        if (responseCode_ & DatagramClient::OPERATION_PENDING)
+        if (responseCode_ & DatagramClient::OPERATION_PENDING ||
+            !timer_.is_triggered())
```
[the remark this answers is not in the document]
bakerstu: @balazsracz I wrote a test for this. I'd like you to review it carefully because I had to make a code change in the MemoryConfigClient, which I did not expect to have to do.

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -289,6 +638,25 @@ TEST_F(MemoryConfigClientTest, unfreeze)
     ASSERT_TRUE(b->data()->done.is_done());
 }
 
+TEST_F(MemoryConfigClientTest, WriteNoReplyPending)
```
bakerstu, line 641: @balazsracz Please review this carefully.

src/openlcb/MemoryConfigClient.cxxtest
```diff
@@ -391,18 +422,27 @@ TEST_F(MemoryConfigClientTest, writeTimeout)
...
+    EXPECT_CALL(tSpace_,
+        write(0, _, test_payload.size(), _, NotifyCapture(&done)))
         .WillOnce(DoAll(WithArgs<0, 1, 2>(
                 Invoke(this, &MemoryConfigClientTest::AcceptPayload)),
             Return(test_payload.size())));
-    b = invoke_flow(&clientTwo_, MemoryConfigClientRequest::WRITE,
-        NodeHandle(TEST_NODE_ID), 0x52, 0, test_payload);
-    wait();
-    clock.advance(SEC_TO_NSEC(31));
+    done->notify();
```
balazsracz, line 441: we shouldn't notify done if there is no ERROR_AGAIN in the *error return.

src/openlcb/MemoryConfigClient.cxxtest
balazsracz: looks good

src/openlcb/MemoryConfigClient.cxxtest
balazsracz: nevermind, i mixed up the code. there is an ERROR_AGAIN a while up.
So then I think we should skip the NotifyCapture() in this expectation.

balazsracz (approved):

PR #812, opened 2025-01-18
Simplifies conditional compilation in the Stm32Uart driver. 

balazsracz:
Simplifies code by reducing the number of places where we need large sequences of ifdef conditionals.
Removes all negative ifdefs.
Removes interrupt handlers defined by the driver.cxx in favor of the interrupt handlers in HwInit.
Removes static interrupt handler; the interrupt handler goes now directly through the object declared in HwInit.

Fixes bug where the shared interrupt enable count was instance local insted of a static variable.


bakerstu (approved):
This works, but I actually suggest a slight tweak of this solution. I would get rid of the "instance" array altogether. The primary reason it existed was to not propagate dependencies between HwInit.cxx and the driver. We've thrown that out the window now, so I would rather have this pattern in HwInit.cxx:

```
Stm32Uart uart0(...);

uart1_interrupt_handler(void)
{
    uart0.interrupt_handler();
}
```

balazsracz:
> This works, but I actually suggest a slight tweak of this solution. I would get rid of the "instance" array altogether. The primary reason it existed was to not propagate dependencies between HwInit.cxx and the driver. We've thrown that out the window now, so I would rather have this pattern in HwInit.cxx:
> 
> ```
> Stm32Uart uart0(...);
> 
> uart1_interrupt_handler(void)
> {
>     uart0.interrupt_handler();
> }
> ```

That's a good idea. Let me do that.

PR #814, opened 2025-01-25
CMake Infrastructure Start

bakerstu:
- Adds the start of some CMake infrastructure
- Starts with unit tests that include a code coverage option with gcov/gcovr
- For now, the GNUMake infrastructure will remain in parallel
  - Will build out more CMake support over time with the eventual goal to replace the existing GNUMake infrastructure.


cmake/test_setup_prologue.cmake
```diff
@@ -0,0 +1,53 @@
...
+FetchContent_Declare(
+    googletest
+    URL https://github.com/google/googletest/archive/refs/tags/${GTEST_TAG}.zip
+)
+
+# On Windows: Prevent overriding the parent project's compiler/linker settings.
+set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+FetchContent_MakeAvailable(googletest)
+
+# This magic is required in order to properly add the "-x" option for the
```
balazsracz, line 30: why don't we just add a compile option -x c++ to the c++ files?

targets/test_cmake/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+cmake_minimum_required(VERSION 3.14)
+
+# Default coverage option to off.
+option(COVERAGE "COVERAGE" OFF)
+
+# Setup directory paths.
+set(OPENMRNPATH ${CMAKE_SOURCE_DIR}/../..)
```
balazsracz, line 7: can you do an abspath on this?

balazsracz (approved):

cmake/test_setup_prologue.cmake
bakerstu: I will keep this as is, except there appears to be an extra "-c" option, which I removed.

PR #817, opened 2025-02-02
Add Automatic Build and Deployment of Doxygen Documentation to GitHub Pages

ngpaladi:
An example of a functional deployment is available at: [https://www.uncommonmodels.com/openmrn/](https://www.uncommonmodels.com/openmrn/) 

Please note that the Pages deployment source must be set to "GitHub Actions" in the repository settings.

### DIFF
--- a/.github/workflows/DoxygenBuild.yml
+++ b/.github/workflows/DoxygenBuild.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Upload artifact
       uses: actions/upload-pages-artifact@v3
       with:
-        path: html/
+        path: doc/html/
     - name: Deploy
       id: deployment
       uses: actions/deploy-pages@v4
--- a/.github/workflows/DoxygenBuild.yml
+++ b/.github/workflows/DoxygenBuild.yml
@@ -1,0 +1,33 @@
+name: Deploy Doxygen Documentation to Pages
+
+on:
+  push:
+    branches: [master] # branch to trigger deployment
+
+jobs:
+  pages:
+    runs-on: ubuntu-latest
+    environment:
+      name: github-pages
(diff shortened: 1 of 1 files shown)


bakerstu:
@ngpaladi This is an excellent contribution. I want to run this by some of the other frequent contributors.

...now of course we should also spend some effort updating the Doxygen to be a bit more up to date.

balazsracz:
I'm not super familiar with actions. This looks good to me and we should have  it.
Stuart, I think you are the only one who can make the change in the settings.

atanisoft:
I've used a similar set of actions but instead had:
```
- name: Setup Pages
        id: pages
        uses: actions/configure-pages@v3
```
included as part of it to retrieve the page_url string. I believe this should work provided the project settings are updated accordingly.

ngpaladi:
Thank you all for considering adding this! I hope this will ultimately make it easier for people to get started.

> I've used a similar set of actions but instead had:
> 
> ```
> - name: Setup Pages
>         id: pages
>         uses: actions/configure-pages@v3
> ```
> 
> included as part of it to retrieve the page_url string. I believe this should work provided the project settings are updated accordingly.

I believe this step is only necessary if you need to force enable Pages without making the aforementioned repository settings changes, or if you need to grab the final URL as an input for running Doxygen. The former does not occur by default and likely would create permission issues.

atanisoft:
> if you need to grab the final URL as an input for running Doxygen.

In my case I'm using Jekyll to transform md into html files and it uses the url as an input, the file was generated as part of configuring pages the first time via repository settings. I think in this case it likely is fine to skip it but we should test it as part of enabling pages support and adjust if required.

PR #818, opened 2025-02-03
Fixes GpioInitializer to be always_inline and use tail recursion.

balazsracz:
Currently hw_init gets split into a bunch of functions that each add unnecessary some stack space.

### DIFF
--- a/src/utils/GpioInitializer.hxx
+++ b/src/utils/GpioInitializer.hxx
@@ -35,6 +35,8 @@
 #ifndef _UTILS_GPIOINITIALIZER_HXX_
 #define _UTILS_GPIOINITIALIZER_HXX_
 
+#include <tuple>
+
 /// Forward declaration to make the template matcher happy.
 template <class T> struct GpioInitHelper;
 


src/utils/GpioInitializer.hxx
```diff
@@ -61,22 +61,22 @@ template <typename Head, typename... Tail>
 struct GpioInitHelper<std::tuple<Head, Tail...>>
 {
     /// Middle step of the recursion for an individual pin.
-    static void hw_init()
+    static void __attribute__((always_inline)) hw_init()
     {
         Head::hw_init();
-        GpioInitHelper<std::tuple<Tail...>>::hw_init();
+        return GpioInitHelper<std::tuple<Tail...>>::hw_init();
```
bakerstu, line 67: I'm a bit confused why there is a return value in a void return method. Can you explain?

bakerstu (approved):

src/utils/GpioInitializer.hxx
atanisoft, line 67: should we add `#include <tuple>` in this file to fix the missing declaration?

balazsracz: done, good catch

balazsracz: You can return with a void function call from a void function. This is just a strong signal to the compiler that tail recursion optimization should be employed.

PR #820, opened 2025-02-03
Adds a test32 target

balazsracz:
- Adds targets/test32 which runs the unit tests under 32-bit compilation under linux.
- Fixes some problems with 32-bit compilation.
- Blacklists some tests in the 32-bit target only that I couldn't fix now.
- Adds a special compilation option that allows runningthe Stm32Gpio.cxxtest with including some header files from the Stm32Cube library (they do compile under linux).


bakerstu (approved):
I'm good with this. We have a number of fragile unit tests. I would like us to make a deliberate effort to clean these up. Some of them are squarely my responsibility, such as Broadcast Time Protocol tests, which should be all moved over to FakeClock.

src/openlcb/If.cxx
```diff
@@ -35,9 +35,14 @@
 #include "openlcb/If.hxx"
 #include "openlcb/Convert.hxx"
 
-/// Ensures that the largest bucket in the main buffer pool is exactly the size
-/// of a GenMessage.
-const unsigned LARGEST_BUFFERPOOL_BUCKET = sizeof(Buffer<openlcb::GenMessage>);
+static constexpr unsigned BUCKET_SIZE = std::min(sizeof(Buffer<openlcb::GenMessage>), 64u + sizeof(BufferBase));
```
atanisoft, line 38: should this have documentation on it?

balazsracz: Actually the documentation is a few lines below ("ensures that the largest bucket is ...") but the code is broken. It should be max() not min().

PR #821, opened 2025-02-06
Move bootloader_hal.h to the openlcb directory.

balazsracz:
Cleans up conditional compilation.

This would normally create breakages in applications that depend on openmrn, so we leave behind a symlink.


bakerstu (approved):

atanisoft:
@balazsracz can you update arduino/libify.sh for the file move?

PR #825, opened 2025-02-08
Fix crash in get space info 

balazsracz:
Directs all calls to the space registry through get_space() helper function.

This fixes a bug where the get_space_info call would access a memory space without first calling set_node() on it and causing a crash in an application.

### DIFF
--- a/src/openlcb/MemoryConfig.hxx
+++ b/src/openlcb/MemoryConfig.hxx
@@ -787,11 +787,11 @@ private:
             available_commands |= MemoryConfigDefs::AVAIL_SR;
         }
         // Figure out about ACDI spaces
-        MemorySpace* memspace = registry_.lookup(message()->data()->dst, 0xFC);
+        MemorySpace* memspace = get_space(0xFC);
         if (memspace) {
             available_commands |= MemoryConfigDefs::AVAIL_R0xFC;
         }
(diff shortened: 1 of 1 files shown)


src/openlcb/MemoryConfig.hxx
```diff
@@ -787,11 +787,11 @@ private:
             available_commands |= MemoryConfigDefs::AVAIL_SR;
         }
         // Figure out about ACDI spaces
-        MemorySpace* memspace = registry_.lookup(message()->data()->dst, 0xFC);
+        MemorySpace* memspace = get_space(0xFC);
```
atanisoft, line 790: Should we consider removing the magic numbers (0xFC, 0xFB, etc) and use the contants defind in MemoryConfigDiffs.hxx instead?

balazsracz:
Not in this place, since the code in the next line has
`MemoryConfigDefs::AVAIL_R0xFC`.

bakerstu (approved):

src/openlcb/MemoryConfig.hxx
balazsracz: This is generally a good idea, but not in this particular place, since the code in the next line has
`MemoryConfigDefs::AVAIL_R0xFC`.

PR #828, opened 2025-02-10
Simple HTTP server

RobertPHeller:
This is a simple HTTP server for OpenMRN nodes.

It defines three main classes:

HTTPD::Httpd -- the main server/listener class
HTTPD::HttpRequest -- a class to hold the request
HTTPD::HttpReply -- a class to hold the reply

The constructor of the HTTPD::Httpd takes two arguments, an Executor and a port number (defaults to port 80).
The application needs to add handlers using the add_uri method of the Httpd instance.  Passed in are the Uri, which can be a constant name (using the class HTTPD::Uri), a Glob match (using the class HTTPD::UriGlob, a braces match (using the HTTPD::UriBraces class), or a regular expression match (using the HTTPD::UriRegex class), a handler function taking three parameters: 

`void(const HTTPD::HttpRequest *request, HTTPD::HttpReply *reply,  void *userContext)`

and a void pointer to a context object.

When a incoming Uri matches, the handler function is called with the HTTPD::HttpRequest containing the request information and an empty HTTPD::HttpReply instance to fill with a reply.  When the reply is filled in, it can be then sent with a call to its member function SendReply().

See the test server in applications/test_http_server/main.cxx for a simple working server.


RobertPHeller:
I have one working node software using this library: https://github.com/RobertPHeller/WendellDepotMusuem that is working properly, in case anyone is looking for a use-case for this library.

balazsracz (commented):
I looked at this PR (without going too much in depth). Thank you for thinking about contributing this code to openmrn. I understand that this code seems to work well for you in your projects.
It seems to me that you would need to do a significant amount of additional work to bring this code to the point where it can be merged:
- The code is not following the style guide in terms of naming conventions (https://github.com/bakerstu/openmrn/blob/master/doc/style_guide.dox) There are a lot of differences, from variable names through method names to constant names, so I won't be listing them individually. Fixingthese is relatively mechanical, but tedious work.
- This code is entirely missing tests. Adding them would be a very significant amount of extra work.

I am not really sure that having a http server is a necessary addition to OpenMRN, as http servers are not specific to model railroading. Usually there are good quality http server libraries available, and till now pretty much all platforms that we've ported openmrn to had well-maintained http servers available as a platform feature (e.g. node.js, arduino, esp32, esp8266, TI simplelink platforms). In such a case it is a lot more productive to use the platform webserver than roll our own. If you are interested in an example for a linux host with openmrn-embedded webserver, take a look at applications/js_hub which includes both static page serving and dynamic (websocket-based) communication with a browser.

My recommendation altogether is to put this code into a separate git repository which you can include as a submodule to the projects where you need it, and symlink the respective source directories into your application directory. The code will then be built with the application target instead of with openmrn; this does not need any changes to the code. All of the openmrn infrastructure that this code is relying on (e.g. stateflows, select loops etc) are still available. I have a number of such libraries in my git repository (https://github.com/balazsracz/openmrn_cue/tree/master/cs) as examples of things which typically are niche use-cases, don't follow the guidelines of openmrn, or are not generic enough to be put into openmrn.

PR #830, opened 2025-03-31
Add GitHub actions tests

balazsracz:
Adds an actions workflow to run the tests.

Special conditions:
- adds a new method to run tests by building the test binariesi n parallel, but executing the actual tests in sequence. Invoke this with make tests-single at toplevel.
- Checks out the googletest git repository to fulfill test dependencies. 
- Installs some deb packages to make our linux code and tests compile

Two tests are blacklisted:
- SocketClient test, because MDNS class can not be instantiated on a linux host when avahi daemon is not running.
- IfTcp test because the remote connect test fails (something related to sockets in the github runner I think).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,6 +15,7 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+          path: openmrn
 
       - name: Checkout googletest
         uses: actions/checkout@v4
@@ -24,4 +25,6 @@ jobs:
           path: googletest
 
       - name: Run tests
-        run: make -j5 tests
+        run: |
+          cd openmrn
+          make -j5 tests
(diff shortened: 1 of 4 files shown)


atanisoft:
A couple comments:

- ModemRx seems to have a couple failures (timeouts) that should be tracked for fixing outside this PR likely.
- Can we get an overall summary of the test runs to display at the end of the run?
- Can we cause an action failure if there is a test failure?

balazsracz:
Right now this will fail if there is a test failure, because make will exit
with error code 1.

I'm not aware of flakiness in modemrx, feel free to file a bug though.
I think for test summary we'll wait for the cmake infrastructure, it does
not make sense to develop this in legacy.

atanisoft:
looking at it again, it looks like modemrx is outputing to stderr/stdout and is not an actual error. It just reads like one, perhaps through debug logging at this point.

In any case, I agree the summery can be deferred for certain. Good to know that it should error out on failure.

bakerstu (approved):

PR #834, opened 2025-05-16
[WIP] Add retry logic to the bootloader client.

bakerstu:


src/utils/GcTcpHub.cxx
```diff
@@ -72,4 +62,12 @@ GcTcpHub::GcTcpHub(CanHubFlow *can_hub, int port)
 GcTcpHub::~GcTcpHub()
 {
     tcpListener_.shutdown();
+    // There is probably a race condition here with regard to the GC Port being
+    // closed elsewhere. Since shutting down a hub is typically proceeding an
+    // attempt at a graceful shutdown anyways, we are letting it go.
+    for (auto it = clients_.begin(); it != clients_.end(); ++it)
```
balazsracz, line 68: We need to take a swap of the vector before calling these ::close calls. Otherwise we would have to do this with the atomic held to prevent callbacks from modifying the vector while we are iterating over it.
We can do a swap with a local vector under the atomic.

src/utils/GcTcpHub.hxx
```diff
@@ -63,27 +65,68 @@ public:
...
-    /// Error callback from the gridconnect socket. This is invoked when a
-    /// client disconnects.
-    void notify() override;
-
     /// @param can_hub Which CAN-hub should we attach the TCP gridconnect hub
     /// onto.
     CanHubFlow *canHub_;
     /// How many clients are connected right now.
-    unsigned numClients_ {0};
+    //unsigned numClients_ {0};
```
balazsracz, line 122: delete?

applications/bootloader_client/targets/linux.x86/main.cxx
```diff
@@ -41,6 +41,99 @@
...
+    }
+
+private:
+    void send_request()
+    {
+        Buffer<openlcb::BootloaderRequest> *b = fill_request();
+        bn_.reset(this);
+        b->set_done(&bn_);
+        maybe_checksum(&b->data()->data);
+        bootloader_client.send(b);
```
balazsracz, line 130: This is not the way I would retry the bootloading. The particular reason it is not super helpful is that it will retry the reboots as well.

Instead I would change the bootloader client's flow. The flow actually gets notified that the datagram client requesting the target reboot has timed out. This is visible in the flow on the return error code. If the datagram client timed out (which is different from OK and different from [remote endpoint rebooted]), then I would start a loop in the state flow where each iteration, run say once a second, sends out a verify node id global packet to the destination node ID. Keeps doing this until the remote target responds with a node ID verified.
Then continue the bootloading flow with the next stage (i guess that's PIP).
The important aspect is that the node ID verified message will update the remote alias cache, so we don't need to manually flush it. And if we don't get a reply (ever), it makes no sense to continue the bootloading flow either. 

We could even make this change just unconditionally in the bootloader flow. It will be working just fine in all cases.

balazsracz (changes requested):

applications/bootloader_client/targets/linux.x86/main.cxx
bakerstu: @balazsracz I would like to take a closer look at this with you. You are correct that there is a timeout. It is 3 seconds. However, it is not actually timing out for a full ~95 seconds. I've proven this with log statements. It makes no sense to me. Until we understand why the bootloader cannot continue to the next state for ~95 seconds, changing the logic won't help.

applications/bootloader_client/main.hxx
```diff
@@ -131,13 +132,15 @@ void usage(const char *e)
         "reply.\n");
     fprintf(stderr,
         "\n\t-D filename  writes the checksummed payload to the given file.\n");
+    fprintf(stderr,
+        "\n\t-R count  number of retries in the case of error, default 0.\n");
```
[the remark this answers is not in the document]
bakerstu: I backed out this change.

applications/bootloader_client/targets/linux.x86/main.cxx
bakerstu: We have a fix for this issue on the server side. There is no longer "retry" logic needed.

src/utils/GcTcpHub.cxx
bakerstu: I have a different fix for it.

src/utils/GcTcpHub.hxx
```diff
@@ -63,27 +65,68 @@ public:
     /// @return currently connected client count.
     unsigned get_num_clients()
     {
-        return numClients_;
+        return clients_.size();
     }
 
 private:
+    /// Helper object for managing connected clients.
+    class Notify : public Notifiable, private Atomic
```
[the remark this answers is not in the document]
bakerstu: Fixed.

src/utils/GcTcpHub.hxx
```diff
@@ -63,27 +65,68 @@ public:
...
+    {
+    public:
+        /// Constructor.
+        /// @param parent parent object pointer
+        /// @param fd file descriptor to register
+        Notify(GcTcpHub *parent, int fd)
+            : parent_(parent)
+            , fd_(fd)
+        {
+            AtomicHolder h(this);
```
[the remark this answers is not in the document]
bakerstu: Fixed.

src/utils/GcTcpHub.hxx
bakerstu: Fixed.

atanisoft (approved):

src/utils/GcTcpHub.hxx
```diff
@@ -63,27 +70,95 @@ public:
...
-    /// How many clients are connected right now.
-    unsigned numClients_ {0};
     /// Helper object representing the listening on the socket.
     SocketListener tcpListener_;
+
+    /// List of connected clients
+    std::vector<Client> clients_;
+
+    /// Provide access to private members from child object.
+    friend class Notify;
```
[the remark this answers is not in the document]
atanisoft: Interesting, I'm going to guess that it is implied local private access but would want @balazsracz to confirm. Likely it would be safe to adjust either way though.

src/utils/GcTcpHub.hxx
```diff
@@ -43,15 +46,19 @@ class ExecutorBase;
...
-class GcTcpHub : private Notifiable, private Atomic
+class GcTcpHub
 {
 public:
     /// Constructor.
     ///
     /// @param can_hub Which CAN-hub should we attach the TCP gridconnect hub
-    /// onto.
+    ///        onto.
     /// @param port TCp port number to listen on.
```
atanisoft, line 56: minor typo: `TCp` -> `TCP`

src/openlcb/Bootloader.hxx
```diff
@@ -1044,6 +1046,17 @@ bool bootloader_init() {
...
 bool bootloader_loop()
 {
+    if (state_.request_reinit_node)
+    {
+        if (state_.init_state == INITIALIZED)
+        {
+            // Back out of the initialized state so that it can try the node
+            // initialized transition again. Assume the alias is still good. If
+            // it is not, then there is already recovery logic for this.
+            state_.init_state = static_cast<InitState>(state_.init_state - 1);
```
balazsracz, line 1056: THis would be better written with the actual enum of the state you want to go into. The current code is "just back one" which likely does not mean "send node initialized message" anymore if the code changes.

src/freertos_drivers/esp32/Esp32WiFiManager.cxx
```diff
@@ -1503,9 +1503,22 @@ StateFlowBase::Action Esp32WiFiManager::WiFiStackFlow::reload()
...
+    {
+        // Alow a bit of time for lingering sockets to terminate.
+        return sleep_and_call(&timer_, MSEC_TO_NSEC(500), STATE(shutdown));
+    }
     return wait_and_call(STATE(reload));
 }
 
+StateFlowBase::Action Esp32WiFiManager::WiFiStackFlow::shutdown()
+{
+    LOG(INFO, "[WiFi] Shutting down.");
```
balazsracz, line 1517: How come we're not calling the gridconnect hub shutdown here? Where do we call that?

applications/bootloader_client/targets/linux.x86/main.cxx
```diff
@@ -41,6 +41,71 @@
...
+        HASSERT(fd_ >= 0);
+        create_gc_port_for_can_hub(&can_hub0, fd_, this);
+        wait_and_call(STATE(disconnected));
+    }
+
+    /// Disconnect occurred.
+    Action disconnected()
+    {
+        LOG(INFO, "disconnected()");
+        g_if_can.remote_aliases()->clear();
```
balazsracz, line 62: ```/// @todo clean up main to use a SimpleStack, then replace this with `SimpleStack::restart_stack()`.```

src/utils/GcTcpHub.hxx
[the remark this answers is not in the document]
balazsracz: WE should remove this line if the code builds without it. Currently this line is not useful because it has the wrong class name.

balazsracz (approved):

PR #836, opened 2025-05-22
Fixes object overlap problem in singleton test.

balazsracz:
Fixes issue in test where depending on compiler version and optimization level two objects could be overlapping due to different lifetimes and making the test fail.
Replaces the objects with guaranteed unique unitinitialized storage.


bakerstu (approved):
Good fix.

PR #838, opened 2025-05-31
Add Link Management to the Traction Modem

bakerstu:
- Adds an object for tracking link status.
- Adds an object for negotiating link bringup and monitoring.
- Adds [src/traction_modem/README.md](https://github.com/bakerstu/openmrn/blob/bakerstu-traction-modem-link-management/src/traction_modem/README.md)
- Does not yet support baud rate negotiation beyond 25Kbps.

Test Coverage:
image
image


src/traction_modem/Link.hxx
```diff
@@ -0,0 +1,404 @@
...
+
+namespace traction_modem
+{
+
+/// Interface for link start and up/down status.
+class LinkStatusInterface
+{
+protected:
+    /// Called when the link is started.
+    virtual void link_start()
```
balazsracz, line 49: since these are event callbacks, we could consider calling these as on_link_start(), on_link_up(), on_link_down().
This eliminates confusion that this finction call would cause the link to start or go up or go down.

src/traction_modem/Link.hxx
```diff
@@ -0,0 +1,404 @@
...
+    /// Bind an interface to the TX and RX interfaces to start transmitting and
+    /// receiving on.
+    /// @param fd interface to transmit and receive messages on
+    void start(int fd)
+    {
+        {
+            AtomicHolder h(this);
+            if (state_ != State::STOP)
+            {
+                // Already started.
```
balazsracz, line 98: is this a valid sequence of calls? We should at least display an error in the log.

src/traction_modem/Link.hxx
```diff
@@ -0,0 +1,404 @@
...
+            (*it)->link_down();
+        }
+    }
+
+    TxInterface *txIface_; ///< reference to the transmit interface interface
+    RxInterface *rxIface_; ///< reference to the receive interface
+    State state_; ///< link state
+
+    /// List of interfaces that have registered for link updates.
+    std::vector<LinkStatusInterface*> linkIfaces_;
```
balazsracz, line 178: Consider some better name for this variable, like linkCallbacks_.

In general, calling a variable Iface_ when that variable is not a network interface makes the code hard to read. Specifically, here "iface" seems to just mean an abstract class. However, the variable is never an abstract class. It is a pointer to an actual implementation of that abstract class and under no circumstances is the variable an interface.

src/traction_modem/Defs.hxx
```diff
@@ -56,6 +56,16 @@ struct Defs
...
     static constexpr uint16_t RESPONSE = 0x8000;
 
+    /// Normal (default) response timeout.
+    static constexpr long long RESP_TIMEOUT = SEC_TO_NSEC(3);
+
+    /// Shortened response timeout, typically used in the link establishment
+    /// phase.
+    static constexpr long long RESP_TIMEOUT_SHORT = SEC_TO_NSEC(1);
+
+    /// Time to wait for inactivity before proactively sending a ping.
```
balazsracz, line 66: // This must be comfortably below RESP_TIMEOUT otherwise we will send the ping too late and consider the link to be down.

src/traction_modem/Link.hxx
```diff
@@ -0,0 +1,404 @@
...
+            return RESTART;
+        }
+
+    private:
+        LinkManager *parent_; ///< parent object
+    };
+
+    Link *link_; ///< parent object
+    StateFlowTimer timer_; ///< timeout helper
+    PingTimer pingTimer_; ///< timer that periodically pings when link is idle
```
balazsracz, line 395: i think this pings even if the link is not idle

src/traction_modem/MemorySpace.cxxtest
```diff
@@ -55,8 +52,8 @@ static const MemorySpace::errorcode_t ERROR_AGAIN =
 
 TEST_F(MemorySpaceTest, Create)
 {
-    CvSpace cs(&g_service, &mTxFlow_, &mRxFlow_);
-    FirmwareSpace fs(&g_service, &mTxFlow_, &mRxFlow_);
+    CvSpace cs(&g_service, &link_);
```
balazsracz, line 55: we could have made the Link object be the Service. Then these constructors become simpler and the stateflows natively get the service pointer.

src/traction_modem/MemorySpace.hxx
```diff
@@ -96,9 +97,20 @@ public:
...
+        Defs::Payload p =
+            Defs::get_memw_payload(get_space_id(), destination, data, len);
+        if (link_->is_link_up())
+        {
+            // Link is up, we can send the write request immediately.
+            link_->get_tx_iface()->send_packet(p);
+        }
+        else
+        {
+            // Link is down, let's save the message in case the link comes up
```
balazsracz, line 110: what is the reason for adding this complexity here? Why do we need to handle memory config requests when there is no link?
Normally the link doesn't just come and go. It should come up at boot and that's it.

The reason for asking is that it feels like there is a lyering problem here. This is an application level protocol. This layer should not be worried about what is going on with the link. If it does, we have to copy-paste this code to N application level protocol handlers that we have. So if there is a legit feature here, that feature hsould not exist at this layer but at a lower layer so that we on't have to keep repeating this code (and these tests) over and over again.

balazsracz (approved):

PR #841, opened 2025-07-03
[WIP] New WiFi manager for ESP-IDF

bakerstu:
Done:
- Creates and abstract interface for WiFi management. Eventually this should be harmonized with the CC32xx.
- handles mDNS advertising and queries.
  - includes special handling to allow for masking mDNS advertising on the STA interface during a query.
- Updates WiFi configuration.
- Enforces a consistent (default) hostname between bootloader and application.
- Bootloader always uses well-known AP credentials.
- Bootloader mode "fast" connects only to the last known STA credentials.
- Workaround for "global" mDNS API.

Todo:
- Unit testing.
- Auto scan trigger when STA mode is up and IP assigned.

### DIFF
--- a/src/freertos_drivers/common/WiFiInterface.hxx
+++ b/src/freertos_drivers/common/WiFiInterface.hxx
@@ -46,25 +46,6 @@
 class WiFiInterface : public WiFiDefs, public Singleton<WiFiInterface>
 {
 public:
-    /// Network info, typically used in an access point scan.
-    struct NetworkEntry
-    {
-        /// Constructor.
-        /// @param ssid SSID of the access point
-        /// @param sec_type security type of the access point
-        /// @param rssi receive signal strength of the access point
-        NetworkEntry(const char *ssid, SecurityType sec_type, int rssi)
-            : ssid(ssid)
-            , secType(sec_type)
-            , rssi(rssi)
-        {
-        }
-
-        std::string ssid; ///< SSID of the access point
-        SecurityType secType; ///< security type of the access point
-        int rssi; ///< receive signal strength of the access point
-    };
-
     /// Initialize the WiFi.
     virtual void init()
     {
--- a/src/freertos_drivers/common/WiFiInterface.hxx
+++ b/src/freertos_drivers/common/WiFiInterface.hxx
@@ -1,0 +1,273 @@
+/** @copyright
+ * Copyright (c) 2025 Stuart Baker
+ * All rights reserved
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are  permitted provided that the following conditions are met:
+ *
+ *  - Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ *  - Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * @file WiFiInterface.hxx
+ *
+ * Common interface for WiFi operations.
+ *
+ * @author Stuart Baker
+ * @date 23 Jun 2025
+ */
+
+#ifndef _FREERTOS_DRIVERS_COMMON_WIFIINTERFACE_HXX_
+#define _FREERTOS_DRIVERS_COMMON_WIFIINTERFACE_HXX_
+
+#include <functional>
+#include <string>
+#include <vector>
+
+#include "utils/Singleton.hxx"
+#include "freertos_drivers/common/WifiDefs.hxx"
+
+/// Abstract interface for WiFi operations
+class WiFiInterface : public Singleton<WiFiInterface>
+{
+public:
+    /// Interface index by type.
+    enum Interface : uint8_t
+    {
+        IFACE_STA, ///< STA mode interface
+        IFACE_AP, ///< AP mode interface
+    };
+
+    /// Security types.
+    enum SecurityType : uint8_t
+    {
+        SEC_OPEN = 0, ///< open (no security)
+        SEC_WEP, ///< WEP security mode
+        SEC_WPA2, ///< WPA2 security mode
+    };
+
+    /// Network info, typically used in an access point scan.
+    struct NetworkEntry
+    {
+        /// Constructor.
+        /// @param ssid SSID of the access point
+        /// @param sec_type security type of the access point
+        /// @param rssi receive signal strength of the access point
+        NetworkEntry(const char *ssid, SecurityType sec_type, int rssi)
+            : ssid(ssid)
+            , secType(sec_type)
+            , rssi(rssi)
+        {
+        }
+
+        std::string ssid; ///< SSID of the access point
+        SecurityType secType; ///< security type of the access point
+        int rssi; ///< receive signal strength of the access point
+    };
+
+    /// Initialize the WiFi.
+    virtual void init()
+    {
+    }
+
+    /// Start the WiFi.
+    /// @param role device role
+    virtual void start(WlanRole role = WlanRole::DEFAULT_ROLE) = 0;
+
+    /// Stop the WiFi.
+    virtual void stop() = 0;
+
+    /// Get the started state of the WiFi.
+    /// @return true if started, else false
+    virtual bool is_started()
+    {
+        return started_;
+    }
+
+    /// Connect to access point.
+    /// @param ssid access point SSID
+    /// @param pass access point password
+    /// @param sec_type access point security type
+    /// @return WlanConnectResult::CONNECT_OK upon connect, else error code
+    virtual WlanConnectResult connect(
+        const char *ssid, const char *pass, SecurityType sec_type) = 0;
+
+    /// Disconnect from the current AP.
+    virtual void disconnect() = 0;
+
+    /// Gets the connected state of the STA interface
+    /// @return true if STA connected/associated to an AP, else false
+    virtual bool is_connected()
+    {
+        return connected_;
+    }
+
+    /// Setup access point role credentials. May require reboot to take effect.
+    /// @param ssid access point SSID
+    /// @param pass access point password
+    /// @param sec_type access point security type
+    virtual void setup_ap(
+        const char *ssid, const char *pass, SecurityType sec_type) = 0;
+
+    /// Retrieve current access point config.
+    /// @param ssid access point SSID
+    /// @param sec_type access point security type
+    virtual void get_ap_config(std::string *ssid, SecurityType *sec_type) = 0;
+
+    /// Retrieves the number of clients connected to the WiFi in access point
+    /// mode.
+    /// @return number of connected stations, else -1 if not in AP mode.
+    virtual int get_ap_sta_count() = 0;
+
+    /// Retrieves if ready to start establishing outgoing connections.
+    /// @return true if ready, else false
+    virtual bool ready()
+    {
+        return (ipAcquiredSta_ || ipAcquiredAp_);
+    }
+
+    /// Get the WiFi role.
+    /// @return current WiFi role
+    virtual WlanRole role() = 0;
+
+    /// Change the default role. This will be used in the next start() if the
+    /// DEFAULT_ROLE is specified. The new setting takes effect when the
+    /// device is restarted (either via reboot or stop + start)
+    /// @param new_role new role, must not be UNKOWN or DEFAULT_ROLE
+    virtual void set_role(WlanRole new_role) = 0;
+
+    /// Add a saved WiFi access point profile.
+    /// @param ssid access point SSID
+    /// @param pass access point password
+    /// @param sec_type access point security type
+    /// @param priority priority when more than one profile is saved, 0 =
+    ///        lowest priority, may be unused
+    /// @return resulting index in the list of profiles, else -1 on error
+    virtual int profile_add(const char *ssid, const char *pass,
+        SecurityType sec_type, uint8_t priority) = 0;
+
+    /// Delete a saved WiFi access point profile.
+    /// @param ssid access point SSID
+    /// @return profile index deleted, else -1 if profile not found
+    virtual int profile_del(const char *ssid) = 0;
+
+    /// Delete a saved WiFi access point profile.
+    /// @param index index of profile to delete, 0xFF removes all
+    /// @return 0 if successful, else -1 if profile not found, may be unused
+    virtual int profile_del(uint8_t index) = 0;
+
+    /// Get a saved WiFi access point profile by index.
+    /// @param index index within saved profilelist to get
+    /// @param ssid 33 byte array that will return the SSID of the index
+    /// @param sec_type will return the security type of the index
+    /// @param priority will return the priority of the index
+    /// @return 0 upon success, -1 on error (index does not exist)
+    virtual int profile_get(
+        int index, char ssid[], SecurityType *sec_type, uint8_t *priority) = 0;
+
+    /// Get a list of available networks. This is based on a prior scan. Use
+    /// rescan() and wait for the scan to complete to refresh the list.
+    /// @param entries returns a list of available network entries
+    /// @param count max size of the entry list to return.
+    /// @return number of valid network entries in the list, same as
+    ///         entries->size()
+    virtual int network_list_get(
+        std::vector<NetworkEntry> *entries, size_t count) = 0;
+
+    /// Get the indexed network entry from the last of scan results. Use
+    /// rescan() and wait for the scan to complete to refresh the results.
+    /// @param entry location to fill in the network entry
+    /// @param index index in the network entry list to get
+    /// @return 0 on success, else -1 if index is beyond the list of entries.
+    virtual int network_get(NetworkEntry *entry, unsigned index) = 0;
+
+    /// Initiate scanning of available networks.
+    virtual void scan() = 0;
+
+    /// Get the RSSI of the AP in STA mode.
+    /// @return signal strength, 0 when not connected to an access point
+    virtual int rssi() = 0;
+
+    /// Get the network hostname for the device.
+    /// @return hostname
+    virtual std::string get_hostname() = 0;
+
+    /// Set the callback for when an IP address is acquired.
+    virtual void set_ip_acquired_callback(
+        std::function<void(Interface, bool)> callback)
+    {
+        ipAcquiredCallback_ = callback;
+    }
+
+    /// Set the callback for when a scan is finished.
+    virtual void set_scan_finished_callback(std::function<void()> callback)
+    {
+        scanFinishedCallback_ = callback;
+    }
+
+protected:
+    /// Constructor.
+    WiFiInterface()
+        : started_(false)
+        , connected_(false)
+        , ipAcquiredSta_(false)
+        , ipAcquiredAp_(false)
+        , ipLeased_(false)
+        , securityFailure_(false)
+    {
+    }
+
+    /// Deliver IP acquired callback, if registered.
+    /// @param iface interface index
+    /// @param acquired true if acquired, false if lost
+    void ip_acquired(Interface iface, bool acquired)
+    {
+        if (ipAcquiredCallback_)
+        {
+            ipAcquiredCallback_(iface, acquired);
+        }
+    }
+
+    /// Deliver scan finished callback, if registered.
+    void scan_finished()
+    {
+        if (scanFinishedCallback_)
+        {
+            scanFinishedCallback_();
+        }
+    }
+
+    /// Callback for when IP is acquired.
+    /// @param iface interface index
+    /// @param acquired true if acquired, false if lost
+    std::function<void(Interface iface, bool acquired)> ipAcquiredCallback_;
+
+    /// Callback for when a scan is finished.
+    std::function<void()> scanFinishedCallback_;
+
+    // Note: Because these are bitmasks, care must be taken in the derived
+    //       class that there are no mutual exclusion issues on write.
+    bool started_          : 1; ///< WiFi started
+    bool connected_        : 1; ///< STA mode connection to AP connected state
+    bool ipAcquiredSta_    : 1; ///< IP address acquired STA mode
+    bool ipAcquiredAp_     : 1; ///< IP address acquired AP mode
+    bool ipLeased_         : 1; ///< IP address leased to a client (AP mode)
+    bool securityFailure_  : 1; ///< Disconnected due to wrong password
+};
+
+#endif // _FREERTOS_DRIVERS_COMMON_WIFIINTERFACE_HXX_
--- a/src/freertos_drivers/common/WiFiInterface.hxx
+++ b/src/freertos_drivers/common/WiFiInterface.hxx
@@ -221,11 +221,7 @@ public:
     /// Get a list of available networks. This is based on a prior scan. Use
     /// scan() and wait for the scan to complete to refresh the list.
     /// @param entries returns a list of available network entries
-    /// @param count max size of the entry list to return.
-    /// @return number of valid network entries in the list, same as
-    ///         entries->size()
-    virtual int network_list_get(
-        std::vector<NetworkEntry> *entries, size_t count) = 0;
+    virtual void network_list_get(std::vector<NetworkEntry> *entries) = 0;
 
     /// Get the indexed network entry from the last of scan results. Use
     /// scan() and wait for the scan to complete to refresh the results.
--- a/src/freertos_drivers/common/WiFiInterface.hxx
+++ b/src/freertos_drivers/common/WiFiInterface.hxx
@@ -108,6 +108,31 @@ public:
         return started_;
     }
 
+    /// Get the default AP password.
+    /// @return default AP password
+    virtual const char *default_ap_password() = 0;
+
+    /// Get the default STA password.
+    /// @return default STA password
+    virtual const char *default_sta_password() = 0;
+
+    /// Get the default AP SSID.
+    /// @return default AP SSID
+    virtual const char *default_ap_ssid() = 0;
+
+    /// Get the default STA SSID.
+    /// @return default STA SSID
+    virtual const char *default_sta_ssid() = 0;
+
+    /// Get the maximum number of STA client connections in AP mode. Be careful,
+    // the max supported by hardware is device dependent.
+    /// @return maximum number of STA client connections in AP mode
+    virtual uint8_t max_ap_client_connections() = 0;
+
+    /// Get the maximum number of stored STA profiles.
+    /// @return maximum number of stored STA profiles
+    virtual uint8_t max_sta_profiles() = 0;
+
     /// Connect to access point. This is a non-blocking call. The results will
     /// be delivered by callback registered with set_wlan_connect_callback().
     /// @param ssid access point SSID
--- a/src/freertos_drivers/common/WiFiInterface.hxx
+++ b/src/freertos_drivers/common/WiFiInterface.hxx
@@ -109,19 +109,19 @@ public:
     }
 
     /// Get the default AP password.
-    /// @return default AP password
+    /// @return default AP password, should point to persistent memory
     virtual const char *default_ap_password() = 0;
 
     /// Get the default STA password.
-    /// @return default STA password
+    /// @return default STA password, should point to persistent memory
     virtual const char *default_sta_password() = 0;
 
     /// Get the default AP SSID.
-    /// @return default AP SSID
+    /// @return default AP SSID, should point to persistent memory
     virtual const char *default_ap_ssid() = 0;
 
     /// Get the default STA SSID.
-    /// @return default STA SSID
+    /// @return default STA SSID, should point to persistent memory
     virtual const char *default_sta_ssid() = 0;
 
     /// Get the maximum number of STA client connections in AP mode. Be careful,
@@ -135,6 +135,8 @@ public:
 
     /// Connect to access point. This is a non-blocking call. The results will
     /// be delivered by callback registered with set_wlan_connect_callback().
+    /// The AP credentials are not saved as a connection profile, but they may
+    /// be saved in non-volatile storage for use on the next connection attempt.
     /// @param ssid access point SSID
     /// @param pass access point password
     /// @param sec_type access point security type
@@ -217,7 +219,7 @@ public:
         int index, char ssid[], SecurityType *sec_type, uint8_t *priority) = 0;
 
     /// Get a list of available networks. This is based on a prior scan. Use
-    /// rescan() and wait for the scan to complete to refresh the list.
+    /// scan() and wait for the scan to complete to refresh the list.
     /// @param entries returns a list of available network entries
     /// @param count max size of the entry list to return.
     /// @return number of valid network entries in the list, same as
@@ -226,17 +228,20 @@ public:
         std::vector<NetworkEntry> *entries, size_t count) = 0;
 
     /// Get the indexed network entry from the last of scan results. Use
-    /// rescan() and wait for the scan to complete to refresh the results.
+    /// scan() and wait for the scan to complete to refresh the results.
     /// @param entry location to fill in the network entry
     /// @param index index in the network entry list to get
     /// @return 0 on success, else -1 if index is beyond the list of entries.
     virtual int network_get(NetworkEntry *entry, unsigned index) = 0;
 
-    /// Initiate scanning of available networks.
+    /// Initiate scanning of available networks. Use 
+    /// set_scan_finished_callback() to register a callback upon completion of
+    /// the scan.
     virtual void scan() = 0;
 
     /// Get the RSSI of the AP in STA mode.
-    /// @return signal strength, 0 when not connected to an access point
+    /// @return signal strength, 0 when not connected to an access point, else
+    ///         should be a negative number.
     virtual int rssi() = 0;
 
     /// Get the network hostname for the device.
@@ -259,13 +264,13 @@ public:
     /// @param connected true if connected, else false if disconnected
     /// @param result result code for the connection event
     /// @param ssid SSID of the connection event
-    virtual void set_wlan_connect_callback(std::function<void(
-        bool connected, ConnectionResult result, std::string &ssid)> callback)
+    virtual void set_wlan_connect_callback(std::function<void(bool connected,
+        ConnectionResult result, const std::string &ssid)> callback)
     {
         wlanConnectedCallback_ = std::move(callback);
     }
 
-    /// Set the callback for when a scan is finished.
+    /// Set the callback for when a WiFi access point scan is finished.
     virtual void set_scan_finished_callback(std::function<void()> callback)
     {
         scanFinishedCallback_ = std::move(callback);
@@ -298,7 +303,7 @@ protected:
     /// @param result result code for the connection event
     /// @param ssid SSID of the connection event
     void wlan_connected(
-        bool connected, ConnectionResult result, std::string &ssid)
+        bool connected, ConnectionResult result, const std::string &ssid)
     {
         if (wlanConnectedCallback_)
         {
@@ -325,7 +330,7 @@ protected:
     /// @param result result code for the connection event
     /// @param ssid SSID of the connection event
     std::function<void(
-        bool connected, ConnectionResult result, std::string &ssid)>
+        bool connected, ConnectionResult result, const std::string &ssid)>
             wlanConnectedCallback_;
 
     /// Callback for when a scan is finished.
@@ -333,11 +338,13 @@ protected:
 
     // Note: Because these are bitmasks, care must be taken in the derived
     //       class that there are no mutual exclusion issues on write.
-    bool started_          : 1; ///< WiFi started
-    bool connected_        : 1; ///< STA mode connection to AP connected state
-    bool ipAcquiredSta_    : 1; ///< IP address acquired STA mode
-    bool ipAcquiredAp_     : 1; ///< IP address acquired AP mode
-    bool ipLeased_         : 1; ///< IP address leased to a client (AP mode)
+    //       Implementations are expected to continuously update these bits as
+    //       state transitions occur.
+    bool started_       : 1; ///< WiFi started
+    bool connected_     : 1; ///< STA mode connection to AP connected state
+    bool ipAcquiredSta_ : 1; ///< IP address acquired STA mode
+    bool ipAcquiredAp_  : 1; ///< IP address acquired AP mode
+    bool ipLeased_      : 1; ///< IP address leased to a client (AP mode)
 };
 
 #endif // _FREERTOS_DRIVERS_COMMON_WIFIINTERFACE_HXX_
--- a/src/freertos_drivers/common/WiFiInterface.hxx
+++ b/src/freertos_drivers/common/WiFiInterface.hxx
@@ -43,33 +43,9 @@
 #include "freertos_drivers/common/WifiDefs.hxx"
 
 /// Abstract interface for WiFi operations
-class WiFiInterface : public Singleton<WiFiInterface>
+class WiFiInterface : public WiFiDefs, public Singleton<WiFiInterface>
 {
 public:
-    /// Interface index by type.
-    enum Interface : uint8_t
-    {
-        IFACE_STA, ///< STA mode interface
-        IFACE_AP, ///< AP mode interface
-    };
-
-    /// Security types.
-    enum SecurityType : uint8_t
-    {
-        SEC_OPEN = 0, ///< open (no security)
-        SEC_WEP, ///< WEP security mode
-        SEC_WPA2, ///< WPA2 security mode
-    };
-
-    /// Result code for connections and disconnections.
-    enum ConnectionResult : uint8_t
-    {
-        CONNECT_OK = 0, ///< connection succeeded
-        AUTHENTICATION_FAILED, ///< authentication failure
-        ASSOCIATION_FAILED, ///< association failure
-        CONNECT_UNKNOWN, ///< unknown result
-    };
-
     /// Network info, typically used in an access point scan.
     struct NetworkEntry
     {
--- a/src/freertos_drivers/common/WiFiInterface.hxx
+++ b/src/freertos_drivers/common/WiFiInterface.hxx
@@ -61,6 +61,15 @@ public:
         SEC_WPA2, ///< WPA2 security mode
     };
 
+    /// Result code for connections and disconnections.
+    enum ConnectionResult : uint8_t
+    {
+        CONNECT_OK = 0, ///< connection succeeded
+        AUTHENTICATION_FAILED, ///< authentication failure
+        ASSOCIATION_FAILED, ///< association failure
+        CONNECT_UNKNOWN, ///< unknown result
+    };
+
     /// Network info, typically used in an access point scan.
     struct NetworkEntry
     {
@@ -99,15 +108,17 @@ public:
         return started_;
     }
 
-    /// Connect to access point.
+    /// Connect to access point. This is a non-blocking call. The results will
+    /// be delivered by callback registered with set_wlan_connect_callback().
     /// @param ssid access point SSID
     /// @param pass access point password
     /// @param sec_type access point security type
-    /// @return WlanConnectResult::CONNECT_OK upon connect, else error code
-    virtual WlanConnectResult connect(
+    virtual void connect(
         const char *ssid, const char *pass, SecurityType sec_type) = 0;
 
-    /// Disconnect from the current AP.
+    /// Disconnect from the current AP. This is a non-blocking call. The
+    /// result will be delivered by callback registered with
+    /// set_wlan_connect_callback().
     virtual void disconnect() = 0;
 
     /// Gets the connected state of the STA interface
@@ -208,16 +219,31 @@ public:
     virtual std::string get_hostname() = 0;
 
     /// Set the callback for when an IP address is acquired.
+    /// @param iface interface index
+    /// @param acquired true if acquired, false if lost
     virtual void set_ip_acquired_callback(
-        std::function<void(Interface, bool)> callback)
+        std::function<void(Interface iface, bool acquired)> callback)
     {
-        ipAcquiredCallback_ = callback;
+        ipAcquiredCallback_ = std::move(callback);
+    }
+
+    /// Set the callback for when a wlan connect or disconnect occurs.
+    /// Back-to-back deliveries of "disconnected" are possible without first
+    /// transitioning to a connected state. This can occur when a connection
+    /// "attempt" has failed to succeed.
+    /// @param connected true if connected, else false if disconnected
+    /// @param result result code for the connection event
+    /// @param ssid SSID of the connection event
+    virtual void set_wlan_connect_callback(std::function<void(
+        bool connected, ConnectionResult result, std::string &ssid)> callback)
+    {
+        wlanConnectedCallback_ = std::move(callback);
     }
 
     /// Set the callback for when a scan is finished.
     virtual void set_scan_finished_callback(std::function<void()> callback)
     {
-        scanFinishedCallback_ = callback;
+        scanFinishedCallback_ = std::move(callback);
     }
 
 protected:
@@ -228,7 +254,6 @@ protected:
         , ipAcquiredSta_(false)
         , ipAcquiredAp_(false)
         , ipLeased_(false)
-        , securityFailure_(false)
     {
     }
 
@@ -240,6 +265,19 @@ protected:
         if (ipAcquiredCallback_)
         {
             ipAcquiredCallback_(iface, acquired);
+        }
+    }
+
+    /// Deliver WLAN connected callback, if registered.
+    /// @param connected true if connected, else false if disconnected
+    /// @param result result code for the connection event
+    /// @param ssid SSID of the connection event
+    void wlan_connected(
+        bool connected, ConnectionResult result, std::string &ssid)
+    {
+        if (wlanConnectedCallback_)
+        {
+            wlanConnectedCallback_(connected, result, ssid);
         }
     }
 
@@ -257,6 +295,14 @@ protected:
     /// @param acquired true if acquired, false if lost
     std::function<void(Interface iface, bool acquired)> ipAcquiredCallback_;
 
+    /// Callback for when WLAN connect or disconnect on STA occurs.
+    /// @param connected true if connected, else false if disconnected
+    /// @param result result code for the connection event
+    /// @param ssid SSID of the connection event
+    std::function<void(
+        bool connected, ConnectionResult result, std::string &ssid)>
+            wlanConnectedCallback_;
+
     /// Callback for when a scan is finished.
     std::function<void()> scanFinishedCallback_;
 
@@ -267,7 +313,6 @@ protected:
     bool ipAcquiredSta_    : 1; ///< IP address acquired STA mode
     bool ipAcquiredAp_     : 1; ///< IP address acquired AP mode
     bool ipLeased_         : 1; ///< IP address leased to a client (AP mode)
-    bool securityFailure_  : 1; ///< Disconnected due to wrong password
 };
 
 #endif // _FREERTOS_DRIVERS_COMMON_WIFIINTERFACE_HXX_
--- a/src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
+++ b/src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
@@ -297,7 +297,7 @@ void EspIdfWiFiBase::mdns_service_remove(const char *service)
     }
     for (auto it = mdnsServices_.begin(); it != mdnsServices_.end(); ++it)
     {
-        if ((*it).name_ == name && (*it).proto_ == proto)
+        if (it->name_ == name && it->proto_ == proto)
         {
             mdnsServices_.erase(it);
             return;
@@ -316,7 +316,7 @@ int EspIdfWiFiBase::mdns_lookup(
     OSMutexLock locker(&lock_);
     for (auto it = mdnsClientCache_.begin(); it != mdnsClientCache_.end(); ++it)
     {
-        if ((*it).service_ != service)
+        if (it->service_ != service)
         {
             // Not a match.
             continue;
@@ -324,7 +324,7 @@ int EspIdfWiFiBase::mdns_lookup(
 
         // Found a potential match. Check for an address list
         looking = true;
-        size_t addr_cnt = (*it).addr_.size();
+        size_t addr_cnt = it->addr_.size();
         if (addr_cnt == 0)
         {
             // No resolved addresses to report.
@@ -338,7 +338,7 @@ int EspIdfWiFiBase::mdns_lookup(
         struct addrinfo *current = *addr;
         struct addrinfo *last = nullptr;
 
-        for (auto lit = (*it).addr_.begin(); lit != (*it).addr_.end(); ++lit)
+        for (auto lit = it->addr_.begin(); lit != it->addr_.end(); ++lit)
         {
 // The caller is expected to free the struct addrinfo using the method
 // freeaddrinfo. In the lwIP implementation, they use a special buffer pool
@@ -405,7 +405,7 @@ void EspIdfWiFiBase::mdns_scan(const char *service)
     OSMutexLock locker(&lock_);
     for (auto it = mdnsClientCache_.begin(); it != mdnsClientCache_.end(); ++it)
     {
-        if ((*it).service_ == service)
+        if (it->service_ == service)
         {
             match = true;
         }
@@ -427,12 +427,30 @@ void EspIdfWiFiBase::factory_default()
     // This mutex lock will hold off a configuration sync until all the updates
     // are made.
     OSMutexLock locker(&lock_);
-    setup_ap(default_ap_ssid(), default_ap_password(), SEC_WPA2);
+
+    // AP setup.
+    const char *pass = default_ap_password();
+    SecurityType sec = SEC_WPA2;
+    if (!pass || pass[0] == '\0')
+    {
+        sec = SEC_OPEN;
+    }
+    setup_ap(default_ap_ssid(), pass, sec);
+
+    // Delete all STA profiles.
     for (unsigned i = 0; i < max_sta_profiles(); ++i)
     {
         profile_del(i);
     }
-    profile_add(default_sta_ssid(), default_sta_password(), SEC_WPA2, 0);
+
+    // Add single default STA profile.
+    pass = default_sta_password();
+    sec = SEC_WPA2;
+    if (!pass || pass[0] == '\0')
+    {
+        sec = SEC_OPEN;
+    }
+    profile_add(default_sta_ssid(), pass, sec, 0);
 }
 
 //
@@ -472,8 +490,8 @@ StateFlowBase::Action EspIdfWiFiBase::mdns_query()
     {
         std::string name;
         std::string proto;
-        mdns_split((*it).service_.c_str(), &name, &proto);
-        (*it).reset((void*)mdns_query_async_new(
+        mdns_split(it->service_.c_str(), &name, &proto);
+        it->reset((void*)mdns_query_async_new(
             nullptr, name.c_str(), proto.c_str(), MDNS_TYPE_PTR,
             MDNS_QUERY_ASYNC_NEW_TIMEOUT_MSEC, 3, nullptr));
         LOG(VERBOSE, "wifi: mDNS new search query: %s.%s",
@@ -494,7 +512,7 @@ StateFlowBase::Action EspIdfWiFiBase::mdns_check()
     // Loop through all of the service clients.
     for (auto it = mdnsClientCache_.begin(); it != mdnsClientCache_.end(); ++it)
     {
-        if ((*it).get() == nullptr)
+        if (it->get() == nullptr)
         {
             // No active search for this cache entry.
             LOG(VERBOSE, "wifi: mDNS no active search for this cache entry.");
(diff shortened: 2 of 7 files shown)


src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+/// @param ifap pointer to return the list of interface addresses to
+/// @return 0 upon success, else -1 with errno set to indicate the error
+int getifaddrs(struct ifaddrs **ifap)
+{
+    esp_netif_t *netif_ap = esp_netif_get_handle_from_ifkey("WIFI_AP_DEF");
+    esp_netif_t *netif_sta = esp_netif_get_handle_from_ifkey("WIFI_STA_DEF");
+    struct ifaddrs *if_addrs_ap = getifaddrs_helper(netif_ap, "ap");
+    struct ifaddrs *if_addrs_sta = getifaddrs_helper(netif_sta, "sta");
+
+    if (netif_ap)
```
bakerstu, line 144: From @balazsracz: `if (if_addrs_ap)` would be safer

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+/// Get interface addresses
+/// @param ifap pointer to return the list of interface addresses to
+/// @return 0 upon success, else -1 with errno set to indicate the error
+int getifaddrs(struct ifaddrs **ifap)
+{
+    esp_netif_t *netif_ap = esp_netif_get_handle_from_ifkey("WIFI_AP_DEF");
+    esp_netif_t *netif_sta = esp_netif_get_handle_from_ifkey("WIFI_STA_DEF");
+    struct ifaddrs *if_addrs_ap = getifaddrs_helper(netif_ap, "ap");
+    struct ifaddrs *if_addrs_sta = getifaddrs_helper(netif_sta, "sta");
+
```
bakerstu, line 143: From @balazsracz: HASSERT(ifap)

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,273 @@
...
+    virtual void stop() = 0;
+
+    /// Get the started state of the WiFi.
+    /// @return true if started, else false
+    virtual bool is_started()
+    {
+        return started_;
+    }
+
+    /// Connect to access point.
```
bakerstu, line 102: From @balazsracz: Please document whether this is a blocking call or an asynchronous call.
On the cc32xx this call seems to be blocking, i.e., it could take several seconds then return an error.
It's unclear to me on the esp32, i think the implementation in this PR is asynchronous (i.e., non-blocking) and the return value seems to be meaningless.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+/// }
+/// @endcode
+///
+/// In order to support non-volatile configuration, use the specialized version
+/// of this object EspIdfWiFiWithConfigUpdateListener instead. In this version,
+/// whenever there is an update to the user configuration, an update flow is
+/// triggered which will commit the user config to both NVS and a companion
+/// MemorySpace.
+///
+class EspIdfWiFi : public WiFiInterface, public StateFlowBase, protected OSMutex
```
bakerstu, line 87: From @balazsracz: instead of inheriting from OSMutex it would be cleaner to have an OSMutex lock_; Then at the declaration of lock_ document in a comment which class variables lock_ is protecting.

eventually at some point we could consider adopting thread safety annotations but this is a hill to climb later (https://clang.llvm.org/docs/ThreadSafetyAnalysis.html).

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+{
+public:
+    /// Default wifi password.
+    static constexpr char DEFAULT_PASSWORD[] = "123456789";
+
+    /// Constructor.
+    /// @param service service to bind this object to.
+    /// @param hostname hostname to publish over the network
```
bakerstu, line 95: From @balazsracz: please document the expected lifetime of this pointer

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+    /// @param can_hub CAN hub to use
```
bakerstu, line 96: From @balazsracz: gone?

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+/// MemorySpace.
+///
+class EspIdfWiFi : public WiFiInterface, public StateFlowBase, protected OSMutex
+{
+public:
+    /// Default wifi password.
+    static constexpr char DEFAULT_PASSWORD[] = "123456789";
+
+    /// Constructor.
+    /// @param service service to bind this object to.
```
bakerstu, line 94: From @balazsracz: Please describe whether we will block this executor or not (e.g. whether it is fine to provide the shared global executor for the stack here).

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+    /// @param hostname hostname to publish over the network
+    /// @param can_hub CAN hub to use
+    EspIdfWiFi(Service *service, const char * hostname)
+        : StateFlowBase(service)
+        , OSMutex(true) // Recursive.
```
bakerstu, line 99: From @balazsracz: is there a reason we need a recursive mutex? It is usually cleaner code when we have unique responsibility of who locks where.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+    {
+    }
+
+    StateFlowTimer timer_; ///< sleep timer helper object
+    esp_netif_t *apIface_; ///< access point network interface
+    esp_netif_t *staIface_; ///< station network interface
+    std::vector<MDNSService> mdnsServices_; ///< registered mDNS services
+    std::vector<NetworkEntry> scanResults_; ///< AP scan results
+    std::string staConnectPass_; ///< last station connect attempt password
+    const char *hostname_; ///< published hostname
```
bakerstu, line 609: From @balazsracz: we should probably make this a const std::string instead

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+            struct sockaddr addr_; ///< address
+            struct sockaddr_in addrIn_; ///< IPv4 address
+#if defined(ESP_IDF_WIFI_IPV6)
+            struct sockaddr_in6 addrIn6_; ///< IPv6 address
+#endif
+        };
+    };
+
+    /// Metadata for a subscribed mDNS client.
+    struct MDNSCacheItem
```
bakerstu, line 414: From @balazsracz: // struct MdnsCacheItem

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+            , port_(port)
+        {
+        }
+        std::string name_; ///< service name
+        std::string proto_; ///< service protocol
+        uint16_t port_; ///< service port
+    };
+
+    /// Metadata for an subscribed mDNS cashed address.
+    struct MDNSCacheAddr
```
bakerstu, line 399: From @balazsracz: // struct MdnsCacheAddr

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+    void factory_default();
+
+    nvs_handle_t cfg_; ///< handle to the nvs config for wifi configuration
+    WiFiConfigNVS userCfg_; ///< cached copy of the wifi user config
+    ConfigPrivate privCfg_; ///< private WiFi configuration
+    bool initialized_; ///< initialization complete
+
+private:
+    /// Metadata for a registered mDNS service.
+    struct MDNSService
```
bakerstu, line 381: From @balazsracz: // struct MdnsService

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+
+    /// Maximum length of a stored password not including '/0' termination.
+    static constexpr size_t MAX_PASS_SIZE =
+        sizeof(WiFiConfigCredentialsNVS::pass_) - 1;
+    static_assert(MAX_PASS_SIZE == 63, "Invalid maximum password length.");
+
+    /// Reset configuration to factory defaults.
+    void factory_default();
+
+    nvs_handle_t cfg_; ///< handle to the nvs config for wifi configuration
```
bakerstu, line 374: From @balazsracz: it would be preferable to have all class variables next to each other (at the very bottom). Right now they are split between far away places in the source code.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+}
+
+//
+// EspIdfWiFi::mdns_scan()
+//
+void EspIdfWiFi::mdns_scan(const char *service)
+{
+    bool match = false;
+    OSMutexLock locker(this);
+    for (auto it = mdnsClientCache_.begin(); it != mdnsClientCache_.end(); ++it)
```
atanisoft, line 457: we should consider STL `find_if` usage here which can return a pointer to the entry if found.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+    {
+        // Start looking for a new service.
+        mdnsClientCache_.emplace_back(std::move(std::string(service)));
+    }
+    if (!mdnsClientStarted_)
+    {
+        // Start the client if not already started.
+        mdnsClientStarted_ = true;
+        notify();
+        LOG(INFO, "wifi: mdns_scan() start.");
```
atanisoft, line 474: consider log level adjustment here and below

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+//
+// EspIdfWiFi::mdns_query()
+//
+StateFlowBase::Action EspIdfWiFi::mdns_query()
+{
+    OSMutexLock lock(this);
+    mdnsClientTrigRefresh_ = false;
+    for (auto it = mdnsClientCache_.begin(); it != mdnsClientCache_.end(); ++it)
+    {
+        std::string name;
```
atanisoft, line 542: in cpp files we can likely remove the `std::` prefix as it should already be used by other libs, we can also consider adding a `using std;` at the top if needed

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+                addr = addr->next)
+            {
+                // For some reason, the mdns_result_t::ip_protocol cannot
+                // be trusted. The address results list can contain both
+                // IPv4 and IPv6 addresses.
+                if (addr->addr.type == ESP_IPADDR_TYPE_V4)
+                {
+                    cache_addr.family_ = AF_INET;
+                }
+#if defined(ESP_IDF_WIFI_IPV6)
```
atanisoft, line 607: `CONFIG_LWIP_IPV6` might be a good replacement here, with possible fallback to force IPv4

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+                    {
+                        duplicate &= (*ait).addrIn_.sin_addr.s_addr ==
+                            addr->addr.u_addr.ip4.addr;
+                    }
+#if defined(ESP_IDF_WIFI_IPV6)
+                    else if (cache_addr.family_ == AF_INET6)
+                    {
+                        duplicate &= 
+                            !memcmp((*ait).addrIn6_.sin6_addr.un.u32_addr,
+                                addr->addr.u_addr.ip6.addr, 16);
```
atanisoft, line 635: it is likely best to use `sizeof` instead of a hard coded size of 16.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+                                cache_addr.addrIn_.sin_addr.s_addr)).c_str(),
+                            ntohs(cache_addr.port_));
+                    }
+#if defined(ESP_IDF_WIFI_IPV6)
+                    else if (cache_addr.family_ == AF_INET6)
+                    {
+                        memcpy(cache_addr.addrIn6_.sin6_addr.un.u32_addr,
+                            addr->addr.u_addr.ip6.addr, 16);
+                        cache_addr.addrIn6_.sin6_port = cache_addr.port_;
+                        LOG(INFO, "wifi: mDNS service discovered: %s\n"
```
atanisoft, line 659: consider log level adjustment

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+    }
+
+    switch (id)
+    {
+        default:
+            LOG(INFO, "wifi: wifi_event_handler() unknown id: %i", (int)id);
+            break;
+        case WIFI_EVENT_WIFI_READY:
+            started_ = true;
+            LOG(INFO, "wifi: WiFi started.");
```
atanisoft, line 807: consider log level adjustments throughout this file, we should likely only report connect / disconnect events specifically by default

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+            {
+                // No valid "fast" connect parameters.
+                esp_wifi_scan_start(&SCAN_CONFIG, false);
+                LOG(INFO, "wifi: STA start, scanning...");
+            }
+            else
+            {
+                // Try the fast connect parameters first.
+                sta_connect(ssid, pass, authmode, channel);
+                LOG(INFO, "wifi: STA start, ssid: %s, pass: %s, connecting...",
```
atanisoft, line 826: logging passwords should be avoided where possible

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+            if (evdata->reason == WIFI_REASON_AUTH_FAIL)
+            {
+                securityFailure_ = true;
+            }
+
+            bool try_fast_reconnect = false;
+            {
+                // We have to take a mutex because for some reason the
+                // IP_EVENT_STA_LOST_IP event does not get delivered when there
+                // is an unexpected WiFi disconnect. I guess it is implied?
```
atanisoft, line 862: it's possible there is a timing / sequencing issue here, it may show up later as lost when disconnected.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+            break;
+    }
+}
+
+//
+// EspIdfWiFi::init_config()
+//
+void EspIdfWiFi::init_config()
+{
+    esp_err_t result = nvs_open(NVS_NAMESPACE_NAME, NVS_READWRITE, &cfg_);
```
atanisoft, line 1049: we may need to have a call to `nvs_flash_init` which can return `ESP_ERR_NVS_NO_FREE_PAGES` which we need to handle ourselves as erase the NVS partition to force reinit. This is critical if the user has not called it previously from their code.


We should also have a call to `nvs_close` for every `nvs_open` call made, this should be the final call after any adjustments are done. We should not rely on nvs being open at all times and instead only open/close when needed.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+            break;
+    }
+
+    // Private configuration.
+    len = sizeof(privCfg_);
+    result = nvs_get_blob(cfg_, NVS_KEY_LAST_NAME, &privCfg_, &len);
+    switch (result)
+    {
+        case ESP_OK:
+            if (len == sizeof(privCfg_))
```
atanisoft, line 1091: should we consider a magic value here?

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+void EspIdfWiFi::init_wifi(WlanRole role)
+{
+    if (role == WlanRole::DEFAULT_ROLE)
+    {
+        role = userCfg_.mode_;
+    }
+
+    ESP_ERROR_CHECK(esp_netif_init());
+
+    /// @todo Do we need this?
```
atanisoft, line 1125: this appears to be unfortunately a required call from either this code or app code of the user. We can also switch to using `ESP_ERROR_CHECK` which internally will log the failure code on demand

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+    esp_err_t result = esp_event_loop_create_default();
+    if (result != ESP_OK)
+    {
+        LOG(LEVEL_ERROR, "wifi: esp_event_loop_create_default() failed: %s",
```
atanisoft, line 1129: we should likely use `LOG_ERROR(...)` if we make similar calls elsewhere

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+            break;
+    }
+    // Start the WiFi.
+    ESP_ERROR_CHECK(esp_wifi_start());
+
+    if (apIface_)
+    {
+        esp_netif_ip_info_t ip_info;
+        esp_netif_get_ip_info(
+            esp_netif_get_handle_from_ifkey("WIFI_AP_DEF"), &ip_info);
```
atanisoft, line 1173: we may want to move the IF names to a constant or better if we can reference their own constants (haven't looked for one though).

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+
+        char ip_addr[16];
+        inet_ntoa_r(ip_info.ip.addr, ip_addr, 16);
+        LOG(INFO, "wifi: Set up softAP with IP: %s", ip_addr);
+
+        LOG(INFO, "wifi: init_softap() finished. SSID:'%s' password:'%s'",
```
atanisoft, line 1179: consider adjusting log level

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+void EspIdfWiFi::init_softap(std::string ssid, std::string pass)
+{
+    apIface_ = esp_netif_create_default_wifi_ap();
+    esp_netif_set_hostname(apIface_, hostname_);
+
+    wifi_config_t conf;
+    memset(&conf, 0, sizeof(wifi_config_t));
+
+    conf.ap.max_connection = 4;
+    conf.ap.beacon_interval = 100;
```
atanisoft, line 1201: IIRC this is the default interval but it should be moved to a constant instead of hard coding with no notes as to what scale this would be (it is milliseconds IIRC)

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+    memset(&conf, 0, sizeof(wifi_config_t));
+
+    size_t ssid_len = std::min(ssid.size(), (size_t)MAX_SSID_LEN);
+    memcpy(conf.sta.ssid, ssid.c_str(), ssid_len);
+
+    size_t pass_len = std::min(pass.size(), (size_t)MAX_PASSPHRASE_LEN);
+    memcpy(conf.sta.password, pass.c_str(), pass_len);
+    staConnectPass_ = std::move(pass);
+
+    LOG(INFO, "wifi: STA SSID: %s, PASS: %s", conf.sta.ssid, conf.sta.password);
```
atanisoft, line 1245: consider log level adjustment

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+        conf.sta.threshold.authmode = (wifi_auth_mode_t)authmode;
+    }
+    else
+    {
+        conf.sta.threshold.authmode = WIFI_AUTH_OPEN;
+    }
+    conf.sta.pmf_cfg.required = false;
+    conf.sta.scan_method = WIFI_FAST_SCAN;
+    conf.sta.sort_method = WIFI_CONNECT_AP_BY_SIGNAL;
+    conf.sta.threshold.rssi = -100;
```
atanisoft, line 1257: consider moving this value to a constant and adding comments on it possibly.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+    std::string ssid, std::string pass, uint8_t authmode, uint8_t channel)
+{
+    if (ssid != privCfg_.last_.ssid_ || pass != privCfg_.last_.pass_ ||
+        authmode != privCfg_.last_.sec_|| channel != privCfg_.channelLast_)
+    {
+        LOG(INFO, "wifi: Update last STA.");
+        strncpy(privCfg_.last_.ssid_, ssid.c_str(), MAX_SSID_SIZE);
+        privCfg_.last_.ssid_[MAX_SSID_SIZE] = '\0';
+        strncpy(privCfg_.last_.pass_, pass.c_str(), MAX_PASS_SIZE);
+        privCfg_.last_.pass_[MAX_PASS_SIZE] = '\0';
```
atanisoft, line 1278: reading this further there may be an issue here, if `MAX_PASS_SIZE` is 64 the actual field size is similar when passing down to underlying APIs. There is no null character present and instead we need to pass down password length in the structure IIRC

src/freertos_drivers/common/WiFiInterface.hxx
bakerstu: I've addressed this fully through documentation and addition of a connect/disconnect user callback. @balazsracz Please have another look.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+    };
+
+    /// C structure version of the WiFi configuration settings.
+    struct WiFiConfigNVS
+    {
+        uint32_t magic_; ///< magic number to detect initialization
+        WlanRole mode_; ///< role that the device will operate (AP, STA, etc.)
+        WiFiConfigCredentialsNVS ap_; ///< access point configuration
+        WiFiConfigCredentialsNVS sta_[7]; ///< station profiles
+    };
```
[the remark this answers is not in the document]
atanisoft: I think that might make sense. The main concern is to try and avoid the future case by having a small bit of padding in there that we can leverage from a default value of zero.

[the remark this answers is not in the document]
bakerstu: @atanisoft The trouble I see is choosing a number of bytes to reserve. The profiles are quite large in terms of bytes, so if we are thinking about more profiles, then that is a huge number of bytes to reserve. Balazs also suggested to change the NVS "blob" names to have a ".v1" suffix on them. That could give us a get out of jail free card to have different versions of this down the road without necessarily breaking backward compatibility.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+//
+// EspIdfWiFi::connect()
+//
+WlanConnectResult EspIdfWiFi::connect(
+    const char *ssid, const char *pass, SecurityType sec_type)
+{
+    securityFailure_ = false;
+    sta_connect(
+        std::string(ssid), std::string(pass), sec_type_translate(sec_type));
+    return WlanConnectResult::CONNECT_OK;
```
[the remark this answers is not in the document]
bakerstu: Reworked, take another look.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+    return static_cast<EspIdfWiFi*>(WiFiInterface::instance())
+        ->mdns_lookup(service, hints, addr);
+}
+
+/// Helper method for getifaddrs().
+/// @param iface interface instance pointer
+/// @param name interface name
+static struct ifaddrs *getifaddrs_helper(esp_netif_t *iface, const char *name)
+{
+#if defined(ESP_IDF_WIFI_IPV6)
```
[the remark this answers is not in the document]
atanisoft: That makes sense, though if we can reuse an existing purpose based define we should rather than introduce a replacement option. It will invariably lead to drift and frustrations for all unfortunately.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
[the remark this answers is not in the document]
atanisoft: looks much better, thank you.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+        std::string(ssid), std::string(pass), sec_type_translate(sec_type));
+    return WlanConnectResult::CONNECT_OK;
+}
+
+//
+// EspIdfWiFi::disconnect()
+//
+void EspIdfWiFi::disconnect()
+{
+    /// @todo Do we need to manually handle status flags and user callbacks, or
```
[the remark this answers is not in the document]
atanisoft: Thanks

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,626 @@
...
+
+    /// Stop the WiFi.
+    void stop() override;
+
+    /// Connect to access point.
+    /// @param ssid access point SSID
+    /// @param pass access point password
+    /// @param sec_type access point security type
+    /// @return WlanConnectResult::CONNECT_OK upon connect, else error code
+    WlanConnectResult connect(
```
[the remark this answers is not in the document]
atanisoft: Looks much better thank you

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+
+//
+// EspIdfWiFi::mdns_service_add()
+//
+void EspIdfWiFi::mdns_service_add(const char *service, uint16_t port)
+{
+    std::string name;
+    std::string proto;
+    mdns_split(service, &name, &proto);
+    LOG(INFO, "wifi: MDNS service add, name: %s, proto: %s, port: %u",
```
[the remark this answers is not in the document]
atanisoft: I'm all for deferral at this stage since it is still in development but wanted to ensure it is documented somewhere for review.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+    std::string proto;
+    mdns_split(service, &name, &proto);
+    LOG(INFO, "wifi: MDNS service remove, service: %s, proto: %s",
+        name.c_str(), proto.c_str());
+    OSMutexLock locker(this);
+    if (!mdnsAdvInhibit_)
+    {
+        ::mdns_service_remove(name.c_str(), proto.c_str());
+    }
+    for (auto it = mdnsServices_.begin(); it != mdnsServices_.end(); ++it)
```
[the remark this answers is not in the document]
atanisoft: `std::erase` for most cases was present in C++11, but you are correct that `std::erase_if` appears to be c++20 (enforced by ESP-IDF in v5.3+ IIRC)

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+        struct addrinfo *current = *addr;
+        struct addrinfo *last = nullptr;
+
+        for (auto lit = (*it).addr_.begin(); lit != (*it).addr_.end(); ++lit)
+        {
+// The caller is expected to free the struct addrinfo using the method
+// freeaddrinfo. In the lwIP implementation, they use a special buffer pool
+// to free the struct addrinfo to. Therefore, we must also allocate from the
+// same pool.
+#if !defined(LWIP_HDR_MEMP_H)
```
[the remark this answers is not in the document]
atanisoft: Hmm, there used to be a `CONFIG_LWIP_MEMP_MEM_MALLOC` flag protecting it's usage that could be disabled. It looks like may be they moved to always enabled. When disabled it used to leverage an internal lwip memory pool feature instead. I've not found any specific need for this specifically during testing, I think it should likely go through further reviews / testing across supported ESP-IDF versions (v5.3+ currently)

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
atanisoft: I can't disagree, I similarly prefer explicit use statements where needed to restrict pollution of namespaces.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
atanisoft: I've seen it handled both ways but I do agree moving this over to app code would make sense. I'd suggest document it as required and if possible add a way to test that they did what is required.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
atanisoft: This is coming from OpenMRN shorthand logging formatting https://github.com/bakerstu/openmrn/blob/120ad4ea4b92cd1a461b82ec369fd510c8b8d798/src/utils/logging.h#L124.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1355 @@
...
+//
+void EspIdfWiFi::init_softap(std::string ssid, std::string pass)
+{
+    apIface_ = esp_netif_create_default_wifi_ap();
+    esp_netif_set_hostname(apIface_, hostname_);
+
+    wifi_config_t conf;
+    memset(&conf, 0, sizeof(wifi_config_t));
+
+    conf.ap.max_connection = 4;
```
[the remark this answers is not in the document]
atanisoft: I think that will work. There were some adjustments in later IDF v5.3+ LwIP code that increased socket limits beyond 16 max and there were some examples of APs exceeding 8 concurrent connections (I'm testing such as well). If we have it as configurable in some form I think it would be good.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
atanisoft: A comment indicating it would be great, I'm not a fan of hard coded "magic" numbers with clear comments on them.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
atanisoft: Yes, I think it's covered so far.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1409 @@
...
+    config_sync();
+}
+
+//
+// EspIdfWiFi::set_role()
+//
+void EspIdfWiFi::set_role(WlanRole new_role)
+{
+    // This is a single byte copy, it is already atomic.
+    static_assert(sizeof(userCfg_.mode_)== sizeof(uint8_t));
```
atanisoft, line 263: missing space before ==

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1409 @@
...
+        }
+    }
+}
+
+//
+// EspIdfWiFi::ip_event_handler()
+//
+void EspIdfWiFi::ip_event_handler(esp_event_base_t base, int32_t id, void *data)
+{
+    if (base != IP_EVENT)
```
atanisoft, line 1011: Technically this should never happen, I wonder if conversion to `HASSERT(base == IP_EVENT);` would make sense?

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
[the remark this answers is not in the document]
bakerstu: In retrospect, it would probably have been a better design to not rely on freeaddrinfo() as the opposite of the mdsn_lookup() API. We could move in that direction over time, but it has wider implications in existing code which we would need to address.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
atanisoft: looks much cleaner.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,907 @@
...
+        WlanRole mode_; ///< role that the device will operate (AP, STA, etc.)
+        uint8_t padding_[3]; ///< padded for alignment
+        WiFiConfigCredentialsNVS ap_; ///< access point configuration
+        WiFiConfigCredentialsNVS sta_[N]; ///< station profiles
+    };
+    static_assert(sizeof(WiFiConfigNVS<HWDefs::MAX_STA_PROFILES>::mode_) ==
+        sizeof(uint8_t));
+    static_assert(
+        sizeof(WiFiConfigNVS<HWDefs::MAX_STA_PROFILES>) ==
+            4 + 4 + 98 + (98 * HWDefs::MAX_STA_PROFILES),
```
atanisoft, line 834: I suspect a comment on where expected sizes come from would be good (and elsewhere) but this is definitely a welcome addition here. Thanks.

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,343 @@
...
+public:
+    /// Interface index by type.
+    enum Interface : uint8_t
+    {
+        IFACE_STA, ///< STA mode interface
+        IFACE_AP, ///< AP mode interface
+    };
+
+    /// Security types.
+    enum SecurityType : uint8_t
```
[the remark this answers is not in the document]
bakerstu: @balazsracz Yeah, I've given this some thought too. Right now, with the CC32xx version, we are also not consistent with these, and I kind of copied some of that inconsistency here. We should probably decide on one or the other, not enums in both. I have a slight preference to put them in the interface, but I'm also fine to put them in the WiFiDefs.hxx. WDYT?

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,343 @@
...
+    /// @param entry location to fill in the network entry
+    /// @param index index in the network entry list to get
+    /// @return 0 on success, else -1 if index is beyond the list of entries.
+    virtual int network_get(NetworkEntry *entry, unsigned index) = 0;
+
+    /// Initiate scanning of available networks.
+    virtual void scan() = 0;
+
+    /// Get the RSSI of the AP in STA mode.
+    /// @return signal strength, 0 when not connected to an access point
```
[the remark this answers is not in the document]
bakerstu: Done.

src/freertos_drivers/common/WiFiInterface.hxx
```diff
@@ -0,0 +1,343 @@
...
+
+    /// Set the callback for when a wlan connect or disconnect occurs.
+    /// Back-to-back deliveries of "disconnected" are possible without first
+    /// transitioning to a connected state. This can occur when a connection
+    /// "attempt" has failed to succeed.
+    /// @param connected true if connected, else false if disconnected
+    /// @param result result code for the connection event
+    /// @param ssid SSID of the connection event
+    virtual void set_wlan_connect_callback(std::function<void(
+        bool connected, ConnectionResult result, std::string &ssid)> callback)
```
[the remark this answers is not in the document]
bakerstu: Fixed.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,907 @@
...
+    ///        respond
+    /// @param addrinfo structure containing one or more service addressess that
+    ///        match the enquery, else nullptr if no matching service found
+    /// @return 0 upon success, or appropriate EAI_* error on failure, use
+    ///         ::freeaddrinfo() to free up memory allocated to the non nullptr
+    ///         *addr returned
+    int mdns_lookup(
+        const char *service, struct addrinfo *hints, struct addrinfo **addr);
+
+    /// Start continuous scan for mDNS service name.
```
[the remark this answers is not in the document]
bakerstu: Done.

src/utils/format_utils.cxx
```diff
@@ -333,3 +333,30 @@ string ipv4_to_string(uint8_t ip[4])
 
     return ret;
 }
+
+string ipv6_to_string(uint8_t ip[16])
+{
+    string ret;
+    ret.reserve(32+8);
+    char tmp[10];
+    for (int i = 0; i < 16; ++i)
```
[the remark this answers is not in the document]
bakerstu: newlib does not have this API. lwIP does implement a version of it though. Maybe we should use it, in the ESP-IDF port, and roll back adding this to the format_utils. We could also "ifdef" the implementation in format utils. Thoughts?

src/utils/format_utils.hxx
```diff
@@ -183,6 +183,14 @@ inline string ipv4_to_string(uint32_t ip)
     return ipv4_to_string((uint8_t*)&ip);
 }
 
+/// Formats an IPv6 address to string.
+/// @param ip a 16-byte array storing the IPv6 address, ip[0] will be printed
+///        at the beginning (network endianness)
+/// @return a string containing a collon-separated printout of the given IPv6
```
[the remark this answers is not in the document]
bakerstu: Fixed.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,914 @@
...
+    /// essentially the connection profiles (SSID, password, security type).
+    template<size_t N> struct WiFiConfigNVS
+    {
+        uint32_t magic_; ///< magic number to detect initialization
+        WlanRole mode_; ///< role that the device will operate (AP, STA, etc.)
+        uint8_t padding_[3]; ///< padded for alignment
+        WiFiConfigCredentialsNVS ap_; ///< access point configuration
+        WiFiConfigCredentialsNVS sta_[N]; ///< station profiles
+    }; // template<size_t N> struct WiFiConfigNVS
+    static_assert(sizeof(WiFiConfigNVS<HWDefs::MAX_STA_PROFILES>::mode_) ==
```
balazsracz, line 837: let's make a typedef for the actual wificonfignvs so that the template does not need to be written out everywhere:
template <size_t N> typedef WifiConfigNVSTemplate {...};
typedef WifiConfigNVSTemplate<HWDefs::MAX_STA_PROFILES> WifiConfigNVS;

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,914 @@
...
+    /// @return configured WiFi role
+    WlanRole get_role() override
+    {
+        return userCfg_.mode_;
+    }
+
+    /// Retrieve current access point config, including password.
+    /// @param ssid access point SSID
+    /// @param pass access point password
+    void get_ap_config(std::string *ssid, std::string *pass) override
```
balazsracz, line 859: why does this get function not return the security type?

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1367 @@
...
+                esp_wifi_scan_start(&SCAN_CONFIG, false);
+                LOG(INFO, "wifi: STA start, scanning...");
+            }
+            else
+            {
+                // Try the fast connect parameters first.
+                sta_connect(ssid, pass, authmode, channel);
+#if (LOGLEVEL == VERBOSE)
+                LOG(VERBOSE,
+                    "wifi: STA start, ssid: %s, pass: %s, connecting...",
```
balazsracz, line 782: I'm not excited about putting the password to the log, even if it is guarded by verbosity. Maybe put only the length?

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1367 @@
...
+            LOG(INFO, "wifi: STA connected, channel: %u, authmode: %u",
+                sdata->channel, (unsigned)sdata->authmode);
+            // There is a very small chance for a race condition whereby another
+            // call "sta_connect()" call has been made with a different password
+            // after this connection has been established. The chances are very
+            // small and the harm is very limited to the next fast connect
+            // attempt failing.
+            last_sta_update(std::string((char*)sdata->ssid, sdata->ssid_len),
+                staConnectPass_, sdata->authmode, sdata->channel);
+            std::string ssid((char*)sdata->ssid, sdata->ssid_len);
```
balazsracz, line 806: you could move this one line up as you have duplicate code now.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1367 @@
...
+            ///       friendly. Revisit this when updating the IDF version.
+            int idx = -1;
+            uint16_t number;
+            esp_wifi_scan_get_ap_num(&number);
+            LOG(INFO, "wifi: Scan done, number of records: %u.", number);
+            number = std::min(number, static_cast<uint16_t>(10));
+
+            // This takes a lot of memory. Not the best API design. See note
+            // above.
+            wifi_ap_record_t *ap_records = new wifi_ap_record_t[10];
```
balazsracz, line 863: consider allocating with `[number]`

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1367 @@
...
+                                sec = profile_sec;
+                                conn_channel = ap_records[i].primary;
+                                // Found a match, stop looking.
+                                break;
+                            }
+                        } while (index >= 0);
+                    }
+                }
+            }
+            delete ap_records;
```
balazsracz, line 905: `delete [] ap_records`

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,907 @@
...
+            return WlanRole::STA;
+        }
+        return WlanRole::UNKNOWN;
+    }
+
+    /// Get a list of available networks. This is based on a prior scan. Use
+    /// rescan() and wait for the scan to complete to refresh the list.
+    /// @param entries returns a list of available network entries
+    /// @param count max size of the entry list to return.
+    /// @return number of valid network entries in the list, same as
```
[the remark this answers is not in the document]
bakerstu: Removed return value.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,907 @@
...
+        {
+            return WlanRole::STA;
+        }
+        return WlanRole::UNKNOWN;
+    }
+
+    /// Get a list of available networks. This is based on a prior scan. Use
+    /// rescan() and wait for the scan to complete to refresh the list.
+    /// @param entries returns a list of available network entries
+    /// @param count max size of the entry list to return.
```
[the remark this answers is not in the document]
bakerstu: Yes, this is not needed. Removed.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,907 @@
...
+    /// Add service to mDNS server.
+    /// @param service service, e.g. _openlcb-can._tcp
+    /// @param port service port
+    void mdns_service_add(const char *service, uint16_t port);
+
+    /// Remove service from mDNS server.
+    /// @param service service, e.g. _openlcb-can._tcp
+    void mdns_service_remove(const char *service);
+
+    /// Lookup an mDNS name.
```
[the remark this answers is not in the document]
bakerstu: Yes, non-blocking. Updated the comment.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,907 @@
...
+        *sec_type = sec_type_encode(userCfg_.ap_.sec_);
+    }
+
+    /// Change the default role. This will be used in the next start() if the
+    /// DEFAULT_ROLE is specified. The new setting takes effect when the
+    /// device is restarted (either via reboot or stop + start)
+    /// @param new_role new role, must not be UNKOWN or DEFAULT_ROLE
+    void set_role(WlanRole new_role) override
+    {
+        // This is a single byte copy, it is already atomic.
```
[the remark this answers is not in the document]
bakerstu: A lock is not required to call config_sync(). I updated the comment header for this API.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,914 @@
...
+        OSMutexLock locker(&lock_);
+        int index = find_sta_profile("");
+        if (index >= 0)
+        {
+            str_populate<MAX_PASS_SIZE>(userCfg_.sta_[index].pass_, pass);
+            userCfg_.sta_[index].sec_ = sec_type_encode(sec_type);
+            str_populate<MAX_SSID_SIZE>(userCfg_.sta_[index].ssid_, ssid);
+            config_sync();
+        }
+        config_sync();
```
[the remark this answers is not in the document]
bakerstu: Not only is it redundant, it should not be called if the proceeding if statement is false. Removed.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
bakerstu: Done.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,914 @@
...
+        (void)sec_type; // unused;
+    }
+
+    /// Find a WiFi STA profile that matches the given SSID
+    /// @param ssid SSID to match
+    /// @param pass where to fill in the password, ignored if nullptr
+    /// @param sec where to fill in the security, ignored if nullptr
+    /// @param index index to start the linear search from
+    /// @return index within the wifi profiles config, else -1 if not found
+    int find_sta_profile(std::string ssid, std::string *pass = nullptr,
```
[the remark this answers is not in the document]
bakerstu: Fixed.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1367 @@
...
+    {
+        if ((*it).service_ == service)
+        {
+            match = true;
+        }
+    }
+    if (!match)
+    {
+        // Start looking for a new service.
+        mdnsClientCache_.emplace_back(std::move(std::string(service)));
```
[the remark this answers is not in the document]
bakerstu: @balazsracz No. Do you think we should?

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1367 @@
...
+//
+int EspIdfWiFiBase::mdns_lookup(
+    const char *service, struct addrinfo *hints, struct addrinfo **addr)
+{
+    bool looking = false;
+    *addr = nullptr;
+    OSMutexLock locker(&lock_);
+    for (auto it = mdnsClientCache_.begin(); it != mdnsClientCache_.end(); ++it)
+    {
+        if ((*it).service_ != service)
```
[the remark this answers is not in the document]
bakerstu: Fixed all instances.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1367 @@
...
+//
+// EspIdfWiFiBase::factory_default()
+//
+void EspIdfWiFiBase::factory_default()
+{
+    LOG(INFO, "wifi: factory_default()");
+    // This mutex lock will hold off a configuration sync until all the updates
+    // are made.
+    OSMutexLock locker(&lock_);
+    setup_ap(default_ap_ssid(), default_ap_password(), SEC_WPA2);
```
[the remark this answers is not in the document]
bakerstu: Good idea. I also did this for STA.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
bakerstu: Let's just remove the password printing.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
bakerstu: Good catch.

src/utils/format_utils.cxx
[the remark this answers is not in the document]
bakerstu: @balazsracz WDYT?

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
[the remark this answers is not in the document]
bakerstu: This is now a configuration option.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
bakerstu: Added.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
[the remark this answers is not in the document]
bakerstu: @atanisoft This has been reworked somewhat with the added "template". Please have another look

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+    {
+        if (it->service_ == service)
+        {
+            match = true;
+        }
+    }
+    if (!match)
+    {
+        // Start looking for a new service.
+        mdnsClientCache_.emplace_back(std::move(std::string(service)));
```
balazsracz, line 416: emplace_back calls the constructor. You don't need to explicitly consturct and move a string. It will be done automatically.
So it's enough to say `mdnsClientCache_.emplace_back(service)`

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+                else
+                {
+                    // Unknown protocol.
+                    continue;
+                }
+                bool duplicate = false;
+                for (auto ait = it->addr_.begin();
+                    !duplicate && ait != it->addr_.end(); ++ait)
+                {
+                    duplicate = ait->family_ == cache_addr.family_;
```
balazsracz, line 572: it would be nice to factor out this code into an operator== on MDNSCacheAddr. The code would be much more readable here.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+        case WIFI_EVENT_SCAN_DONE:
+        {
+            /// @todo The newer IDF versions have a new api that can read
+            ///       AP records one at a time, which is much more memory
+            ///       friendly. Revisit this when updating the IDF version.
+            int idx = -1;
+            uint16_t number;
+            esp_wifi_scan_get_ap_num(&number);
+            LOG(INFO, "wifi: Scan done, number of records: %u.", number);
+            number = std::min(number, static_cast<uint16_t>(10));
```
balazsracz, line 871: couldwe move this "10" to a static constant?
I remember a situation we encountered in a convention where a hotel had about 20 wifi aps visible and we were unableto connect to LAYOUTWIFI because all we saw in the list was hotel_ap hotel_ap etc.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+
+            // This takes a lot of memory. Not the best API design. See note
+            // above.
+            wifi_ap_record_t *ap_records = new wifi_ap_record_t[number];
```
balazsracz, line 875: instead of dynamically allocatingthis memory I think we should consider allocating it statically. This avoids memory fragmentation. We're not super tight on RAM, so spending 1k of memory on this would be perfectly acceptable. We can use the constant I mentioned above. The callback thread ensures we will only ever use it one at a time.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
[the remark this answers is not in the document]
atanisoft: The new approach of templatizing it should help in that regard. The only point that would be remaining is ensure there is migration strategies in place for loading old and saving new, but that I think should be handled separately from this as there are likely a lot more implications to such behaviors. If you can log a tracking ticket that would be great, otherwise I have no issues right now

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
atanisoft: Restricting to 10 maximum is also a potential problem of when it might be applied, from the snippet of context we will display potentially 15 records and the code then will force restrict to 10 maximum. We should defer restriction until after deduplication (hotel case having each floor SSID being identical but different BSSID) and filtering to "interesting" names (LAYOUTWIFI case).

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+            esp_wifi_scan_get_ap_records(&number, ap_records);
+            std::string ssid(MAX_SSID_SIZE + 1, '\0');
+            std::string pass(MAX_PASS_SIZE + 1, '\0');
+            uint8_t sec = WIFI_AUTH_WPA2_PSK;
+            uint8_t conn_channel = 0;
+            {
+                OSMutexLock locker(&lock_);
+                scanResults_.clear();
+                for (unsigned i = 0; i < number; ++i)
```
balazsracz, line 884: Are the returned results sorted by rssi? Or by channel? Or arbitrarily?

balazsracz: sorted by rssi.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+                {
+                    // Cache the scan record for the user to read later.
+                    scanResults_.emplace_back((char*)ap_records[i].ssid,
+                        sec_type_encode(ap_records[i].authmode),
+                        ap_records[i].rssi);
+
+                    if (staIface_ && !connected_ && idx < 0)
+                    {
+                        int index = 0;
+                        do
```
balazsracz, line 894: I don't think this loop should be here. We don't support having the same ssid in more than one profile, even if the security parameter is different.
(Anyway it's an infinite loop, as there is a missing increment of the index.)

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+            }
+            delete [] ap_records;
+            if (staIface_ && !connected_)
+            {
+                // If in STA mode and not connected, always be trying to make
+                // a connection.
+                if (idx < 0)
+                {
+                    // No profile match found, scan again.
+                    esp_wifi_scan_start(nullptr, false);
```
balazsracz, line 925: If we're continuously scanning, does that compromise the quality of the AP? I guess we don't have more than one radio.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+                break;
+            }
+            // Not initialized yet.
+            // fall through
+        case ESP_ERR_NVS_NOT_FOUND:
+            // fall through
+        case ESP_ERR_NVS_INVALID_LENGTH:
+            // Reset to factory defaults.
+            memset(&privCfg_, 0, sizeof(privCfg_));
+            nvs_set_blob(cfg, NVS_KEY_LAST_NAME, &privCfg_, sizeof(privCfg_));
```
balazsracz, line 1079: same here, i don't think we should write an empty blob.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+                                std::string((char*)ap_records[i].ssid),
+                                &profile_pass, &profile_sec, index);
+                            if (index >= 0 &&
+                                ap_records[i].authmode >= profile_sec)
+                            {
+                                // profile SSID and security mode match.
+                                idx = index;
+                                ssid = (char*)ap_records[i].ssid;
+                                pass = std::move(profile_pass);
+                                sec = profile_sec;
```
balazsracz, line 908: this should be the ap authmode, not the profile authmode

src/freertos_drivers/common/WiFiInterface.hxx
balazsracz: looks good

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
balazsracz: It is a bit confusing that there is both a public and a nonpublic virtual function called get_ap_config but with different arguments.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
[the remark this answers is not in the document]
bakerstu: Todo what though? Can you be more specific?

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,915 @@
...
+#endif
+        };
+    }; // struct MDNSCacheAddr
+
+    /// Metadata for a subscribed mDNS client.
+    struct MDNSCacheItem
+    {
+        /// Constructor.
+        MDNSCacheItem(std::string service)
+            : service_(service)
```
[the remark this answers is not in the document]
bakerstu: Fixed.

src/utils/format_utils.cxx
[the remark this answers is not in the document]
bakerstu: I'm using inet_ntop. I've rolled back my format_utils changes.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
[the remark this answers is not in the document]
bakerstu: For now, we bumped the background scan up to 1200 seconds, 20 minutes. This is okay because we fixed mdns_lookup(). I've also added a todo comment for the timeout param to explain the implications.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+                    cache_addr.family_ = AF_INET6;
+                }
+#endif
+                else
+                {
+                    // Unknown protocol.
+                    continue;
+                }
+                bool duplicate = false;
+                for (auto ait = it->addr_.begin();
```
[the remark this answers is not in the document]
bakerstu: Added an implementation of the Time To Live (TTL).

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
bakerstu: We will upgrade the IDF to resolve this.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
bakerstu: I'm not sure. I think it is probably okay. I've added a todo comment to revisit if we run into issues.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,915 @@
...
+    };
+
+    /// Magic number to detect initialization.
+    static constexpr uint32_t WIFI_CONFIG_INIT_MAGIC = 0x6160CBC6;
+
+    /// Magic number to detect initialization.
+    static constexpr uint32_t PRIV_CONFIG_INIT_MAGIC = 0xBD959C63;
+
+    /// NVS namespace for the wifi configuration.
+    static constexpr char NVS_NAMESPACE_NAME[] = "wifi_config";
```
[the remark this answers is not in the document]
bakerstu: Too long to fit, I've changed it to "wifi_config"

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+        return;
+    }
+
+    // User configuration.
+    uint32_t magic;
+    size_t user_config_len;
+    void *user_config = get_user_config(&user_config_len, &magic);
+    size_t len = user_config_len;
+    result = nvs_get_blob(cfg, NVS_KEY_USER_NAME, user_config, &len);
+    switch(result)
```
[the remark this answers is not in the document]
bakerstu: You are correct, this has been reworked into the derived object that performs user config.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
bakerstu: I've added a check such formatting the NVS is only allowed if a derived object says it is allowed, default implementation does not allow it. I've also updated this with the magic number, which was a bug.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -0,0 +1,1375 @@
...
+//
+void EspIdfWiFiBase::last_sta_get(
+    std::string *ssid, std::string *pass, uint8_t *authmode, uint8_t *channel)
+{
+    std::string profile_pass;
+    uint8_t profile_sec;
+    int index = 0;
+    do
+    {
+        index = find_sta_profile(
```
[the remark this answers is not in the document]
bakerstu: This has been reworked to not require a profile for fast connect.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
bakerstu: I will work either way, but there are some corner cases where ap authmode is better, so I changed it.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
bakerstu: I've added an operator == and some helper API.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -0,0 +1,907 @@
...
+#define ESP_IDF_WIFI_IPV6
+#endif
+#endif
+
+/// Object for managing WiFi and network interfaces. The following sequence
+/// must be called within or after app_main() in the following order:
+///   -# init()
+///       - performs initialize before starting WiFi
+///   -# setup_ap()
+///       - option, can be used to override the stored AP profile in bootloader
```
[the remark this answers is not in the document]
bakerstu: Updated comments.

PR #844, opened 2025-07-13
ESP IDF 5.4.2 Port

bakerstu:


include/esp-idf/can_frame.h
balazsracz, line 1: let's add a README.md to explain why this directory is needed and what should go in here

balazsracz (approved):

include/esp-idf/can_frame.h
bakerstu: Done.

src/freertos_drivers/esp32/Esp32Gpio.hxx
```diff
@@ -566,7 +566,7 @@ template <class Defs> struct GpioInputPUPD : public GpioInputPin<Defs, true, tru
     {                                                                          \
         static const gpio_num_t PIN_NUM = (gpio_num_t)NUM;                     \
     public:                                                                    \
-        static const gpio_num_t pin()                                          \
+        static gpio_num_t pin()                                                \
```
atanisoft, line 569: I'm curious on the removal of const here, was this due to a compilation error/warning?

src/os/TempFile.cxx
```diff
@@ -58,7 +58,7 @@ TempDir::TempDir()
 #else
     dirName_ = "./openmrntmpdirXXXXXX";
 #endif
-    dirName_.c_str();
+    (void)dirName_.c_str();
```
atanisoft, line 61: Also curious on this and the reference below

src/freertos_drivers/esp32/Esp32Gpio.hxx
balazsracz: Yes the compiler issued a warning on this.
There is no use of returning an integral type by value as const. This will not force the caller to assign it to a const variable. Const only makes a difference for pointer and reference types, where it talks about the object being pointed *to*.

src/os/TempFile.cxx
balazsracz: This was a warning on "unused return value".

atanisoft: Interesting, Did we have a specific need to call c_str() here in the first place though?

PR #845, opened 2025-07-13
ESP IDF WiFi Tweak User Config

bakerstu:
- Removes NVS from class EspIdfWiFi
- Adds classEspIdfWiFiNVS, derived from EspIdfWiFi to implement NVS non-volatile user configuration storage.

### DIFF
--- a/src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
+++ b/src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
@@ -951,9 +951,6 @@ protected:
     /// Magic number to detect initialization.
     static constexpr uint32_t WIFI_CONFIG_INIT_MAGIC = 0x6160CBC6;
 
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -1083,32 +1127,61 @@ private:
...
-                if (sec)
-                {
-                    *sec = userCfg_.sta_[i].sec_;
-                }
-                return i;
-            }
+            sec = WiFiDefs::SEC_OPEN;
         }
-        return -1;
+        str_populate<this->MAX_SSID_SIZE>(
```
balazsracz, line 1145: it looks like this code is very similar to the init_config_user implementation in the base class

bakerstu: Fixed.

PR #846, opened 2025-07-13
Fix crash caused by EspIdfWiFi::stop() API

bakerstu:


src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -945,6 +946,12 @@ void EspIdfWiFiBase::wifi_event_handler(
                 mdns_disable_sta();
                 ip_acquired(IFACE_STA, false);
             }
+            if (!initialized_)
+            {
+                // This is likely a "stop" condition. Do not try to reconnect
+                // or scan for APs. Otherwise, there will be a crash.
+                break;
+            }
```
balazsracz, line 954: move up the callback

balazsracz (approved):

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
bakerstu: Good catch. Done.

PR #852, opened 2025-07-30
Arduino support for Feather M4 CAN 

balazsracz:
- Adds a CAN driver wrapper for arduino for ATSAME51's fdcan controller, using an external library.
- Makes the Arduino Can base class have virtual methods.
- Fixes some annoying compile errors due to global namespace symbol conflicts.
- Adds example sketches: one for USB-CAN and one for a clock consumer.

### DIFF
--- a/src/freertos_drivers/arduino/Can.cxx
+++ b/src/freertos_drivers/arduino/Can.cxx
@@ -36,5 +36,9 @@
 #include "can_frame.h"
 #include <cstdint>
 
+namespace openmrn_arduino {
+
 unsigned Can::numReceivedPackets_{0};
 unsigned Can::numTransmittedPackets_{0};
+
+} // namespace openmrn_arduino
--- a/src/freertos_drivers/sam/FeatherM4Can.hxx
+++ b/src/freertos_drivers/sam/FeatherM4Can.hxx
@@ -1,0 +1,249 @@
+/** \copyright
+ * Copyright (c) 2025, Balazs Racz
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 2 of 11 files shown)


src/freertos_drivers/arduino/Can.cxx
```diff
@@ -36,5 +36,9 @@
 #include "can_frame.h"
 #include <cstdint>
 
+namespace openmrn_arduino {
```
atanisoft, line 39: I believe the trailing `{` should be on the next line, this likely should be applied elsewhere as well

src/freertos_drivers/sam/FeatherM4Can.hxx
```diff
@@ -0,0 +1,249 @@
...
+/// #define CAN1_MESSAGE_RAM_SIZE (1728)
+/// #include <ACANFD_FeatherM4CAN.h>
+///
+/// #include <OpenMRNLite.h>
+/// FeatherM4Can can_driver;
+///
+/// void setup() {
+///   ...
+///   HASSERT(can_driver.begin());
+///   openmrn.add_can_port(&CanDriver);
```
atanisoft, line 62: should this be `CanDriver` or `can_driver`?

src/utils/DirectHub.cxx
```diff
@@ -39,10 +39,14 @@
 #include "utils/DirectHub.hxx"
 
 #include <algorithm>
+#include <vector>
+
```
atanisoft, line 43: should this be after `fcntl.h` include?

atanisoft (approved):

src/freertos_drivers/sam/FeatherM4Can.hxx
balazsracz: fixed

PR #864, opened 2025-08-09
Adds pause and resume API to the BusMaster.

balazsracz:


bakerstu (approved):
I'm not sure about the use case, other than maybe unit testing, but it seems correct. I wonder if the needShutdown_ parameter should be a bool type and/or std::atomic.

I recently discovered through testing that while 'bit' variables save RAM, they actually add a non-trivial amount of code space for targets that do not have single instruction bit operators, such as ARM, RISC-V, etc.

balazsracz:
> I'm not sure about the use case, other than maybe unit testing, but it seems correct. I wonder if the needShutdown_ parameter should be a bool type and/or std::atomic.
> 
> I recently discovered through testing that while 'bit' variables save RAM, they actually add a non-trivial amount of code space for targets that do not have single instruction bit operators, such as ARM, RISC-V, etc.

I made it a bool. It does not make a difference here, as there is no other bitfield variable.

PR #870, opened 2025-08-12
Fix relative includes

balazsracz:
Replaces all relative include paths with absolute directory includes.

Exceptions:
- files in include/freertos are left alone. These seem to be replacement for system headers.
- stm32f_hal_conf is left without a path, since this is also provided by the arduino environment but at a different path.
- cases where there was more than one file with the same name.


bakerstu (approved):
I think this is a net positive. Before merging this, we should also remove the -I paths from the various Makefiles that allowed these to build without the "relative" paths in the first place. That will prevent future regression and ensure we didn't miss anything.

atanisoft:
@balazsracz can you check the stm32 build, it looks like it may be related to the include shuffle and something missing in libify.sh.

atanisoft:
> we should also remove the -I paths from the various Makefiles that allowed these to build without the "relative" paths in the first place.

I fully agree on the cleanup of the -I paths but I believe that it will still compile either way due to how GCC handles `".."` vs `<..>` with `".."` being handled as relative to the current file (applies to short paths only likely) before it goes to the -I path approach.

PR #872, opened 2025-08-15
Fixes formatting of fall through comments

balazsracz:
Recent gcc compilers are emitting warning in places where we have fall through comments.
This covers some known formatting problems, for example when there is a closing brace after the fall through comment.

### DIFF
--- a/src/openlcb/Bootloader.hxx
+++ b/src/openlcb/Bootloader.hxx
@@ -366,8 +366,8 @@ void handle_memory_config_frame()
                 set_error_code(DatagramDefs::INVALID_ARGUMENTS);
                 return;
             }
-            // fall through
         }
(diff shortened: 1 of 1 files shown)


atanisoft:
Ahh, didn't see that note on it. However, it doesn't seem to do much beyond providing it only as an attribute to the C++ compiler and C may not notice it. The `// fall through` (case insensitive, its a mix currently it seems) is more of C style and provided more user doc side of it, the compiler happens to also interpret it as an attribute but is not obligated to do so (standard also says it shouldn't do what we observe it doing even). 

I'd keep it as it does work with C++20 (or later, used by ESP-IDF 5.4+). We should likely review if we want / can move up to C++17 (likely C would need to move up as well).

PR #874, opened 2025-08-16
Add STM32431 to the targets/freertos.armv7.exc

bakerstu:


src/freertos_drivers/common/Fileio.cxx
```diff
@@ -412,6 +412,15 @@ int _isatty_r(struct _reent *reent, int fd)
     return 1;
 }
 
+/** Get the tty information of a file or device.
+ * @param fd file descriptor determine if it is a tty
+ * @return 1 if a tty, else 0
+ */
+int _isatty(int fd)
```
bakerstu, line 419: @balazsracz I had to add this to get the "exc" target (different libc I think) to link. WDYT? Have you seen this?

balazsracz: I've seen this. Maybe flex is using istty, because i had to have it in HwInit.cxx:
```
openmrn$ grep -A 3 isatty ./boards/st-stm32f767zi-nucleo/HwInit.cxx
extern int isatty(int fd) {
  return 0;
}
```

PR #875, opened 2025-08-31
Simplify mDNS Lookup Behavior

### LINKED ISSUE #871: EspIdfWiFi::mdns_lookup() Behavior
EspIdfWiFi::mdns_lookup() is a blocking call. Normally, that's okay. However, the ESP mDNS implementation has a quirk that advertising cannot be masked from "registered" interfaces (AP, STA, ETH...). Sometimes, we only want to advertise the OpenLCB service(s) on a single interface, but need the lookup to work on all interfaces. In this case, we must stop advertising temporarily while we perform the mdns_lookup().

If the mdns_lookup() is always blocking, we can get into a situation where the caller continuously calls mdns_lookup() in a loop, and very little, or possibly no, time remains for advertising. Therefore, the EspIdfWiFi::mdns_lookup() behavior shall be changed to the following:

1. If mDNS publish on STA is enabled, always perform a blocking lookup. Do not use cache or background scanning.
2. Else if there is a valid cached result, return with that result.
3. Else if there is an active background scan for the service, return with the error EAI_AGAIN.
4. Else perform a blocking lookup.
    1. If successful, cache and return the result
    2. If unsuccessful, start background scanning for the service

Background scans shall stop after two iterations if no response has been received.

Configurable behavior, default to no background refresh: Cached results may be refreshed when they expire, and purged if there is no response after expiration.

bakerstu:
Fixes #871.

- Removes state machine logic that disables mDNS advertisement during asynchronous mDNS quires.
- Removes complex API for enabling/disabling mDNS advertising and registering/unregistering the STA interface.
- Makes mdns_lookup() non-blocking without regard to whether STA or AP interfaces are registered with mDNS.
- Unregistered STA interface when adding mDNS services for advertising if advertising on STA interface is disabled.
- Re-adds mDNS services for advertising when an IPv4 address is acquired if advertising on STA interface is enabled.
- adds DIE() call if disabling advertising on STA is invoked but not supported by the ESP configuration.
- Cleans up resulting dead code.

### DIFF
--- a/src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
+++ b/src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
@@ -33,7 +33,7 @@
  */
 
 // Uncomment following line to enable verbose logging.
-#define LOGLEVEL VERBOSE
+//#define LOGLEVEL VERBOSE
 
 #include "freertos_drivers/esp_idf/EspIdfWiFi.hxx"
 
--- a/src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
+++ b/src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
@@ -363,48 +363,17 @@ int EspIdfWiFiBase::mdns_lookup(
         return EAI_AGAIN;
     }
 
-    // The ESP-IDF mDNS implementation cannot mask registered interfaces
-    // for mDNS advertising. Therefore, in order to query on all interfaces,
-    // advertising may need to be disabled on the STA interface, if
-    // disable_mdns_publish_on_sta() is being utilized. In order to prevent
-    // a situation where advertising is masked for too long, it is important
-    // to force some dead time between queries.
-    if (mdnsLookupTimestamp_ > 0 && mdnsAdvInhibitSta_ && ipAcquiredSta_)
-    {
-        // All of the following must be true:
-        //   1. not the first time through (mdnsLookupTimestamp_ > 0)
-        //   2. mDNS advertising is inhibited on the STA interface
-        //   3. the STA has an acquired IP (meaning it could be registered with
-        //      mDNS)
-        // ...otherwise, we should not get here.
-        long long now = OSTime::get_monotonic();
-        // introduce some randomness to avoid different systems lining up in
-        // time.
-        long long expires =
-            mdnsLookupTimestamp_ + MSEC_TO_NSEC(mdnsLookupUd_(mdnsLookupRd_));
-        if (expires > now)
-        {
-            // Need to add some extra blanking time between mDNS queries.
-            usleep(NSEC_TO_USEC(expires - now));
-        }
-    }
-
-    // Not found yet, and not looking, do a "fast" blocking query.
     std::string name;
     std::string proto;
     mdns_split(service, &name, &proto);
 
-    // Enable mDNS on STA if required.
-    //bool mdns_key = mdns_adv_inhibit_on_sta_maybe_disable();
-
-    // Do the blocking query for 1.5 seconds, only 1 result.
+    // Do the blocking query for 3 seconds, only 1 result.
+    /// @todo In the future, may want to do a faster query which collects more
+    ///       than one result and then tries to asses which of multiple results
+    ///       will be the highest performance option.
     mdns_result_t *results;
     esp_err_t err =
-        mdns_query_ptr(name.c_str(), proto.c_str(), 1500, 1, &results);
-    mdnsLookupTimestamp_ = OSTime::get_monotonic();
-
-    // Disable mDNS on STA if required.
-    //mdns_adv_inhibit_on_sta_restore(mdns_key);
+        mdns_query_ptr(name.c_str(), proto.c_str(), 3000, 1, &results);
 
     // Check for an error result from the query.
     if (err != ESP_OK)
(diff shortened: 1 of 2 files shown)


bakerstu:
@balazsracz @atanisoft Hold off on reviewing this. I'm not sure the mdns_lookup() is actually working. I'm investigating.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
```diff
@@ -767,28 +603,13 @@ private:
     unsigned scanResultsEntryIndex_ {INT_MAX};
     std::string staConnectPass_; ///< last station connect attempt password
     const std::string hostname_; ///< published hostname
-    int mdnsStaLockCount_; ///< counter for recursive mDNS STA lock
     uint8_t apClientCount_; ///< number of connected wifi clients
```
atanisoft, line 606: During my testing, I've seen `WIFI_EVENT_AP_STADISCONNECTED` raised when the AP was not running as we shift over from APSTA (to support network scanning, no AP running or started) to STA only mode. We likely should explore and augment the logic around this variable when we have not seen AP running prior to receiving the event.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -340,565 +341,184 @@ void EspIdfWiFiBase::mdns_service_remove(const char *service)
...
+        // interface.
+        OSMutexLock locker(&lock_);
+        if (mdnsAdvInhibited_)
         {
-            // Only return one result.
-            struct addrinfo *addr_info =
-                (struct addrinfo*)memp_malloc(MEMP_NETDB);
-            if (addr_info == nullptr)
+            // We do not need to check for valid staIface_ or the
+            // mdnsAdvInhibitSta_ flag since we can only get here if it both
```
atanisoft, line 396: Is `mdnsAdvInhibitSta_` the correct reference here? It seems that we have two vars referenced that have similar comments on usage/dependency.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -1146,30 +765,29 @@ void EspIdfWiFiBase::ip_event_handler(
...
-            mdns_restore_sta();
-            {
-                OSMutexLock locker(&lock_);
-                if (mdnsClientCache_.size())
-                {
-                    // Trigger an immediate scan in order to refresh the cache.
-                    mdns_scanning_start_or_trigger_refresh();
-                }
-            }
             char ip_addr[16];
```
atanisoft, line 768: does this need to be increased from 16 to 17 to account for the trailing null?
or a comment to indicate why not to increase.

atanisoft (approved):

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -340,565 +341,184 @@ void EspIdfWiFiBase::mdns_service_remove(const char *service)
...
+        OSMutexLock locker(&lock_);
+        if (mdnsAdvInhibited_)
         {
-            // Only return one result.
-            struct addrinfo *addr_info =
-                (struct addrinfo*)memp_malloc(MEMP_NETDB);
-            if (addr_info == nullptr)
+            // We do not need to check for valid staIface_ or the
+            // mdnsAdvInhibitSta_ flag since we can only get here if it both
+            // were was previously valid/set before starting the query.
```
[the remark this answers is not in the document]
bakerstu: removed `was`

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
bakerstu: @atanisoft Please have another look at this logic in `wifi_event_handler()`. I've made some updates. look for `ipAcquiredAp_`, `apClientCount_`, and `apLeased_`.

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -340,565 +341,184 @@ void EspIdfWiFiBase::mdns_service_remove(const char *service)
...
+                mdns_netif_action(staIface_, MDNS_EVENT_DISABLE_IP4));
+            // (Re-)register all the services.
+            for (auto it = mdnsServices_.begin(); it != mdnsServices_.end();
+                ++it)
             {
-                return EAI_MEMORY;
+                LOG(VERBOSE, "wifi: got IP, mDNS service add, service: %s, "
+                    "proto: %s",
+                    it->name_.c_str(), it->proto_.c_str());
+                ::mdns_service_add(nullptr, it->name_.c_str(),
```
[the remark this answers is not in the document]
atanisoft: I'm also less certain there is a need for multiple mDNS instances, it can really complicate things at times partially due to how the default mDNS library is implemented as you found as well.

src/freertos_drivers/esp_idf/EspIdfWiFi.hxx
atanisoft: I don't find any references to `apLeased_` with a search, but I do find the other two so I'm guessing it may be a misreference? The event will still be raised by ESP-IDF but your code will ignore it. I likely missed it but I also don't see these vars declared

src/freertos_drivers/esp_idf/EspIdfWiFi.cxx
```diff
@@ -340,565 +342,197 @@ void EspIdfWiFiBase::mdns_service_remove(const char *service)
...
+            // We do not need to check for valid staIface_ or the
+            // mdnsAdvInhibitSta_ flag since we can only get here if it both
+            // were previously valid/set before starting the query.
+
+            // Update the timestamp for the last scan.
+            mdnsLookupTimestamp_ = OSTime::get_monotonic();
+            // Disable mDNS on STA. Note that the STA interface will also be
+            // unregistered, equivalent of calling mdns_unregister(staIface_).
+            ESP_ERROR_CHECK(
+                mdns_netif_action(staIface_, MDNS_EVENT_DISABLE_IP4));
```
bakerstu: This is potentially setting us up for a race condition, though it is unlikely to actually occur. The netif actions and mdns remove/add are guaranteed to be sequential because they get queued up, in order, for the same worker thread to execute on. However, the `mdns_netif_action()` makes a synchronous check that the interface is registered before continuing. This could setup the following sequence:

1. `mdns_netif_action(staIface_, MDNS_EVENT_DISABLE_IP4)` is queue up
2. ...return from `mdns_lookup()`
3. ...call `mdsn_look()` again...
4. `mdns_register_netif(staIface_); ESP_ERROR_CHECK(mdns_netif_action(staIface_, MDNS_EVENT_ENABLE_IP4));` is called before the mDNS worker thread can address the queue.

PR #876, opened 2025-09-07
[Do Not Merge] Traction Modem Reset Link Timeout on Any Rx

bakerstu:
I found an issue when doing firmware updates on the main decoder MCU. Basically, execution stalls whenever FLASH is erased/programmed. This causes the main decoder MCU to miss an occasional ping, and since we only reset the link timeout on a pong response, this results in a link down, eventually.

This could also be an issue for other memory space writes, such as CVs.

- Adds a callback registration to the RxInterface that is called whenever any packet is received.
- Adds callback to the the LinkManager to reset the link timeout.
  - This logic is moved from the PING_RESP handler.
- Fixes unit tests.

### DIFF
--- a/src/traction_modem/RxFlow.hxx
+++ b/src/traction_modem/RxFlow.hxx
@@ -83,9 +83,18 @@ public:
     {
         return resyncCount_;
     }
+
+    /// Register for a callback when any packet is received.
+    virtual void register_packet_rx_callback(std::function<void()> callback)
+    {
+        pktRxCallback_ = callback;
+    }
+
 protected:
     /// Track the number of times that we try to resync.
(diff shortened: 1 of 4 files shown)


src/traction_modem/RxFlow.hxx
```diff
@@ -83,9 +83,18 @@ public:
     {
         return resyncCount_;
     }
+
+    /// Register for a callback when any packet is received.
+    virtual void register_packet_rx_callback(std::function<void()> callback)
```
balazsracz, line 88: this probably should not be virtual, since the only meaningful implementation is inline here.

balazsracz (approved):

PR #908, opened 2026-02-23
Adds support for entry/exit semantics of DccDetectorWN in the protocol client class.

balazsracz:
When the high bits of the detection report are 00, that's an exit. We remove the respective locomotive from the list of locomotives in the particular block.


bakerstu (approved):

balazsracz:
This is now done in #935

PR #911, opened 2026-02-25
Adding ExclusiveConsumer feature to OpenMRN

rilull:
This is a new consumer class that is aimed at a use case to drive a signal head from a group of outputs on a node. This matches a typical prototype approach of only having a single color or position being lit (turned on) inside a single signal head.

The new class is called ConfiguredExclusiveConsumer, matching the naming scheme of existing producer/consumers. This class enables creating a group of outputs (from 2 to N) that, upon receipt of a matching event ID by the node, will turn on that particular output and turn off all other outputs in the same group.

There is sample usage of this feature in the Nucleo DevKit application (nucleo_io) on Port D and E. I have put various compile time configuration guards in place to match what we have previously done for the Snap Switch and IO expansion features.

I have tested this on DevKit hardware with a F303 Nucleo.


src/openlcb/ConfiguredExclusiveConsumer.hxx
```diff
@@ -0,0 +1,242 @@
...
+/// };
+/// openlcb::ConfiguredExclusiveConsumer signal_consumer(stack.node(),
+///    kSignalGpio, ARRAYSIZE(kSignalGpio),
+///    cfg.seg().signal_consumers());
+/// ```
+class ConfiguredExclusiveConsumer : public ConfigUpdateListener,
+                                    private SimpleEventHandler
+{
+public:
+    typedef ExclusiveConsumerConfig config_entry_type;
```
bakerstu: C++ types should use PascalCase.

bakerstu (approved):
If you address the comments you can merge.

PR #924, opened 2026-04-16
Style Guide Markdown

bakerstu:
Moves the style guide to markdown format so that it is viewable onGitHub. Doxygen is aware of this and also knows how to pull it in.

The work was mostly done by Claude Code.


CLAUDE.md
```diff
@@ -0,0 +1,223 @@
...
+make llvm-tests
+
+# Run tests on JavaScript/Emscripten target
+make js-tests
+
+# Run all tests (native + LLVM)
+make alltests
+
+# Run tests for a specific module (e.g., openlcb)
+make -C targets/test/openlcb tests
```
balazsracz, line 40: does this actually work? I don't think so.

CLAUDE.md
```diff
@@ -0,0 +1,223 @@
...
+
+# Run a single test file
+make -C targets/test tests TESTBLACKLIST="utils/SocketClient.cxxtest"
```
balazsracz, line 43: I don't think this is correct.
It would be something like make -C targets/test openlcb/FooBar.testout

README.md
```diff
@@ -12,6 +12,15 @@ creating standards that, once approved by the NMRA, become LCC standards. The
 [OpenLCB & LCC Standards](http://openlcb.org/openlcb-and-lcc-documents/layout-command-control-lcc/)
 page will give you a good idea of status of the specifications.
 
+## Table of Contents
+
+- [Why OpenMRN](#why-openmrn)
+- [Getting Started](#getting-started)
```
balazsracz, line 18: is there a way to autogetnerate these table of contents? I expect it will get out of date rather soon

balazsracz (approved):
i didn't review the file that was deleted and completely rewritten because there is no side by side diff visible.
The claude.md might be better in a different PR?

CLAUDE.md
bakerstu: Removed. It is not all that useful anyways.

CLAUDE.md
bakerstu: Your example does work. I've made the update.

README.md
bakerstu: The issue I ran into was finding a solution that works for both Doxygen and the GitHub web interface. This seems to work on both.

PR #926, opened 2026-04-25
Fix ESP32 pull-down restriction for valid GPIOs

JohnSL:
## Summary
Fix the plain ESP32 `GpioInputPD` restriction so valid GPIOs are allowed again.

The previous static assertion used `PIN_NUM == 15` where it should have used `PIN_NUM != 15`. As a result, it both failed to reject GPIO 15 and incorrectly rejected almost every other valid plain-ESP32 GPIO when `GpioInputPD` was instantiated.

## Change
This updates the plain ESP32 pull-down assertion in `Esp32Gpio.hxx` so GPIO 0, 5, and 15 remain rejected for `GpioInputPD`, while other valid GPIOs are no longer blocked by the bad condition.

## Validation
The edited header reports no editor errors after the change. I also reviewed existing repo coverage and did not find an automated test that instantiates this exact ESP32 `GpioInputPD` case.

## Notes
This is a compile-time guard fix only. It restores intended template usage and does not change runtime behavior for already-valid builds.

### DIFF
--- a/src/freertos_drivers/esp32/Esp32Gpio.hxx
+++ b/src/freertos_drivers/esp32/Esp32Gpio.hxx
@@ -334,7 +334,7 @@ public:
                   "enabling pull-up is not possible.");
     // GPIO 0, 5 and 15 typically have pull-up resistors.
     static_assert(!PDEN ||
-                  (PDEN && (PIN_NUM != 0 && PIN_NUM != 5 && PIN_NUM == 15)),
+                  (PDEN && (PIN_NUM != 0 && PIN_NUM != 5 && PIN_NUM != 15)),
                   "GPIO 0, 5, 15 typically have built-in pull-up resistors, "
                   "enabling pull-down is not possible.");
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/esp32/Esp32Gpio.hxx
```diff
@@ -334,7 +334,7 @@ public:
                   "enabling pull-up is not possible.");
     // GPIO 0, 5 and 15 typically have pull-up resistors.
     static_assert(!PDEN ||
-                  (PDEN && (PIN_NUM != 0 && PIN_NUM != 5 && PIN_NUM == 15)),
+                  (PDEN && (PIN_NUM != 0 && PIN_NUM != 5 && PIN_NUM != 15)),
```
atanisoft, line 337: This looks good but only applies to the esp32 target only. We should likely protect it accordingly.

JohnSL: Done.

PR #927, opened 2026-04-26
Fix handling of CAN frames with priority bit zero

### LINKED ISSUE #715: Should be ignoring high bit of CAN frames in MTI decode
OpenMRN is apparently checking the top 0x1,0000,0000 bit of CAN frames for a 1.  That bit is “send as 1, don’t check on receipt” so the zero should be valid.

For example the Fast Clock replies to a VerifyNodeAddressed when a 1 is there and doesn’t with a zero there. 

See Frame Standard section 4 and TN section 2.4.

balazsracz:
Fixes #715.

Updates CAN frame registration masks and various other checks for the topmost bit of the CAN identifier.

The standard says that this bit (the priority bit) should be "send as 1, ignore upon receipt". Previously the code was checking at receipt, which is incorrect.

### DIFF
--- a/src/openlcb/Bootloader.cxxtest
+++ b/src/openlcb/Bootloader.cxxtest
@@ -606,6 +606,19 @@ TEST_F(BootloaderTest, VerifyGlobal)
     exit_bootloader();
 }
 
+TEST_F(BootloaderTest, PriorityBitZeroBootloader)
+{
+    proper_startup();
+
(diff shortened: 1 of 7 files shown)


balazsracz:
This is actually a union of changes by different models. Some changes were missed by JSL's prompt, whereas others were missed by mine.

bakerstu (approved):

PR #928, opened 2026-04-26
Fixes compile errors from armgcc 15.

balazsracz:
- there were some missing includes.
- the Device infrastructure generates a new warning due to having both static ::open and virtual ::open functions on the same classes in the hierarchy. This warning needs to be disabled at this point.
- there was a legit bug discovered in PCA8685PWM that impacts only big endian processors.


bakerstu (approved):
Do we have an issue filed for the -Wno-overloaded-virtual option? If not, let's file one. If yes, let's link it to this pull request, but not close it.

PR #935, opened 2026-07-12
Fixes handling of loco exit messages

balazsracz:
We made a change in the draft spec that designated the top three bits as 00=exit, 01=entry direction A, 10=entry direction B, 11=entry direction unknown.
This PR adjusts the handling of the exit messages according to the new draft. Previusly these were misinterpreted as entry messages which made the client class divergent from the server class.


bakerstu (approved):
Just a general comment: Should we use a const expression for the masking operations rather than magic number(s)?

PR #937, opened 2026-07-23
Makes Factory reset perform factory reset on all eventID variables.

balazsracz:
- Adds a helper in the SimpleStack that calls factory_reset_all_events when factory_reset() is invoked by the config update flow.
- Changes the config update flow to be FIFO in its member addition mechanism. Since it's a single linked list, we have to keep a last_ pointer around for this purpose.
- Adds a constructor argument to configupdateflow that registers listeners.

These all together ensure that events gets factory reset first, then configured objects get their factory reset method called. This means that a configured object can override what any given event ID would be after factory reset.

### DIFF
--- a/src/openlcb/ConfigUpdateFlow.cxx
+++ b/src/openlcb/ConfigUpdateFlow.cxx
@@ -69,7 +69,8 @@ void ConfigUpdateFlow::factory_reset()
 void ConfigUpdateFlow::register_update_listener(ConfigUpdateListener *listener)
 {
     AtomicHolder h(this);
-    pendingListeners_.push_front(listener);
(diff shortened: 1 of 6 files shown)


bakerstu (approved):
If I understand correctly, I think the point of this is in part to ensure the new "helper" is always at the beginning of the list, and subsequently runs first. We could be a bit more verbose in the commentary around this intent.

balazsracz:
> If I understand correctly, I think the point of this is in part to ensure the new "helper" is always at the beginning of the list, and subsequently runs first. We could be a bit more verbose in the commentary around this intent.

done. This is also called out explicitly in the PR description, but I added several comments in the code to this effect now.

PR #938, opened 2026-07-24
Adds a helper function on dcc packet to generate a NOOP.

balazsracz:
A NOOP is a long-address=10239 NOP packet: 0xE7 0xFF 0x00 XOR. This falls into the mobile decoder address range.

### DIFF
--- a/src/dcc/Packet.cxx
+++ b/src/dcc/Packet.cxx
@@ -78,6 +78,12 @@ void Packet::add_dcc_checksum()
     packet_header.skip_ec = 1;
 }
 
+void Packet::add_dcc_noop()
+{
+    payload[dlc++] = 0;
+    add_dcc_checksum();
+}
+
 void Packet::set_dcc_idle()
 {
     start_dcc_packet();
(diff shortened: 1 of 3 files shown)


src/dcc/Packet.cxx
```diff
@@ -87,6 +87,13 @@ void Packet::set_dcc_idle()
     packet_header.skip_ec = 1;
 }
 
+void Packet::set_dcc_noop()
+{
+    add_dcc_address(DccLongAddress(10239));
```
bakerstu, line 98: I think we should paramaterize the address for the caller to pass in. We might have future use cases that may want to send NOPs to other addresses.

bakerstu (approved):

src/dcc/Packet.cxx
balazsracz: I refactored to an add_dcc_noop function. If you want a noop for a custom address, you call add_dcc_address() then add_dcc_nop(). Tihs is how we work with other packet types as well.

PR #941, opened 2026-07-30
Traction Modem State Transfer

bakerstu:
## Summary

Adds support for the traction modem to query and receive decoder state (velocity and function status) instead of only pushing commands, and wires an initial state sync on link-up.

- Naturalize the on/off `OutputState` encoding (`OFF = 0x0000`, `ON = 0xFFFF`) to match the wire protocol.
- Add `VelocityStatus` and `FunctionStatus` packet handlers that send `CMD_SPEED_QUERY` / `CMD_FN_QUERY` and parse the corresponding responses, wired into `ModemTrain` alongside the existing `Output` handler.
- Extend `ModemTrainHwInterface` with `velocity_status()`, `function_status()`, `get_fn()`, `legacy_address()`, `legacy_address_type()`, and `on_link_up()` so hardware implementations can respond to modem-initiated queries and report real values instead of `/// @todo` stubs.
- Trigger a velocity status query on link-up so initial train state is synchronized when the modem comes online; expose `ModemTrain::get_function_status()` so function-status queries on link-up can be driven externally instead of from within `ModemTrain`.

### DIFF
--- a/src/traction_modem/Defs.hxx
+++ b/src/traction_modem/Defs.hxx
@@ -81,7 +81,7 @@ struct Defs
         CMD_SPEED_SET          = 0x0100, ///< set velocity
         CMD_FN_SET             = 0x0101, ///< set function
         CMD_ESTOP_SET          = 0x0102, ///< emergency stop request
-        CMD_SPEED_QUERY        = 0x0110, ///< query current speed
+        CMD_SPEED_QUERY        = 0x0110, ///< query current velocity
         CMD_FN_QUERY           = 0x0111, ///< query function status
         CMD_DC_DCC_PRESENT     = 0x0200, ///< DC/DCC present
         CMD_WIRELESS_PRESENT   = 0x0201, ///< wireless present
@@ -104,7 +104,7 @@ struct Defs
         RESP_FN_SET             = RESPONSE | CMD_FN_SET,
         /// emergency stop response
         RESP_ESTOP_SET          = RESPONSE | CMD_ESTOP_SET,
-        /// query current speed response
+        /// query current velocity response
         RESP_SPEED_QUERY        = RESPONSE | CMD_SPEED_QUERY,
         /// query function status response
         RESP_FN_QUERY           = RESPONSE | CMD_FN_QUERY,
@@ -158,8 +158,16 @@ struct Defs
     static constexpr unsigned LEN_BAUD_RATE_REQUEST = 2;
     /// Length of the data payload of a set function packet.
     static constexpr unsigned LEN_FN_SET = 6;
+    /// Length of the data payload of a function query packet.
+    static constexpr unsigned LEN_FN_QUERY = 4;
+    /// Length of the data payload of a function query response packet.
+    static constexpr unsigned LEN_FN_QUERY_RESP = 6;
     /// Length of the data payload of a set speed packet.
     static constexpr unsigned LEN_SPEED_SET = 1;
+    /// Length of the data payload of a speed query packet.
+    static constexpr unsigned LEN_SPEED_QUERY = 0;
+    /// Length of the data payload of a speed query response packet.
+    static constexpr unsigned LEN_SPEED_QUERY_RESP = 4;
     /// Length of the data payload of a set estop packet.
     static constexpr unsigned LEN_ESTOP_SET = 0;
     /// Length of the data payload of a wireless present packet.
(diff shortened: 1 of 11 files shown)


balazsracz:
looks like the tests are failing

src/traction_modem/Defs.hxx
```diff
@@ -281,6 +303,36 @@ struct Defs
...
+    static_assert(
+        std::is_standard_layout<FunctionStatusQueryResponse>::value == true);
+
+    /// Structure of a speed query response.
+    struct SpeedQueryResponse
+    {
+        Header header_; ///< packet header
+        uint16_t error_; ///< error code
+        uint8_t targetSpeed_; ///< commanded speed (DCC 128-step encoding)
+        uint8_t currentSpeed_; ///< current speed (DCC 128-step encoding)
```
balazsracz, line 327: we usually have three speed values:
- last received speed command (target speed)
- what the decoder is currently driving (e.g. due to acceleration, decelaration speed)
- that is the measured speed (feedback, e.g. backEMF)

src/traction_modem/FunctionStatus.cxx
```diff
@@ -0,0 +1,90 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * @file FunctionStatus.cxx
+ *
+ * Logic for querying function status.
+ *
+ * @author Stuart Baker
+ * @date 27 June 2025
```
balazsracz, line 32: is this correct?

src/traction_modem/Defs.hxx
```diff
@@ -389,6 +441,19 @@ struct Defs
         return p;
     }
 
+    /// Computes payload to query a function.
+    /// @param fn function number
+    /// @return wire formatted payload
+    static Payload get_fn_query_payload(unsigned fn)
```
balazsracz, line 447: we wanted to add the number of functions to query in this message (to be able to query fn 0 to 68 in one message)

src/traction_modem/FunctionStatus.cxx
```diff
@@ -0,0 +1,90 @@
...
+void FunctionStatus::send(Buffer<Message> *buf, unsigned prio)
+{
+    auto b = get_buffer_deleter(buf);
+    switch (b->data()->command())
+    {
+        case Defs::RESP_FN_QUERY:
+        {
+            Defs::FunctionStatusQueryResponse *fsqr =
+                (Defs::FunctionStatusQueryResponse *)b->data()->payload.data();
+            hwIf_->function_status(be32toh(fsqr->fn_), be16toh(fsqr->value_));
```
balazsracz, line 84: we would want this to be repeated as well

src/traction_modem/ModemTrainHwInterface.hxx
```diff
@@ -76,6 +77,48 @@ public:
         OUT_OF_BOUNDS     = openlcb::MemoryConfigDefs::ERROR_OUT_OF_BOUNDS,
     };
 
+    /// Called when link transitions to "up" state.
+    virtual void on_link_up()
+    {
+    }
+
+    /// Report the current velocity in response to a query.
```
balazsracz: I don't think this correctly describes what this function is for.

afaik this function is called when a speed status response arrives from the decoder via the modeminterface. The arguments are the speed parameters coming from the decoder

Similar for some of the next functions below.

src/traction_modem/FunctionStatus.hxx
```diff
@@ -0,0 +1,79 @@
...
+
+    /// Send a function status query to the decoder.
+    /// @param fn function number to query
+    void query(unsigned fn);
+
+private:
+    /// Receive handler for function status response messages.
+    /// @param buf incoming message
+    /// @param prio message priority
+    void send(Buffer<Message> *buf, unsigned prio) override;
```
balazsracz, line 70: It seems pretty heavyweight to add one new such class for each message that is defined to pass through the modem link. There is no actual business logic in these classes and I am not sure that we are planning to add any business logic. This is also visible on the lack of unit tests for these new classes.
For this reason I think it might make more sense to have these trivial bridging handlers to be consolidated into a single class for trivial message handlers (possibly in ModemTrain directly)

bakerstu: I'm not sure that I agree. I need to think more about this.

bakerstu: This might evolve a bit based on our discussion of rearchitecting to use Memory Config.

bakerstu:
> looks like the tests are failing

Tests pass now.

src/traction_modem/FunctionStatus.cxx
bakerstu: Fixed.

src/traction_modem/ModemTrainHwInterface.hxx
bakerstu: Added additional commentary.

src/traction_modem/Defs.hxx
bakerstu: I understand. I will enhance this on a separate pull request.

src/traction_modem/Defs.hxx
bakerstu: I will address this on a separate pull request.

src/traction_modem/FunctionStatus.cxx
bakerstu: I will address this on a separate pull request.

PR #944, opened 2026-08-08
Add Esp32RmtPwm, an RMT-based PWM driver

bakerstu:
## Summary
- Adds `src/freertos_drivers/esp32/Esp32RmtPwm.hxx`, a header-only driver implementing the abstract `PWM` interface (`PWM.hxx`) on top of the ESP32 RMT peripheral, as an alternative to the existing LEDC-based `Esp32Ledc`.
- Each output pin gets its own independent RMT TX channel (no shared timer, unlike LEDC), supporting up to 20kHz at 8+ bit resolution by default (configurable up to 15-bit resolution).
- A continuous waveform is generated by looping a single RMT symbol per period (`loop_count = -1`) so the hardware repeats it with no CPU/interrupt involvement once started.
- 0% and 100% duty cycles are exact (not approximated), using a same-level two-half-period symbol to work around the RMT hardware's "zero duration = end of transmission" quirk.
- Changing duty or frequency briefly disables/re-enables the affected channel only (per the ESP-IDF driver's documented mechanism for interrupting an infinite-loop transmission), causing a sub-millisecond glitch on that one output — acceptable for LED dimming.
- The constructor asserts pin count against `SOC_RMT_TX_CANDIDATES_PER_GROUP`, which correctly reflects each SoC's TX-capable channel budget (e.g. only 2 on ESP32-C3, vs. 6-8 LEDC channels on the same hardware).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/freertos_drivers/esp32/Esp32RmtPwm.hxx
+++ b/src/freertos_drivers/esp32/Esp32RmtPwm.hxx
@@ -1,0 +1,414 @@
+/** @copyright
+ * Copyright (c) 2026, Stuart Baker
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are  permitted provided that the following conditions are met:
+ *
+ *  - Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ *  - Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * @file Esp32RmtPwm.hxx
+ *
+ * ESP-IDF RMT adapter that exposes a PWM interface.
+ *
+ * @author Stuart Baker
+ * @date 8 August 2026
+ */
+
+#ifndef _DRIVERS_ESP32RMTPWM_HXX_
+#define _DRIVERS_ESP32RMTPWM_HXX_
+
+#include "freertos_drivers/common/PWM.hxx"
+#include "utils/logging.h"
+#include "utils/macros.h"
+#include "utils/Uninitialized.hxx"
+
+#include <driver/gpio.h>
+#include <driver/rmt_encoder.h>
+#include <driver/rmt_tx.h>
+#include <soc/soc_caps.h>
+
+#include <vector>
+
+namespace openmrn_arduino
+{
+
+/// ESP32 RMT provider for PWM like output on GPIO pins.
+///
+/// Unlike @ref Esp32Ledc, which multiplexes several outputs onto one shared
+/// LEDC timer, each pin passed to a @ref Esp32RmtPwm instance is assigned
+/// its own, fully independent RMT TX channel. There is no shared-timer or
+/// "first channel" concept, and each output's frequency can be changed
+/// independently of the others.
+///
+/// The waveform for one PWM period is expressed as a single RMT symbol and
+/// transmitted with an infinite loop count, so the hardware repeats it
+/// forever with no CPU or interrupt involvement once started. Because the
+/// RMT hardware treats a symbol half with a duration of literal zero as an
+/// end-of-transmission marker, a 0% or 100% duty cycle is represented as two
+/// consecutive halves at the same (constant) level rather than a single
+/// zero-duration half -- this still yields an exact, glitch-free 0% or 100%
+/// output, not an approximation.
+///
+/// Changing the duty cycle or the frequency of a channel requires briefly
+/// disabling and re-enabling that channel in order to interrupt its
+/// in-progress infinite-loop transmission before a new one can be queued
+/// (this is the mechanism documented by the ESP-IDF RMT driver itself for
+/// terminating a looped transmission). This causes a brief, sub-millisecond
+/// glitch on the affected output only -- acceptable for LED dimming and
+/// similar uses.
+///
+/// The number of RMT channels available is limited and varies by ESP32
+/// variant (as few as two usable TX channels on the ESP32-C3), and is
+/// considerably smaller than the number of LEDC channels available on the
+/// same hardware. The constructor will HASSERT if more pins are requested
+/// than SOC_RMT_TX_CANDIDATES_PER_GROUP allows for the target SoC.
+///
+/// Example of usage:
+///```
+/// OpenMRN openmrn(NODE_ID);
+/// Esp32RmtPwm rmt_pwm({16,    // pin gets its own RMT channel
+///                      17,
+///                      18});
+/// ServoConsumer servo_0(openmrn.stack()->node(), cfg.seg().servo(), 1000,
+///                       rmt_pwm.get_channel(1));
+/// PWMGPO led_0(rmt_pwm.get_channel(0), 2500, 0);
+/// void setup() {
+///   ...
+///   rmt_pwm.hw_init();
+///   openmrn.begin();
+///   openmrn.stack()->set_tx_activity_led(&led_0);
+///   ...
+/// }
+///```
+class Esp32RmtPwm
+{
+public:
+    /// Constructor using std::vector.
+    ///
+    /// @param pins is the collection of output pins to use for this
+    /// instance; each pin is assigned its own independent RMT TX channel.
+    /// @param resolution_bits is the PWM duty cycle resolution in bits,
+    /// range 1 to 15, default is 8 (256 duty steps).
+    /// @param freq_hz is the initial PWM frequency in Hz for all channels,
+    /// default is 20kHz.
+    /// @param clk_src is the RMT peripheral clock source, default is
+    /// RMT_CLK_SRC_DEFAULT. Per ESP-IDF, channels that belong to the same
+    /// RMT group must use the same clock source.
+    /// @param mem_block_symbols is the number of RMT hardware memory block
+    /// symbols to reserve per channel, default is
+    /// SOC_RMT_MEM_WORDS_PER_CHANNEL (exactly one hardware memory block per
+    /// channel; only a single symbol is ever needed by this driver, so the
+    /// default is already conservative).
+    /// @param trans_queue_depth is the RMT transmit queue depth, default is
+    /// 1, since only one (looped) transmission is ever in flight per
+    /// channel.
+    Esp32RmtPwm(const std::vector<uint8_t> &pins,
+                uint8_t resolution_bits = 8, uint32_t freq_hz = 20000,
+                rmt_clock_source_t clk_src = RMT_CLK_SRC_DEFAULT,
+                size_t mem_block_symbols = SOC_RMT_MEM_WORDS_PER_CHANNEL,
+                size_t trans_queue_depth = 1)
+        : pins_(pins)
+        , resolutionBits_(resolution_bits)
+        , freqHz_(freq_hz)
+        , clkSrc_(clk_src)
+        , memBlockSymbols_(mem_block_symbols)
+        , transQueueDepth_(trans_queue_depth)
+    {
+        init_helper();
+    }
+
+    /// Constructor using std::initializer_list.
+    ///
+    /// @param pins is the collection of output pins to use for this
+    /// instance; each pin is assigned its own independent RMT TX channel.
+    /// @param resolution_bits is the PWM duty cycle resolution in bits,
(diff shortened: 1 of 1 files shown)


src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,414 @@
...
+#include "utils/Uninitialized.hxx"
+
+#include <driver/gpio.h>
+#include <driver/rmt_encoder.h>
+#include <driver/rmt_tx.h>
+#include <soc/soc_caps.h>
+
+#include <vector>
+
+namespace openmrn_arduino
```
bakerstu, line 50: This namespace was copied over form the Esp32Ledc.hxx that I used as a reference. I don't know if/why it is needed. Thoughts?

balazsracz: I don't think we should use this namespace. It is only for stuff that is in the arduino/ folder, like interface classes for serial and such drivers.

atanisoft: Agreed, perhaps move it over to similar to how the esp-idf WiFi code is including namespace?

atanisoft:
We should also clarify, esp32-s3 (and other newer chips!) will have a maximum number of tx available, it is 4 max on esp32-s3, and similar for memory blocks. Ideally you should never need more than one memory block.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+{
+
+/// ESP32 RMT provider for PWM like output on GPIO pins.
+///
+/// Unlike @ref Esp32Ledc, which multiplexes several outputs onto one shared
+/// LEDC timer, each pin passed to a @ref Esp32RmtPwm instance is assigned
+/// its own, fully independent RMT TX channel. There is no shared-timer or
```
atanisoft, line 57: channels are *NOT* fully independent, all channels RX/TX share a single ISR which is used for refilling the TX channel or informing RX callbacks. The timer *IS* shared still but it is slightly more flexible than the LEDC timers.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+/// independently of the others.
+///
+/// The waveform for one PWM period is expressed as a single RMT symbol and
+/// transmitted with an infinite loop count, so the hardware repeats it
+/// forever with no CPU or interrupt involvement once started. Because the
+/// RMT hardware treats a symbol half with a duration of literal zero as an
+/// end-of-transmission marker, a 0% or 100% duty cycle is represented as two
+/// consecutive halves at the same (constant) level rather than a single
+/// zero-duration half -- this still yields an exact, glitch-free 0% or 100%
+/// output, not an approximation.
```
atanisoft, line 68: It will be an approximation, usually 0-1uS variance in practice.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+/// disabling and re-enabling that channel in order to interrupt its
+/// in-progress infinite-loop transmission before a new one can be queued
+/// (this is the mechanism documented by the ESP-IDF RMT driver itself for
+/// terminating a looped transmission). This causes a brief, sub-millisecond
+/// glitch on the affected output only -- acceptable for LED dimming and
+/// similar uses.
+///
+/// The number of usable RMT TX channels is limited and varies by ESP32
+/// variant -- e.g. as few as two on the ESP32-C3, four on the ESP32-S3, up
+/// to eight on the original ESP32 -- and on every variant is considerably
```
atanisoft, line 80: esp32 and esp32-s2 are the only variances that allow up to eight TX/RX/etc (any combination) channels. All others are generally limited to four or fewer and `SOC_RMT_TX_CANDIDATES_PER_GROUP` should work for detection of TX channels. Do note that some of these constants were renamed in ESP-IDF v6, this specific one has been dropped entirely in ESP-IDF v6.0.0 and will only show up as a compilation failure if you exceed the limits or at runtime with calling `rmt_new_tx_channel` not returning ESP_OK.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+/// smaller than the number of LEDC channels available on the same
+/// hardware. The constructor will HASSERT if more pins are requested than
+/// SOC_RMT_TX_CANDIDATES_PER_GROUP allows for the target SoC.
+///
+/// RMT hardware memory block symbols are likewise a shared, limited
+/// per-SoC resource: all of a chip's RMT channels draw from a common pool
+/// of SOC_RMT_CHANNELS_PER_GROUP * SOC_RMT_MEM_WORDS_PER_CHANNEL words, so
```
atanisoft, line 87: SOC_RMT_CHANNELS_PER_GROUP has been dropped in ESP-IDF v6.0.0, the other constant still exists.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+    /// symbols to reserve per channel, default is
+    /// SOC_RMT_MEM_WORDS_PER_CHANNEL (exactly one hardware memory block per
+    /// channel; only a single symbol is ever needed by this driver, so the
+    /// default is already conservative -- increasing it borrows memory
+    /// blocks away from, and so reduces the number of, other channels).
+    /// @param trans_queue_depth is the RMT transmit queue depth, default is
+    /// 1, since only one (looped) transmission is ever in flight per
+    /// channel.
+    Esp32RmtPwm(const std::vector<uint8_t> &pins,
+                uint8_t resolution_bits = 8, uint32_t freq_hz = 20000,
```
atanisoft, line 136: Would it make sends to move `20000` into a constant? Similar for resolution_bits.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+/// needs a single symbol per channel, so mem_block_symbols defaults to
+/// exactly one hardware memory block (SOC_RMT_MEM_WORDS_PER_CHANNEL) and
+/// should not normally be increased.
+///
+/// Example of usage:
+///```
+/// OpenMRN openmrn(NODE_ID);
+/// Esp32RmtPwm rmt_pwm({16,    // pin gets its own RMT channel
+///                      17,
+///                      18});
```
atanisoft, line 100: Is it possible to pass in `Esp32RmtPwm rmt_pwm({GPIO_NUM_16, GPIO_NUM_17, GPIO_NUM18});` ?

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+            LOG(VERBOSE,
+                "[Esp32RmtPwm] Configuring RMT channel on GPIO %d "
+                "(resolution:%d, frequency:%" PRIu32 ")",
+                pin, 1 << resolutionBits_, freqHz_);
+
+            rmt_tx_channel_config_t config;
+            memset(&config, 0, sizeof(rmt_tx_channel_config_t));
+            config.gpio_num = static_cast<gpio_num_t>(pin);
+            config.clk_src = clkSrc_;
+            config.resolution_hz = freqHz_ * (1u << resolutionBits_);
```
atanisoft, line 203: If we assume freqHz_ is 20k and resolutionBits_ as 8, will this end up being 20k*256? I've never set it beyond a specific hertz rate ever.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+
+            rmt_copy_encoder_config_t encoder_config;
+            memset(&encoder_config, 0, sizeof(rmt_copy_encoder_config_t));
+            rmt_encoder_handle_t encoder = nullptr;
+            ESP_ERROR_CHECK(rmt_new_copy_encoder(&encoder_config, &encoder));
+
+            ESP_ERROR_CHECK(rmt_enable(channel));
+
+            channels_[i].emplace(channel, encoder,
+                static_cast<gpio_num_t>(pin), resolutionBits_, freqHz_,
```
atanisoft, line 219: instead of static_cast, use `config.gpio_num` instead to remove the need to re-cast it.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+                symbol.level1 = 0;
+                symbol.duration1 = periodTicks_ - dutyTicks_;
+            }
+
+            ESP_ERROR_CHECK(rmt_disable(channel_));
+            ESP_ERROR_CHECK(rmt_enable(channel_));
+
+            rmt_transmit_config_t tx_config;
+            memset(&tx_config, 0, sizeof(rmt_transmit_config_t));
+            tx_config.loop_count = -1;
```
atanisoft, line 412: It may be beneficial to also set `tx_config.flags.queue_nonblocking = 1;` which will prevent a  blocking call with `rmt_transmit`. Without it being set it will block 100% of the time due to having a `trans_queue_depth` of one when configuring the TX channel. 

`tx_config` could also be statically declared and used for any channel as there are zero parameters that would be specific to the channel. I have used this declaration in a few places:
```
static constexpr const rmt_transmit_config_t transmit_config = {};
```
This initializes it as zero for any member variables.

atanisoft (approved):

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+public:
+    /// Constructor using std::vector.
+    ///
+    /// @param pins is the collection of output pins to use for this
+    /// instance; each pin is assigned its own independent RMT TX channel.
+    /// @param resolution_bits is the PWM duty cycle resolution in bits,
+    /// range 1 to 15, default is 8 (256 duty steps).
+    /// @param freq_hz is the initial PWM frequency in Hz for all channels,
+    /// default is 20kHz.
+    /// @param clk_src is the RMT peripheral clock source, default is
```
balazsracz, line 123: why does the client ever need to change this?

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+    /// RMT_CLK_SRC_DEFAULT. Per ESP-IDF, channels that belong to the same
+    /// RMT group must use the same clock source.
+    /// @param mem_block_symbols is the number of RMT hardware memory block
```
balazsracz, line 126: why does the client ever need to change this?
Why aren't we reserving exactly one word per channel?

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+    /// symbols to reserve per channel, default is
+    /// SOC_RMT_MEM_WORDS_PER_CHANNEL (exactly one hardware memory block per
+    /// channel; only a single symbol is ever needed by this driver, so the
+    /// default is already conservative -- increasing it borrows memory
+    /// blocks away from, and so reduces the number of, other channels).
+    /// @param trans_queue_depth is the RMT transmit queue depth, default is
```
balazsracz, line 132: why does the client ever need to change this?
It seems the answer is permanently 1.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+        /// @param mem_block_symbols is the RMT hardware memory block symbol
+        /// count that channel was configured with; retained for the same
+        /// reason as clk_src.
+        /// @param trans_queue_depth is the RMT transmit queue depth that
+        /// channel was configured with; retained for the same reason as
+        /// clk_src.
+        Channel(rmt_channel_handle_t channel, rmt_encoder_handle_t encoder,
+                gpio_num_t pin, uint8_t resolution_bits, uint32_t freq_hz,
+                rmt_clock_source_t clk_src, size_t mem_block_symbols,
+                size_t trans_queue_depth)
```
balazsracz, line 280: we are storing a lot of variables here on a per channel basis even though there is not a very good reason to do so. First of all, most of these should be const; second, even if we have valid use-cases for the user changing these values, we should refer to the parent class for theri values.
For example I don't think that you can independently set the frequency for different channels.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+            , clkSrc_(clk_src)
+            , memBlockSymbols_(mem_block_symbols)
+            , transQueueDepth_(trans_queue_depth)
+        {
+            apply_duty();
+        }
+
+        /// Sets the PWM frequency for this channel only. Like @ref
+        /// Esp32Ledc, this repurposes PWM::set_period() to mean the
+        /// channel's frequency in Hz rather than a literal duty-cycle-unit
```
balazsracz, line 297: why do we repurpose a PWM API to mean something else?
That does not seem like a good idea. It seems like this would break the contract of the APIs and how clients would use it.

This also means that if you have a client using this class in a way that works, then it would not be compatible with any other PWM* peripheral.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+        /// channel was configured with; retained for the same reason as
+        /// clk_src.
+        Channel(rmt_channel_handle_t channel, rmt_encoder_handle_t encoder,
+                gpio_num_t pin, uint8_t resolution_bits, uint32_t freq_hz,
+                rmt_clock_source_t clk_src, size_t mem_block_symbols,
+                size_t trans_queue_depth)
+            : channel_(channel)
+            , encoder_(encoder)
+            , pin_(pin)
+            , resolutionBits_(resolution_bits)
```
balazsracz, line 284: please delete all of these variables that are not used.

a lot of this code here seems like the author never heard of the YAGNI principle.

src/freertos_drivers/esp32/Esp32RmtPwm.hxx
```diff
@@ -0,0 +1,479 @@
...
+                pin_, counts);
+
+            ESP_ERROR_CHECK(rmt_disable(channel_));
+            ESP_ERROR_CHECK(rmt_del_channel(channel_));
+
+            rmt_tx_channel_config_t config;
+            memset(&config, 0, sizeof(rmt_tx_channel_config_t));
+            config.gpio_num = pin_;
+            config.clk_src = clkSrc_;
+            config.resolution_hz = counts * periodTicks_;
```
balazsracz, line 321: what is the use-case for changing the Hz on the fly?

I know of no other PWM implementation that changes the Hz. It is usually fixed based on the clock of the MCU.

balazsracz (changes requested):
Sorry, claude, but I think this code is broken and generally poor quality.

PR #949, opened 2026-08-30
Fix a potential race condition in the cc32xx uart interrupt handler.

balazsracz:
- when there is a (possibly nested) interrupt within the send() function, we could possibly have cleared a future TX EOT interrupt. This would make us end up in a situation where we have txPending==true but a TX interrupt will never come.
- makes sure that !UARTBusy() is treated the same as an EOT interrupt.

### DIFF
--- a/src/freertos_drivers/ti/CC32xxUart.cxx
+++ b/src/freertos_drivers/ti/CC32xxUart.cxx
@@ -259,6 +259,12 @@ void CC32xxUart::send()
         /* no more data left to send */
         MAP_UARTTxIntModeSet(base_, UART_TXINT_MODE_EOT);
         MAP_UARTIntClear(base_, UART_INT_TX);
+        if (!MAP_UARTBusy(base_))
+        {
+            // We lost a race and might have removed the expected TX interrupt.
+            // This will ensure that we enter the irq handler again.
+            MAP_IntPendSet(interrupt_);
+        }
     }
 }
 
@@ -329,7 +335,7 @@ void CC32xxUart::interrupt_handler()
         }
     }
     /* transmit a character if we have pending tx data */
-    if (txPending_ && (status & UART_INT_TX))
+    if (txPending_ && ((status & UART_INT_TX) || (!MAP_UARTBusy(base_))))
     {
         if (txBuf->pending())
         {
--- a/src/freertos_drivers/ti/CC32xxUart.cxx
+++ b/src/freertos_drivers/ti/CC32xxUart.cxx
@@ -234,6 +234,22 @@ void CC32xxUart::set_mode()
  */
 void CC32xxUart::send()
 {
+    /// @todo: We observed situations where the TX interrupt got lost, causing
+    /// a deadlock of the UART driver, with the txBuf being full, txPending ==
+    /// true, select being not write-active, the hardware TX fifo being empty,
+    /// and this situation persisting for an extended period of time.
+    ///
+    /// The tx_char and send function is being called either in an interrupt
+    /// context or in application context under a critical section lock. This
+    /// function normally does not take an extended period of time. However,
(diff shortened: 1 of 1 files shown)


bakerstu (approved):
This probably works, but I'm not sure it is the best fix. I think a better fix might be to place an atomic lock in tx_char() starting just before the send() call and exiting after the MAP_UARTIntEnable() call. WDYT?

balazsracz:
There is an atomic lock already around the tx_char(). However, there are
priority zero interrupts (e.g. DCC interrupt) which the atomic lock does
not hold against.

bakerstu:
> There is an atomic lock already around the tx_char(). However, there are priority zero interrupts (e.g. DCC interrupt) which the atomic lock does not hold against.

Okay, I'm confused why this race condition would occur in the serial logging then. Logging should not be a priority zero interrupt. The DCC driver doesn't use the UART driver for TX, so why would its priority matter?

balazsracz:
> > There is an atomic lock already around the tx_char(). However, there are priority zero interrupts (e.g. DCC interrupt) which the atomic lock does not hold against.
> 
> Okay, I'm confused why this race condition would occur in the serial logging then. Logging should not be a priority zero interrupt. The DCC driver doesn't use the UART driver for TX, so why would its priority matter?

The question is how much time can elapse between these two pieces of code:

```
   MAP_UARTCharPutNonBlocking(base_, data);
   ...
   MAP_UARTTxIntModeSet(base_, UART_TXINT_MODE_EOT);
   MAP_UARTIntClear(base_, UART_INT_TX);
```  

These are normally consecutive instructions or something the like.
However, if there is enough time between them that the UART actually finishes transmitting the contents, then we clear the EOT interrupt after it triggered. So we never get the call.

This is why I think an intervening priority zero interrupt can cause the race condition to be triggered.

bakerstu:
> > > There is an atomic lock already around the tx_char(). However, there are priority zero interrupts (e.g. DCC interrupt) which the atomic lock does not hold against.
> > 
> > 
> > Okay, I'm confused why this race condition would occur in the serial logging then. Logging should not be a priority zero interrupt. The DCC driver doesn't use the UART driver for TX, so why would its priority matter?
> 
> The question is how much time can elapse between these two pieces of code:
> 
> ```
>    MAP_UARTCharPutNonBlocking(base_, data);
>    ...
>    MAP_UARTTxIntModeSet(base_, UART_TXINT_MODE_EOT);
>    MAP_UARTIntClear(base_, UART_INT_TX);
> ```
> 
> These are normally consecutive instructions or something the like. However, if there is enough time between them that the UART actually finishes transmitting the contents, then we clear the EOT interrupt after it triggered. So we never get the call.
> 
> This is why I think an intervening priority zero interrupt can cause the race condition to be triggered.

Ah, yes, I see. I suspect that reordering some of the TivaWare calls might eliminate this race, but it would require quite a bit of experimenting. For now, let's go with this, but can you provide more inline documentation as to why you think the fix is required, like the argument you articulated in this conversation for example.

bakerstu:
> TivaWare

I mean Driverlib, this is not a Tiva driver, though the Tiva driver may have the same issue.

balazsracz:
I thought about it but could not find a better ordering of the calls.

- we can't change from FIFO to EOT interrupt before filling the tx fifo, otherwise we could get the interrupt straight away.
- we do need a tx int clear, because the fifo low interrupt might have triggered the tx int.
- we can't have the tx int clear before the switch to EOT either, becauser the fifo low is a level triggered interrupt.
- we don't really get to know whether we will empty our software buffer into the tx fifo until we do copy the bytes, so we woukld not realy know at the beginning of tx_char whether we will end up coming out with the  fifo or the eot interrupt shape.

PR #950, opened 2026-08-30
CC32xxnet: makes the wakeup socket non-blocking.

balazsracz:
When the NWP is out of transmit buffers, the select_wakeup could block the main executor. This is really bad.

As an alternative strategy, we make the wakeup socket nonblocking. To avoid getting stuck with the fd_set, we limit the select timeout to 20 msec. We always wake up select and re-do the fd_set after this time.

### DIFF
--- a/src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
+++ b/src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
@@ -1014,9 +1014,9 @@ void CC32xxWiFi::wlan_task()
         SlFdSet_t wfds_tmp = wfds;
         SlFdSet_t efds_tmp = efds;
         SlTimeval_t tv;
-        // Wake up every 20 msec in case we missed a select_wakeup.
+        // Wake up every 40 msec in case we missed a select_wakeup.
         tv.tv_sec = 0;
-        tv.tv_usec = 20000;
+        tv.tv_usec = 40000;
 
         result = sl_Select(fdHighest + 1, &rfds_tmp, &wfds_tmp, &efds_tmp, &tv);
 
--- a/src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
+++ b/src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
@@ -988,6 +988,14 @@ void CC32xxWiFi::wlan_task()
 
     unsigned next_wrssi_poll = (os_get_time_monotonic() >> 20) + 800;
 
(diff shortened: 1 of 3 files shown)


src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -1055,6 +1064,9 @@ void CC32xxWiFi::wlan_task()
                     /* this is the socket we use as a signal */
                     char data;
                     sl_Recv(wakeup, &data, 1, 0);
+                    portENTER_CRITICAL();
```
bakerstu, line 1067: Should we make these atomic locks instead? More below.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -1006,8 +1014,9 @@ void CC32xxWiFi::wlan_task()
         SlFdSet_t wfds_tmp = wfds;
         SlFdSet_t efds_tmp = efds;
         SlTimeval_t tv;
-        tv.tv_sec = 1;
-        tv.tv_usec = 0;
+        // Wake up every 20 msec in case we missed a select_wakeup.
+        tv.tv_sec = 0;
```
bakerstu, line 1018: I find it curious that we had 1 second here before. I wonder why we did that.

bakerstu (approved):

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
balazsracz: This was due to the WRSSI poll running in this loop.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
balazsracz: There is no difference, but the rest of this code is written with explicit enter/exit critical commands. So in order to keep a consistent style, I used that.

src/freertos_drivers/net_cc32xx/CC32xxWiFi.cxx
```diff
@@ -1006,8 +1014,9 @@ void CC32xxWiFi::wlan_task()
...
+        tv.tv_usec = 20000;
```
balazsracz, line 1019: note: I changed this top 40 msec because I found that we were spending 5% of the CPU time in select when using 20 msec.

PR #954, opened 2026-08-31
Learn remote aliases from initialization complete and node verified messages.

balazsracz:
Updates the remote alias listener class to act on incoming messages of MTIs
100, 101 (node initialization complete / with simple) and 170, 171 (global
verified node id).

Adds the 101 and 171 to the MTI enum. These have been in the standard for a
while now.


bakerstu (approved):
This code has virtually zero useful comments, including missing Doxygen comments. Consider adding verbose comments.

PR #960, opened 2026-09-18
ESP32-C6 Support

ngpaladi:
Added support for the ESP32-C6 in order to run OpenMRN on [FemtoLCC](https://www.uncommonmodels.com/FemtoLCC/).

Also updated lookup tables that needed to be kept in sync with ESP32 APIs, as necessary (see `src/freertos_drivers/esp32/Esp32SocInfo.cxx:254-264`).


atanisoft:
There are some adjustments in there to squash ESP32-C3 and ESP232-C6 in places likely, but overall it looks good.

src/freertos_drivers/esp32/Esp32HardwareTwai.cxx
```diff
@@ -70,6 +70,17 @@
 #include <soc/gpio_sig_map.h>
 #include <stdint.h>
 
+#if SOC_TWAI_CONTROLLER_NUM > 1
```
atanisoft, line 73: IIRC this class is not used by the Arduino code now, it was adjusted I believe. It is also not compatible at all beyond ESP-IDF v5.4.4. For the newer versions there is a full replacement that I am working on using the standard TWAI driver code but is a bit simpler than this one is.

src/freertos_drivers/esp32/Esp32Gpio.hxx
```diff
@@ -53,14 +53,21 @@
 #define IS_GPIO_OUTPUT(pin) (GPIO_IS_VALID_OUTPUT_GPIO(pin) &&                 \
                              GPIO.enable.data & \
                              BIT(pin & SOC_GPIO_VALID_OUTPUT_GPIO_MASK))
-#else // NOT ESP32-C3
+#elif defined(CONFIG_IDF_TARGET_ESP32C6)
```
atanisoft: since this block is identical to the above for esp32-c3, can we combine them by adjusting the #if statement? something like:
```
#if defined(CONFIG_IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32C6)
```

ngpaladi: I had kept this separate in case we wanted them split in the future, but have combined them now.

atanisoft: It's quite unfortunate that ESP-IDF won't likely add the missing API, it is missing so far for all platforms.

src/freertos_drivers/esp32/Esp32HardwareTwai.cxx
ngpaladi: Should this be dropped then? Or is it better to leave it as a holdover solution until the replacement is ready and add a comment to that effect?

atanisoft: I believe we can keep this, the replacement is ready now as EspCan. It is possibly less efficient than the replacement (it is virtually ready now, just not proposed for merge), it would need to be put up to a test to confirm.

